PR #53, opened 2016-08-19
Now making map zoom in extra far for fire events

rboller:
Thought I'd attempt this hack for zooming into fire events to make them more clearly visible.


tgunnoe:
Looks good to me

PR #72, opened 2016-11-15
Gifshot fixes

Benjaki2:
Made additions to GIF
* Date stamp option
* Error handling for bad urls
* Now using forked version of gifshot


tgunnoe:
Looks good to me, we still need to fix the node modules directory (something I need to continue working with)

PR #74, opened 2016-12-01
Analytics

Benjaki2:
Added GA Event tracking to various features: Animation, Tour, Natural Events, Share link, Data Download, and Image Download

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
         "promise-queue": "^2.2.3",
         "react": "^15.3.1",
         "react-dom": "^15.3.1",
-        "worldview-timeline-components": "1.1.6"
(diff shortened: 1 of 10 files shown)


tgunnoe:
Why is the GA portion in the timeline module?  Just curious.  And I suppose you'll be adding analytics as a feature in the options repo as well?

Benjaki2:
@tgunnoe good question. I plan on renaming the repository to worldview-components or worldview-external-components as it is home to a lot of non-timeline specific modules already. Another option would be to create various more specific repositories. I see positive and negatives to both approaches.

tgunnoe:
cool. sounds good to me.

Issue #206, opened 2017-03-24
Consider using svg for WV logo and icons


CodeMacabre:
Currently, the WorldView logo and icons in the main widget panel are png, which makes them fairly blurry, especially at higher resolutions.

![wv-widget-icons](https://cloud.githubusercontent.com/assets/19981049/24309260/13883752-10c3-11e7-9c1a-1388c141456d.PNG)

Where possible, can they be updated to either:
1) Vector (svg) icons (possibly larger filesizes?)
2) Use Font Awesome equivalent icons (faster as FA fonts are already used?)?

Font Awesome replacement suggestions:
+ Layers - [`fa-clone`](http://fontawesome.io/icon/clone/)
+ Events - [`fa-calendar`](http://fontawesome.io/icon/calendar/)
+ Data - [`fa-download`](http://fontawesome.io/icon/download/)
+ Show - [`fa-angle-double-down`](http://fontawesome.io/icon/angle-double-down/)
+ Hide - [`fa-angle-double-up`](http://fontawesome.io/icon/angle-double-up/)
+ Show Layer - [`fa-eye`](http://fontawesome.io/icon/eye/)
+ Hide Layer - [`fa-eye-slash`](http://fontawesome.io/icon/eye-slash/)
+ Layer Options - [`fa-sliders`](http://fontawesome.io/icon/sliders/)
+ Remove Layer - [`fa-times`](http://fontawesome.io/icon/times/)

It would potentially be a little more tricky for the latter four layer icons, as they are all currently part of the same png spritesheet (`./web/images/wv.layers/show-hide.png`). I believe svgs can be used in the same way, but using Font Awesome might be faster than loading and calculating the spritesheet coordinates (may need testing to verify).

CodeMacabre:
For the WorldView logo, if you don't have the source artwork at higher resolution in vector format, I volunteer to copy it into an svg. 🙂

Benjaki2:
@CodeMacabre It would be great if you can copy the logo to svg! I also agree with using FA for the others icons, I can work on that.

Thanks!

CodeMacabre:
Done! I couldn't find an equivalent font (without paying), so I just redrew the text (with slight modifications and improved kerning). I used the official NASA logo from [nasa.gov](https://www.nasa.gov/) so that's why there are minor color and shape differences. See below for comparison (svg at the top, png at the bottom):

[...]

PR #233, opened 2017-03-31
Replace WV logo png with svg

CodeMacabre:
See issue #206.


CodeMacabre:
@Benjaki2 I've made a PR in the `worldview-options-eosdis` repo adding the svg logos (I've also updated the mobile logo). I haven't removed the old png logos in case they are used elsewhere.

I also haven't made any changes to `wv-icon-w-shadow.png`; shadows are difficult to do in svg so perhaps the shadow could be added via css rather than use a separate image? Possibly something for a separate issue.

Benjaki2:
@CodeMacabre I had trouble with `wv-icon-w-shadow.png`. I remember trying to make it an svg and being unsuccessful. It is being used as a stamp in the animation GIFs that you can generate. Each frame is generated with canvas. So, maybe the svg could be used and we style the canvas on-the-fly?

Issue #249, opened 2017-04-10
Simplify Install


Benjaki2:
With node server

davidbernat:
@Benjaki2 : What are the likelihoods of migrating from vagrant to, for example, docker in the future?  Is this something that can be helpful as provided from the user community?  Thanks,

Benjaki2:
We actually do use docker for deployment and deployment-related testing. But for local installs we plan to avoid both docker and vagrant, and go with simple node server. Initializing everything with `npm start`. I will probably create a new branch today if you are interested in contributing... I will link all commits to this issue.

Also feel free to look at other issues in the queue and approach what you see fit. For now, using [these manual instructions](https://github.com/nasa-gibs/worldview/blob/master/doc/manual_setup.md) is probably the easiest way to get developing locally.

Thanks @davidbernat

rboller:
@Benjaki2 I've submitted PR #271 to add support for Windows but note that Travis is now [failing the build](https://travis-ci.org/nasa-gibs/worldview/builds/231207713) which appears to be an unrelated OL problem.  

Also note that the `npm install` command under Ubuntu fails during the chromedriver installation step.  I didn't do much troubleshooting there but we should investigate more in the future. 

One last thing for this ticket: we should remove the vagrant-related installation files since that method will be deprecated.

Benjaki2:
@rboller Looks like travis is failing due to it using an extremely outdated version of node. I tested this by updating the node version in the config on a separate branch. Looks like it works. But before we go ahead with this update we should look into what version of node bamboo is using and make sure these two correspond.

Benjaki2:
Also specifying the advised node versions to use in the package.json file would be a good ticket as well. We can set up so that if a user is using a node version above or below the specified version range, they will receive an warning.

rboller:
FYI I found the problem with Ubuntu installation - basically certain installations of node are only available as `nodejs` rather than `node` ([background info](https://stackoverflow.com/questions/18130164/nodejs-vs-node-on-ubuntu-12-04)).  So the `npm install` command was failing when it was trying to invoke `node`.

The workaround is to install `nodejs-legacy` which also makes node available as `node` rather than only `nodejs`:
`sudo apt-get install nodejs-legacy`

After this, the installation works well.  I'll update the documentation accordingly.

Benjaki2:
Good news! I just merged master into the install branch to resolve conflicts. Feel free to add the updates there or create a PR. Almost there!

-Ben

[...]

PR #271, opened 2017-05-11
Updated install to be Windows-friendly

rboller:
@Benjaki2 please review overall and especially check my [replacement of the rsync command in bin/wv-options-build with a cp](https://github.com/nasa-gibs/worldview/compare/install...install-universal?expand=1#diff-9f809eedf969643c8356ab0a09c6d00c).  Thanks!


Benjaki2:
@rboller I removed the only remaining vagrant-specific file I could identify and updated travis to use node v6.4.1. Shall we merge?

rboller:
Yep, please merge when ready!  Note that I just retried the installation on Ubuntu and it failed in the same way as noted in #249 so I'll file a separate ticket for that.

Issue #316, opened 2017-06-01
Add layer description option to Layer List


minniewong:
Refer to #65

minniewong:
Use an "i" icon

Issue #65, opened 2016-09-14
Ability to view layer descriptions more easily


rboller:
Worldview currently provides detailed, "human readable" layer descriptions to help users understand how each layer can be used, their limitations, and why they are important.  Here's an example:

https://github.com/nasa-gibs/worldview-options-eosdis/blob/master/common/config/metadata/amsr2/CloudLiquidWater.md

At present, these descriptions are only available when using the layer selection widget (i.e., the "+ Add Layers" button).  But it would be very helpful if they were more easily available after layers were added to the active layer list.  

So this ticket is to add the capability to view these layer descriptions "on demand", likely invoked from the active layer list.

minniewong:
Connects to #329

PR #329, opened 2017-06-09
Layer descriptions

ZachTRice:
Layer descriptions contained within am "i" button in the sidebar. If no description is available a disabled "i" button is shown. If a description is available, clicking the "i" button expands a modal box. The "i" button and modal box are only available within the desktop view.
Connects to #65
connects to #316


web/index.html
```diff
@@ -382,13 +383,13 @@
       </header>
       <div id="layer-modal-main">
         <div id="layer-categories">
-          
+
         </div>
         <div id="selected-category">
-          
+
```
Benjaki2: I'd leave out these spaces if there isn't anything inside the divs `<div></div>`

ZachTRice: I agree. This is how it is currently on master, though I should have cleaned it up while I was working in this area. I'm not sure why git is picking this up as a change.

web/js/layers/wv.layers.info.js
```diff
@@ -0,0 +1,85 @@
...
+
+    var loaded = function(custom) {
+        $dialog = wv.ui.getDialog();
+        $dialog
+            .attr("id", "wv-layers-info-dialog")
+            .attr("data-layer", layer.id);
+        renderDescription($dialog);
+
+
+        var names = models.layers.getTitles(layer.id);
```
Benjaki2: declare vars at top of scope

```
var loaded = function(custom) {
     var names;
```

web/index.html
ZachTRice: Ah, the differences is that there were spaces after the line return that are no longer present.

web/js/layers/wv.layers.info.js
```diff
@@ -0,0 +1,85 @@
...
+        loaded();
+    };
+
+    var loaded = function(custom) {
+        $dialog = wv.ui.getDialog();
+        $dialog
+            .attr("id", "wv-layers-info-dialog")
+            .attr("data-layer", layer.id);
+        renderDescription($dialog);
+
```
Benjaki2: excess line

web/index.html
Benjaki2: ahhh makes sense. Thanks for cleaning up those spaces.

ZachTRice:
Thanks, I appreciate the nit picky! 

Couple things I added to that last commit. Removed a `</div>` after the opening `<body>` tag that's been in master. Also, just noticed I had to tweak the css a bit to add back in the `li` / `a` elements to the class selectors since bank.css is setting these and my hover state colors were being over written with the default link color. 

At some point, I'd also like to remove as much element.class selector's as possible. Just to make scoping and compartmentalizing the css a bit easier.

Benjaki2:
Wait till after next release to merge

ZachTRice:
Closing this PR because this feature will be merged into a new branch with search-descriptions. The new combined feature branch will then have a PR into development.

Issue #332, opened 2017-06-13
Add configurable modalView options to permalinks documentation


ZachTRice:
After #327 has been implemented, the WOR/Permalinks documentation should be updated to include the new URL parameter `modalView`

Issue #327, opened 2017-06-08
Change polar product picker to list of measurements


minniewong:
Make it look the same as when you click the "All" category in the product picker.

rboller:
After some discussion this morning, we decided that there should be three configurable viewing options for the product picker on a per-projection basis: "categories", "measurements", and "layers".  

"Categories" is what's currently shown in the geographic projection.

"Measurements" would look like what happens when you currently click on the "All" category in the geographic view.

"Layers" is what we currently have in the polar views which is a raw listing of available layers for the current projection.

PR #358, opened 2017-06-29
Polar picker

ZachTRice:
This branch changes the way the Add Layer modal is displayed. Connects to #327 & Connects to #332
- The Artic/Antarctic regions will now display measurements by default instead of all layers. 
- The modal window logic has been rewritten to only check for the projections once on load.
- Tweaked breadcrumb logic so that the current correct "top" view will be linked based on which view has been loaded in.
- There is a new permalink option that can be added to the url, `modalView`, which can be set to a value of `layers`, `measurements`, or `categories`. Setting this parameter will change the Add Layers modal to the desired view.
- Added `modalView` documentation  to the testing documentation page.


ZachTRice:
This feature's measurement and category views are not pulling in the correct projections. Canceling PR now until this bug is fixed.

Issue #314, opened 2017-06-01
Improve WMTS interoperability


rboller:
Improve ability to incorporate layers from WMTS services other than GIBS.

rboller:
Added a `wmts-interoperability` branch to both `worldview` and `worldview-options-eosdis` repos as a work in progress.  Basically this added a Gridded Population of the World layer from the [SEDAC WMTS](http://sedac.ciesin.columbia.edu/maps/services) which then warranted several updates to the codebase. This included adding layer configuration parameters for `style`, `matrixIds`, and `extent`, though `extent` is currently implemented as a quick [hack](https://github.com/nasa-gibs/worldview/blob/wmts-interoperability/web/js/map/wv.map.layerbuilder.js#L174-L182).

Currently the main problem is that it appears that SEDAC's WMTS tile origin is in the upper left corner but the map origin is in the lower left.  That makes it challenging to keep the map aligned at every zoom level.  Example of the top left tile at zoom levels [0](http://sedac.ciesin.org/geoserver/gwc/service/wmts?Service=WMTS&REQUEST=GetTile&Version=1.0.0&layer=gpw-v4:gpw-v4-population-count_2020&tilematrixset=EPSG:4326&Format=image%2Fpng&TileMatrix=0:2%20Degrees&TileCol=0&TileRow=0), [1](http://sedac.ciesin.org/geoserver/gwc/service/wmts?Service=WMTS&REQUEST=GetTile&Version=1.0.0&layer=gpw-v4:gpw-v4-population-count_2020&tilematrixset=EPSG:4326&Format=image%2Fpng&TileMatrix=1:1%20Degree&TileCol=0&TileRow=0), [2](http://sedac.ciesin.org/geoserver/gwc/service/wmts?Service=WMTS&REQUEST=GetTile&Version=1.0.0&layer=gpw-v4:gpw-v4-population-count_2020&tilematrixset=EPSG:4326&Format=image%2Fpng&TileMatrix=2:30%20Arc%20Minutes&TileCol=0&TileRow=0).

May be able to work around this if a different extent and map origin is set at every zoom level in wv.map.layerbuilder;  [see hack here](https://github.com/nasa-gibs/worldview/blob/wmts-interoperability/web/js/map/wv.map.layerbuilder.js#L174-L182) for aligning the top zoom level.  Some background OL docs [here](http://openlayers.org/en/latest/apidoc/ol.tilegrid.TileGrid.html) for how to specify the alignment.

[...]

PR #371, opened 2017-07-03
Improved WMTS interoperability

rboller:
Updates to support additional parameters that WMTS servers (other than GIBS) may use


rboller:
connects to #314

web/js/map/wv.map.layerbuilder.js
```diff
@@ -149,12 +149,19 @@ wv.map.layerbuilder = wv.map.layerbuilder || function(models, config, cache, Par
...
+			});
+		}
+		else {
+			matrixIds = def.matrixIds;
+		}
+
         extra = "";
-        if(day) {
+
+		if(day) {
```
Benjaki2: @rboller Just some extra indentation at lines 164 and 152-160. We are going to automate the indentation soon so it isn't a big deal. All else looks good!

Benjaki2: @rboller  I'm seeing the tabbing is better than it was when you started. We should add this to our linter after we switch over to 2-spaces.

Issue #372, opened 2017-07-03
Add CSS Autoprefixer as a dev tool


Benjaki2:
* Probably will be removed after preprocessors are incorporated.

ZachTRice:
I think we should run autoprefixer on the tracked css files rather than piping those prefixes into only the debug/release folders because these css changes should be tracked. If someone is committing new css changes, they add a property that has fallback prefixes, those prefixes should be added and tracked. This also means we should run autoprefix once, check the changes, and then commit all the css files again.

minniewong:
Look into Travis/CI

PR #376, opened 2017-07-05
Added autoprefix task to Grunt

ZachTRice:
Connects to #372
- Added autoprefixer to package.json
- Added grunt-postcss to package.json
- Added 'postcss' task to Grunt which runs autoprefixer
- `grunt postcss` runs autoprefixer (we can add more postcss plugins
to this process)
- Added postcss to grunt build
- Moved postcss to run before concat
- Changed grunt command from postcss to autoprefix. Added documentation.
- Now... `grunt autoprefix` runs autoprefixer

### DIFF
--- a/web/css/wv.alert.css
+++ b/web/css/wv.alert.css
@@ -1,5 +1,4 @@
 #wv-info-button {
-     -webkit-transition: background 1s;
     -webkit-transition: background 500;
     transition: background 500;
 }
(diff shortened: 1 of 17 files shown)


web/css/wv.alert.css
```diff
@@ -1,5 +1,6 @@
 #wv-info-button {
      -webkit-transition: background 1s;
```
Benjaki2: might want to get rid of the first transition

Issue #140, opened 2017-02-09
Update share options for social media


mcechini:
Add ability to share directly to twitter or facebook when you click the share icon. Additionally, there would be a "copy link" option to support the existing option.  Maybe also have a "email" option.

rboller:
Agreed, back in Jira-land (WV-1582) we also discussed this but haven't pulled the ticket over to here yet.

rboller:
The new [State of the Oceans](https://podaac-tools.jpl.nasa.gov/soto/) has a nice share feature which includes direct links to social media.  We might look at doing something similar.

rboller:
It would also be nice if this showed a preview image (which will probably be static) as discussed in WV-1665.

minniewong:
The share should include a generic image of Worldview or if easy enough, a snapshot of the actual URL being shared.

minniewong:
Focus on Twitter and Facebook; look into Instagram

ZachTRice:
You cannot share media on Instagram using the API. Here is their reasoning:

> At this time, uploading via the API is not possible. We made a conscious choice not to add this for the following reasons:
> 
> Instagram is about your life on the go – we hope to encourage photos from within the app.
> We want to fight spam & low quality photos. Once we allow uploading from other sources, it's harder to control what comes into the Instagram ecosystem. All this being said, we're working on ways to ensure users have a consistent and high-quality experience on our platform.
>

ZachTRice:
For Twitter: If you look at the [State of the Oceans](https://podaac-tools.jpl.nasa.gov/soto/) sharing feature, when you share a Tweet on there, the user experience is not very good because nothing is being passed to the text box, including the URL. The intended action will be to write a short message and have a shortned link afterwards.

I am thinking something along the lines of:
> Check out what I found in NASA's Worldview - Shortened_URL

Without any information passed, there is no reference back to what was suppose to be shared and in the case of SOTO, it is up to the user to navigate back to the page and add text + a link.

@rboller, @minniewong Let me know what you think of the sharing text I came up with. Ideally the text should be as short as possible. The shortened link will use up to 30 of the 140 character maximum. If the message is 40 characters then that still leaves about 70 characters for the end-user to add his own text to the message.

minniewong:
Or some text that incorporates #NASAWorldview
(Other users, and our communications people, use that hastag - https://twitter.com/hashtag/NASAWorldview?src=hash)

rboller:
[...]

Issue #139, opened 2017-02-09
Update share icon


mcechini:
Update the share icon to a more standard "box with arrow" or "3-dot-connected" icon.

rboller:
To this old timer, the "more standard" icon _is_ the "link" icon.  :)   But I suppose we could consider one of the other share icons [here](http://fontawesome.io/icons/).

CodeMacabre:
My personal preference (partly inspired by [this article](https://www.fastcodesign.com/3031872/why-isnt-there-a-standard-share-button)) is the 'box with arrow', as I think it conveys the sharing meaning most clearly. FontAwesome's `share-square-o` is great, but depending on how it looks alongside the others, `share` could also work:

![share-square](https://cloud.githubusercontent.com/assets/19981049/24306986/6619bb98-10ba-11e7-8948-55ce101d8404.PNG)

Benjaki2:
Thoughts @minniewong @rboller ?

My initial instinct was to go with the `fa-share-alt` but agree with the above article's comment on it's lack of intuitiveness.

rboller:
If I had to choose between `fa-share` and `fa-share-square-o`, I like the latter better.

That said, is the subtext to this discussion that the "link" icon is dead?  I may be the curmudgeon on this one, but I'll also point out that Google Maps still uses a link-like icon.

minniewong:
which one is fa-share-alt? I also like fa-share-square-o if we had to choose between the two.

Benjaki2:
@minniewong [share-alt](http://fontawesome.io/icon/share-alt/) . I am in support of `fa-share-square-o` if we do decide to make a change

ZachTRice:
In the past I have used the the 3-dots (`fa-share-alt`), because this was the icon that the ShareThis uses and often that service was being dropped in to my projects to provide analytics. I imagine `fa-share-square-o` is going to be more easily recognized by people who use Apple products. Both have a huge user base and I feel both icons convey the intended action. I support using either icon. I'll add `fa-share-square-o` to my share-feature branch.

ZachTRice:
The icon needs to be updated in the start-up tour window.

Issue #60, opened 2016-09-13
Permalink window doesn't fit into Android window when in portrait mode


rboller:
See attached screen capture

![android-permalink](https://cloud.githubusercontent.com/assets/7243823/18457881/5849d4f8-792a-11e6-96cf-55ea80046741.png)

Dylan-Brown:
I was able to recreate this issue on my Android Galaxy S2 phone. I can work on this.
What I notice is that on both the android and desktop versions, the box containing the share info seems to  align on the right side of the link icon. My  thought is that if it was reduced in width, say, to 2/3 (or around that value),  that it would fit on the android screen?

![screenshot_2016-10-03-10-53-43-2](https://cloud.githubusercontent.com/assets/10562115/19042013/6057a93e-8958-11e6-8ebe-7c23e0b21c60.png)

Benjaki2:
@Dylan-Brown Sounds good. Could you make it responsive so that it stays the same size on desktop views and is a smaller width on mobile views?

Dylan-Brown:
@Benjaki2 I've looked at / played with the style attribute of this "<div>" element, and through some searching online, I haven't found a way to alter the in-line style of the element in a way that makes it responsive to the screen size. 

I've thought about changing the wv.link.css file to include the style elements and seeing if there was a way to make the width dynamic from there, but that doesn't seem to be the approach with all of the other "major" divs, as they also use in-line style.

I think I lack perspective when it comes to web dev and am struggling to figure out how everything pieces together. Could you possibly point me to a resource that could explain this at a high level?

Benjaki2:
@Dylan-Brown you can find whether the device is under our break point width using this method:

https://github.com/nasa-gibs/worldview/blob/master/web/js/util/wv.util.browser.js#L203

e.g. `var mobile = ui.util.browser.small(); // mobile will be true or false `

PR #398, opened 2017-07-13
Social sharing feature

ZachTRice:
Connects to #140 
Connects to #139
Connects to #60

**This feature needs to be tested on Master to ensure URL shortening is working properly. On UAT, Twitter & Email links will appear as the full link in the message fields but on Master these should be shortened links.


web/js/link/wv.link.ui.js
```diff
@@ -14,135 +14,205 @@ wv.link = wv.link || {};
...
+    // Render Dialog Box Content
+    self.reactComponent = ReactDOM.render(Widget, $dialog[0]);
+
+    // Update react link states when dialog is shown.
+    self.setLink(fbLink, twLink, rdLink, emailLink, self.reactComponent.updateLinkState);
+
+    // If selected during the animation, the cursor will go to the
+    // end of the input box
+    var updateLink = function() {
+      $('#permalink_content').val(models.link.get());
```
Benjaki2: I wouldn't do any selecting of react components outside of react. `e.g.` Pass a state into the component that Focuses the element. This way we are decoupling the component from Worldview and Jquery (maybe we can dispose of it in the future)

Benjaki2:
@ZachTRice looks good! Just need to release a new version of WVC and update the package.json with that verision.

Issue #348, opened 2017-06-21
Switch jshint for eslint


Benjaki2:
* Swap in Package.json
* Swap in Gruntfile
* Maybe add a .eslint file

PR #406, opened 2017-07-17
Switch linter

ZachTRice:
Connects to #348


ZachTRice:
@Benjaki2 let's tweak this / add any additional rules that you think would work now. 

My thought is that this file will evolve quite a bit over time as we [Overhaul JS](https://github.com/nasa-gibs/worldview/projects/5). We will create a ticket to address an issue, such as removing undeclared variables, perform the action, add an eslint rule, and close the ticket.

package.json
```diff
@@ -1,58 +1,59 @@
...
+    "grunt-contrib-csslint": "0.2.x",
+    "grunt-contrib-cssmin": "1.0.x",
+    "grunt-contrib-jshint": "0.8.x",
+    "grunt-contrib-uglify": "2.0.x",
+    "grunt-contrib-watch": "^1.0.0",
+    "grunt-exec": "0.4.x",
+    "grunt-git-rev-parse": "0.1.x",
+    "grunt-line-remover": "0.0.x",
+    "grunt-markdown": "0.7.x",
+    "grunt-minjson": "0.4.x",
```
Benjaki2: @ZachTRice Looks like the spacing was swapped here again

Issue #427, opened 2017-07-24
Degrees map when rotated should not exceed 360 degrees.


ZachTRice:
In Arctic views, if you rotate the map with alt + drag, you can rotate the map more than 360 degrees. The same applies for mobile devices; using finger dragging on the phone builds up the degrees passed 360 degrees. This happens in both the URL and the sidebar indicator.

After 360 degrees, the map should go reset to 0 degrees (in url and in sidebar).

This issue occurred on Windows in Firefox and Chrome latest stables and on Android 4.4 in Chrome.

ref: (note -1174 degrees)
![degrees-issue](https://user-images.githubusercontent.com/17348884/28529799-7ffba148-705f-11e7-9cda-45da4356a2a1.png)

ZachTRice:
Also, we discussed briefly in slack that there are negative degrees being displayed. Does it make sense to have negative degrees at all? 45 degrees is the same as -315 degrees:

https://worldview.earthdata.nasa.gov/?p=arctic&l=VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor,Reference_Labels(hidden),Reference_Features(hidden),Coastlines&t=2017-07-24&z=3&v=-23982385.12125391,-22650432.859063867,22690960.821441174,19499038.286227696&r=45

https://worldview.earthdata.nasa.gov/?p=arctic&l=VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor,Reference_Labels(hidden),Reference_Features(hidden),Coastlines&t=2017-07-24&z=3&v=-23982385.12125391,-22650432.859063867,22690960.821441174,19499038.286227696&r=-315

Would like to hear your opinion on this @ryan

minniewong:
We should reach out to Polar users to see what they think

PR #447, opened 2017-08-02
Fixed alt-dragging in polar projection to always display a value between -360 & 360 degrees

ZachTRice:
Connects to #427

### DIFF
--- a/web/js/map/wv.map.rotate.js
+++ b/web/js/map/wv.map.rotate.js
@@ -22,31 +22,31 @@ wv.map.rotate = wv.map.rotate || function(ui, models, map) {
   var self = this;
 
   /*
-     * Initializes by triggering build methods
-     *
-     * @method init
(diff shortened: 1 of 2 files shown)


web/js/map/wv.map.rotate.js
```diff
@@ -146,15 +144,28 @@ wv.map.rotate = wv.map.rotate || function(ui, models, map) {
...
+      currentDeg;
 
     view = ui.selected.getView();
     radians = view.getRotation();
     models.map.rotation = radians;
     self.setResetButton(radians);
-  };
-
 
+    currentDeg = (view.getRotation() * (180.0 / Math.PI));
```
Benjaki2: @ZachTRice this logic looks identical to what's happening in `this.rotate` . Should probable isolate that into it's own method and use it in both places

PR #451, opened 2017-08-03
Fixed files modified after running grunt were slightly different.

ZachTRice:
- After updating to 2 spaces, running grunt now caused slight changes in the edsc js directory. The .call and .join at the end of files have been moved up from the new line.


localjo:
I noticed this happening too. I'm curious why Grunt is modifying files in the repo. Shouldn't it only be modifying files in ignored directories like `/build` and `/dist`?

localjo:
Answering my own question here. Looks like the `edsc` stuff is CoffeeScript that's built by Grunt. That was [added two years ago](https://github.com/nasa-gibs/worldview/commit/3794003c2fbd8c7480cb80a58d9e01a241e5ecba). It's probably not an urgent thing to change now, but I'm guessing we could probably ditch the CoffeeScript source and the Grunt task and just leave the JS files in there. But that'd be a different issue to tackle. These changes are probably fine for now.

ZachTRice:
Appears to happen when it runs `grunt coffee` during the build step

Benjaki2:
@ZachTRice good point. Should we pull coffee out of the build step with this PR as well?

ZachTRice:
Yes, let's remove the coffee directory, coffee from grunt and the coffee dependency from package.json. I don't see any reason to keep all of that in there.

Issue #388, opened 2017-07-07
Preserve Active Layers When Changing Event Date


minniewong:
e.g. Sea Ice/Icebergs events - Brightness Temperature is loaded but not on when you select an iceberg event. Enter some logic so that when you turn on the brightness temperature layer, it will stay on if you change the event date

minniewong:
If you've turned on a recommended layer, and change the event date, the recommended layer should stay on

localjo:
I'm thinking about making this a simple text filter;

![screen shot 2017-08-03 at 4 11 52 pm](https://user-images.githubusercontent.com/516388/28941722-e96a8432-7866-11e7-90d7-136701cba7f7.png)
![screen shot 2017-08-03 at 4 14 31 pm](https://user-images.githubusercontent.com/516388/28941721-e9695c1a-7866-11e7-9153-17af4f3f530d.png)

That way you could filter by event types, just by starting to type something like "Tropical Storm", "Iceberg", "Fire", etc. And you could also filter for specific names of events like "A68B" or "Sandy" or "California", etc.

I could also add a subtle "x" button in the filter input to clear it quickly.

Thoughts?

localjo:
Going to tackle #407 first because it'll be easier to test this if I can quickly filter to relevant events.

localjo:
Is there another way to "change event dates" other than clicking on an event in the list with the same name on a different date? Ie;

Step 1 (click on initial event);
![screen shot 2017-08-03 at 4 24 46 pm](https://user-images.githubusercontent.com/516388/28942188-741d1b20-7868-11e7-907d-eff86ef011bd.png)

Step 2 (click on different instance of event);
![screen shot 2017-08-03 at 4 25 05 pm](https://user-images.githubusercontent.com/516388/28942189-741d34ac-7868-11e7-9c44-a6f5a6df3995.png)

I guess you could also manually scrub the timeline looking for other event dates, but that would leave the state of the layers the same, so I'm guessing that this issue is referring only to the user action I described above. Just checking that there aren't any other UI actions (yet) that would "change event dates". :)

rboller:
@localjo those events that you listed happen to be different events (A68B vs A68A) and happen to only have one date associated with them.  Others like Iceberg B15Z have a ton of them:

![screen shot 2017-08-03 at 5 35 37 pm](https://user-images.githubusercontent.com/7243823/28944878-423376b8-7872-11e7-96c9-d43e6b507421.png)

[...]

PR #463, opened 2017-08-09
Display all markers for events

### LINKED ISSUE #407: Show All Active Event Markers on Map
Display all event markers for what's active, need to include labels and event icons

localjo:
Fixes #407, but is still a work in progress.


localjo:
@Benjaki2 This first commit here fixes the stuff I noticed with OpenLayers. We can either leave it as a part of this PR, and merge it when I'm done here, or you can cherry pick it out into another branch. Just let me know what you want to do.

Benjaki2:
@localjo sounds good! Prob best just to leave it in and merge it with the feature. 👍

localjo:
Some outstanding things to take care of;

- [ ] Add SVGs for every marker type
- [ ] Add hover styles to markers
- [ ] Improve selection/deselection of events
- [ ] Cross browser testing (especially SVG implementation)
- [ ] Persist marker state on refresh

localjo:
I added SVG makers for every event type we'll expect to see. The markers are similar to the icons in the list, but I added a "map pin" shape behind them and made a few minor tweaks to the iceberg and snow event icons. I'm going to need to update the icons in the list too for consistency, and I'll probably convert the list to use inline SVGs instead of CSS backgrounds. Aside from consistency, that will allow us to change the color of the icons with CSS (to create better hover states, for example).

screen shot 2017-08-16 at 5 16 01 pm

localjo:
I've been playing around with different configurations of the SVG files, and we're really limited with what we can do with external files. It would be ideal to have a way to move the SVGs inline on build. But I can't really think of an easy way to add that to our build system and still have the icons work in dev. I guess for now, leaving the SVGs as external files is best, and we can come back to the ability to include them inline when we overhaul our build system. Unless you have any ideas how we could do that now @Benjaki2

localjo:
I improved marker visibility. I think this looks pretty nice;

screen shot 2017-08-17 at 10 51 19 am

I was also thinking about exploring a brighter color scheme. When I was looking at how other maps make their markers more visible, I noticed that most markers are brightly colored. For now, I think the white backgrounds make these work. But it might be worth having another discussion about colors in the future.

localjo:
Added filtering based on current view #466 but it's still buggy on refresh and switching between tabs

localjo:
29 days and 52 commits later this is ready for cross browser testing and review!

giphy

localjo:
Feedback from our weekly meeting;

- [ ] Make animations smoother
- [ ] Fix SVGs in IE11
- [ ] Make tab toolbars fixed position
- [ ] Change cursor on marker hover
- [ ] Remove scale transform for selected markers on hover

localjo:
- [ ] Add "Read More" text to notification
- [ ] Change footer text to say "Only showing events on map."
- [ ] Fix footer layout when empty, and spacing of text

rboller:
@localjo I think the zooming action is way better now and I can't overload it with rapid-fire event selections.  

The one thing that I think needs to be updated is the "Off-map events are hidden" notice.  It's styled the same as a button and seems like you should be able to click it.  Likewise with the "Showing all active events in list" notice.

I'd advocate a styling change and/or update to what is shown.

@minniewong what do you think?

minniewong:
@localjo and @rboller - I agree the styling makes it look like a button. I'm not sure we need text saying that it's showing all the events. Also, I think the wording could be a little different - I will think about what this better wording is :)

localjo:
@rboller @minniewong The text saying that we're showing all events is there to fill the empty space in the tab's footer when the "Show All" button isn't displayed. Because of the way those tabs are built, I'd have to write some hacky JavaScript to remove the empty space. But, if you think it looks bad as it is, I can do that.

For the wording, I'm trying to choose something short that fits in the available space, because adding more space to that footer also means writing more hacky JavaScript where we don't need to. :-/ That's why I chose "Off-map events are hidden."

I'd rather not add lots of hacky code for simple layout changes, but the bottom line is how well it works for users, so I'm happy to make the tradeoff if you think it's worth it. Let me know what you think. :)

rboller:
@localjo thanks for the background.  Hopefully we wouldn't need to make the code too hacky, but usability and clarity usually are the highest priorities.

Let's see what @minniewong drafts for wording and go from there.

Benjaki2:
@localjo can you add a class like `showing-all-events` that is toggled on/off on the events tab div el and do the styling changes in CSS?

Zoom looks much better! I would consider slowing down some of the animations to give more time for the tiles to load.

localjo:
New wording and layout complete. :)

screen shot 2017-09-13 at 11 27 17 am

web/js/layers/wv.layers.sidebar.js
```diff
@@ -241,81 +250,26 @@ wv.layers.sidebar = wv.layers.sidebar || function(models, config) {
...
-      $('.ui-tabs-nav li.second')
-        .removeClass("ui-state-active");
-      $('.ui-tabs-nav li.third')
-        .removeClass("ui-state-active");
-      $footerBtns.show();
-      $footerBtns.find("button")
-        .hide();
-      $("#layers-add")
-        .show();
+      $('.ui-tabs-nav li.first').addClass("ui-state-active");
```
Benjaki2: nice simplification

web/js/naturalEvents/wv.naturalEvents.markers.js
```diff
@@ -0,0 +1,175 @@
...
+
+  // Create Overlay
+  return new ol.Overlay({
+    element: wrapper,
+    positioning: 'bottom-center',
+    id: id
+  });
+};
+
+var createBoundingBox = function(coordinates){
```
Benjaki2: CreateBoundingBox and createPin are now global. Will want to include these inside the main function.

localjo: Good catch!  😬

localjo:
Turns out that the way I'm creating markers from SVG elements has terrible performance in IE, and makes the event tab and markers take up to 30 seconds to activate. Not good. So, first I played around with writing more efficient code to generate the SVG elements, but that only reduced the time by 10-15 seconds. @ZachTRice did some research and found out that IE's performance when rendering many SVG elements is notoriously bad.

So I did an experiment to render all SVG icons as background images with CSS instead of inline, and the performance is good! The main trade off is that we can't control the icon colors with CSS, so hover states will probably just be size changes, without color changes.

buster.js
```diff
@@ -20,7 +20,7 @@ config["wv"] = {
     "web/ext/main/jquery.migrate-1.2.1/jquery-migrate.min.js",
     "web/ext/main/jscache-gitba01cdc/cache.js",
     "web/ext/proj/proj4js-2.3.3-1/proj4.js",
-    "web/ext/map/openlayers-3.17.1.wv1/ol.js",
+    "node_modules/openlayers/dist/ol.js",
```
localjo: @Benjaki2 should this be ol-debug.js?

localjo:
@Benjaki2 Can you take a look at how I'm interacting with the model? It seems to work, but I'm not 100% confident that the things I'm doing with the model follow the correct patterns.

Benjaki2:
@localjo Sure thing. Can you point me to any sections of particular concern?

localjo:
Sure @Benjaki2 

When I added an event deselection function I had to manually add a trigger to save the change to the model;
https://github.com/nasa-gibs/worldview/pull/463/commits/ce0eb0a4eef208a3363cd12c4908aa6519e1e75d#diff-a40963c60fdce3cf6e87e62c9a5da447R132

I'm not sure if that's the right approach, but it seems to work. However, the selectEvent function doesn't manually trigger a model change, because it's calling `models.date.select` which triggers a change anyway; https://github.com/nasa-gibs/worldview/blob/edf187eff7bc03d7a7f5d1499f18ce4f665ef4e2/web/js/naturalEvents/wv.naturalEvents.ui.js#L104

I don't want to trigger redundant change events that cause the model to save state multiple times when nothing has actually changed, but I'm also not sure if relying on the `models.date.select` trigger will come back to bite us in the future.

Benjaki2:
@localjo found a little css bug. This item in blue should be white. I think 😁  ?
screen shot 2017-09-18 at 1 09 37 pm

web/js/naturalEvents/wv.naturalEvents.model.js
```diff
@@ -42,25 +25,29 @@ wv.naturalEvents.model = wv.naturalEvents.model || function(models, config) {
...
     }
   };
 
   self.load = function(state) {
-    if (state.e == 't') {
-      models.wv.events.on("startup", function() {
-        wvx.ui.sidebar.selectTab("events");
+    if (!state.e) return;
+    models.wv.events.on('startup', function() {
+      wvx.ui.sidebar.selectTab('events');
```
Benjaki2: I would get rid of reference to this global `wvx` object in this file and pass a ui parameter in from `wv.main.js`

Benjaki2:
@localjo I understand your concern with redundant events but after looking at the code it seems that you've set up the triggers in the best way I can think of.

web/index.html
```diff
@@ -71,7 +71,7 @@
     <!-- link.prod --> <!-- <link type="text/css" rel="Stylesheet" href="ext/image/jcrop-0.9.12/jquery.Jcrop.min.css"> -->
 
                             <link type="text/css" rel="Stylesheet" href="ext/ui/noUiSlider-6.1.10/jquery.nouislider.css">
-                            <link type="text/css" rel="Stylesheet" href="ext/map/openlayers-3.17.1.wv1/ol.css">
+                            <link type="text/css" rel="Stylesheet" href="ext/node_modules/openlayers/dist/ol.css">
```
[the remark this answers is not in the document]
localjo: Oops, I guess I added it by mistake

localjo:
Latest changes are deployed to UAT. Going to do some more browser testing tomorrow and make sure that my fixes didn't break anything new.

localjo:
Summary of changes in this PR;

* Adds ability to see all event markers at once on the map (Fixes #407)
* Changes event selection function to use event id rather than DOM index
* Improves day selection for Storms and Wildfires
* Makes markers clickable to select events
* Improves event marker style for visibility on the map
* Adds unique SVG map markers for each event type
* Adds hover styles to event markers
* Improves precision of event marker position on map
* Adds map markers for events with a bounding box
* Improves bounding box style
* Moves selected event state into model
* Saves event selection state to URL
* Preserves selected event markers on refresh (Fixes #472)
* Allows linking to selected events
* Adds notification for links to inactive/missing events
* Sorts events in model in reverse chronological order by the event's most recent active date
* Improves animations and fixes animation glitches (Fixes #471)
* Improves order of execution during event selection (layers change after animation)
* Adds code to handle interrupted animations gracefully
* Adds filtering to event list to only show events that are visible on the map (Fixes #466)
* Adds button to show all events in a filtered list
* Adds loading indicator for events panel
* Makes tab footers unique to each tab, and fixed position, rather than calculated
* Preserves active layers when selecting a different event/date of the same type (Fixes #388)
* Improves spacing around buttons in tab footer
* Updates OpenLayers to 4.3.1 and fix broken references
* Renames naturalEvents.map to naturalEvents.markers
* Renames variables and function names for clarity
* Fixes code that removes ignored events/categories from model
* Triggers 'queryResults' event only once
* Removes unused variables and dead code
* Refactors and simplifies events code
* Moves code to correct files for better organization
* Cleans up tab UI
* Adds vector icons for events tab and tab expand/collapse arrow
* Fixes various bugs
* Enhances performance

localjo:
And, a more human readable version of the changes for release notes;

New Features

* You can now see all natural events on the map at once. (Fixed #407)
* Selected events are preserved when you refresh the app and you can share links to selected events using the `e` parameter in the URL with the format `?e={event_id},{event_date}`, for example, `?e=EONET_2874,2014-07-24`. (Fixed #472)
* The events list is now automatically filtered to only show events that are visible in the current map view, and a "List All" button has been added to see the full list of active events. (Fixes #466)
* You can now select events by clicking on their marker on the map.
* Events with a bounding box now have a marker in addition to the bounding box.
* Event marker styles have been improved; each event category now has a unique marker. Markers have been given an outline to make them more visible on the map, and they now use cleaner SVG icons instead of raster images.
* Event bounding box styles have been improved to be more visible, cleaner, and more aesthetically pleasing.

Fixes and Minor Changes

* Active layers are now preserved when selecting a different event/date in the same category. (Fixed #388)
* Animations are smoother and some bugs have been worked out. (Fixed #471)
* The event list is now sorted more reliably in reverse chronological order.
* The default date selection for Storms and Wildfires has been improved.
* Event markers are now positioned more precisely on the map.
* A warning notification has been added for links to inactive/missing events.
* A loading indicator has been added to the events panel, since the events may load slowly in some circumstances.
* Spacing around the buttons at the bottom of tabs has been improved.
* OpenLayers was updated to version 4.3.1.
* Vector icons were added for the events tab and the tab expand/collapse arrows.
* Extensive bug fixes, performance enhancements, and refactoring.

PR #476, opened 2017-08-18
Check localStorage safely

### LINKED ISSUE #474: LocalStorage sometimes causes crash at startup (especially in Safari)
It appears that certain browser plugins may cause a crash at startup.  In particular, a user reported a problem which had the following console error which we should handle more gracefully:

`TypeError: null is not an object (evaluating 'localStorage.getItem')`

localjo:
Fixes #474

### DIFF
--- a/test/proj/wv.proj.change-test.js
+++ b/test/proj/wv.proj.change-test.js
@@ -158,7 +158,7 @@ buster.testCase("wv.proj.change", {
   },
 
   "Not notified if 'Dont show again' selected": function() {
-    this.stub(wv.util, "localStorage")
+    this.stub(window.localStorage, 'getItem')
       .withArgs("arcticProjectionChangeNotification")
       .returns("true");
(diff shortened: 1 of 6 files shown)


localjo:
This fix is ready if it looks good to you guys. I'm not sure how we want to go about releasing it, and if the patch version number bump should happen before or after it's merged, so let me know if you need anything more from me to get this ready. I'm pretty sure one of you will have to deploy it though since I don't know the whole deployment process yet and I don't want to break anything.

localjo:
Aside from you guys reviewing this, we should probably check that everything works as expected in other browsers. Do we have a Browserstack account we can use to test?

Benjaki2:
Looks good! but probably should be tested with active events. (https://github.com/nasa-gibs/worldview-options-eosdis/blob/master/common/features.json#L15)[Replace with test api link] : `https://status.earthdata.nasa.gov/api/v1/notifications?domain=https%3A%2F%2Fworldview.uat.earthdata.nasa.gov`

Issue #482, opened 2017-08-24
notification module bug


Benjaki2:
* alert number misalignment
* message-type not being received

PR #483, opened 2017-08-24
Notification patch

Benjaki2:
@localjo Added localStorage logic to the init as this component is dependent on LocalStorage
Connects to #482

### DIFF
--- a/web/css/wv.alert.css
+++ b/web/css/wv.alert.css
@@ -66,6 +66,7 @@
   font-size: 12px;
   right: 10px;
   top: 4px;
+  z-index: 33;
 }
 
 .ui-menu-item.exclamation-circle a::before {
--- a/web/css/wv.alert.css
+++ b/web/css/wv.alert.css
@@ -65,6 +65,7 @@
   position: absolute;
   font-size: 12px;
   right: 10px;
+  top: 4px;
 }
 
 .ui-menu-item.exclamation-circle a::before {
--- a/web/js/notifications/wv.notifications.ui.js
+++ b/web/js/notifications/wv.notifications.ui.js
@@ -58,6 +58,10 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
    */
   var init = function() {
     var reactComponent, options, p, alertUser;
+    if(!wv.util.browser.localStorage) {
+      console.warn('User does not have localStorage');
+      return;
+    }
     mainIcon = $('#wv-info-button')[0];
     mainIconLabel = $('#wv-info-button label')[0];
     p = wv.util.get(url);
@@ -95,7 +99,6 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
     if (message) {
       messageBlockExists = true;
     }
-
     if (message && !objectAlreadySeen(message)) {
       priority = 'message';
       mainNotification = 'message';
@@ -154,7 +157,7 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
    *  in LocalStorage
    */
   var getNumberOfTypeNotseen = function(type, arra) {
-    var storageItem = wv.util.browser.localStorage?!!localStorage.getItem(type):false;
+    var storageItem = localStorage.getItem(type);
     var count, len;
 
     len = arra.length;
@@ -188,9 +191,9 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
 
     type = obj.notification_type;
     idString = obj.created_at.toString();
-    fieldExists = wv.util.browser.localStorage?!!localStorage.getItem(type):false;
+    fieldExists = false;
     fieldValueMatches = false;
-
+    fieldExists = localStorage.getItem(type);
     if (fieldExists) {
       fieldValueMatches = localStorageValueMatches(type, idString);
     }
@@ -346,9 +349,10 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
     var messages;
     this.className = 'ui-icon fa fa-fw fa-gift';
     self.messageIconActive = false;
-    if (wv.util.browser.localStorage) localStorage.setItem('message', activeMessageId);
+    if (activeMessageId) {
+      localStorage.setItem('message', activeMessageId);
+    }
     activeMessageId = null;
-
     messages = sortedNotifications.messages[0];
     if (messages) {
       create$whatsNew(messages, 'Latest Release');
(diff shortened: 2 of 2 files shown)


web/js/notifications/wv.notifications.ui.js
```diff
@@ -151,7 +157,7 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
    *  in LocalStorage
    */
   var getNumberOfTypeNotseen = function(type, arra) {
-    var storageItem = wv.util.browser.localStorage?!!localStorage.getItem(type):false;
+    var storageItem = localStorage.getItem(type);
```
localjo: You need to check `wv.util.browser.localStorage` here before you use `localStorage`. This was the problem that was causing a crash before.

web/js/notifications/wv.notifications.ui.js
```diff
@@ -185,9 +191,9 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
 
     type = obj.notification_type;
     idString = obj.created_at.toString();
-    fieldExists = wv.util.browser.localStorage?!!localStorage.getItem(type):false;
+    fieldExists = false;
     fieldValueMatches = false;
-
+    fieldExists = localStorage.getItem(type);
```
localjo: You need to check `wv.util.browser.localStorage` here before you use `localStorage`. That can either be with a ternary operator, like this;
`wv.util.browser.localStorage?!!localStorage.getItem(type):false;`
Or with two lines, like this;
```
fieldExists = false;
if (wv.util.browser.localStorage) fieldExists = !!localStorage.getItem(type);
```
I prefer the ternary operator, since it's shorter.

web/js/notifications/wv.notifications.ui.js
```diff
@@ -344,9 +349,10 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
     var messages;
     this.className = 'ui-icon fa fa-fw fa-gift';
     self.messageIconActive = false;
-    if (wv.util.browser.localStorage) localStorage.setItem('message', activeMessageId);
+    if (activeMessageId) {
+      localStorage.setItem('message', activeMessageId);
```
localjo: Need to safely use `localStorage`. This should be;
```
if (activeMessageId && wv.util.browser.localStorage) localStorage.setItem('message', activeMessageId);
```

web/js/notifications/wv.notifications.ui.js
```diff
@@ -372,11 +378,12 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
     self.infoIconActive = false;
     self.notifyIconActive = false;
     mainNotification = null;
+
     createNotifyDialog();
-    if (activeNotifications.outage && wv.util.browser.localStorage) {
+    if (activeNotifications.outage ) {
```
localjo: You need to leave the `wv.util.browser.localStorage` check here.

web/js/notifications/wv.notifications.ui.js
```diff
@@ -372,11 +378,12 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
...
       localStorage.setItem('outage', activeNotifications.outage);
     }
-    if (activeNotifications.alert && wv.util.browser.localStorage) {
+    if (activeNotifications.alert) {
```
localjo: You need to leave the `wv.util.browser.localStorage` check here.

web/js/notifications/wv.notifications.ui.js
```diff
@@ -486,7 +493,6 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
    * @returns {Boolean}
    */
   var localStorageValueMatches = function(property, value) {
-    if (!wv.util.browser.localStorage) return;
     var oldValue = localStorage.getItem(property);
```
localjo: You need to leave the `wv.util.browser.localStorage` check here.

web/css/wv.alert.css
```diff
@@ -65,6 +65,7 @@
   position: absolute;
   font-size: 12px;
   right: 10px;
+  top: 4px;
```
ZachTRice: It might be a good idea to add a z-index here; something like z-index: 33; only because when I looked at this the other day, if the icon was overlaying another element it was partially hidden unless the z-index was set. Just a bit of a safety rule to ensure the element always appears on top.

web/js/notifications/wv.notifications.ui.js
localjo: See #474 and #476 for why these checks are here.

web/js/notifications/wv.notifications.ui.js
```diff
@@ -58,6 +58,10 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
    */
   var init = function() {
     var reactComponent, options, p, alertUser;
+    if(!wv.util.browser.localStorage) {
```
Benjaki2: @localjo disabled this module here if localstorage isn't present. As this module is pointless without it.

localjo: Ah, I see. My bad! The rest of my comments weren't relative then. 🙊 I've removed them, but left two other comments.

web/js/notifications/wv.notifications.ui.js
```diff
@@ -58,6 +58,10 @@ wv.notifications.ui = wv.notifications.ui || function(models, config) {
...
+      console.warn('User does not have localStorage');
```
localjo: `console.warn` can cause some older browsers to crash, we might want to leave it out, and just return in one line;
```
if(!wv.util.browser.localStorage) return;
```

Benjaki2:
@localjo Good call. I will take out the warn.

I am pro brackets and against tertiary conditionals for the sake of readability.. But open to change that if it's the consensus @localjo @ZachTRice . We should add whatever we determine to the `style-guide`

@ZachTRice  I will add a z-index for safety, thanks

ZachTRice:
I lean towards brackets in js as well; I don't have a good reasony why but it just feels better to me. I prefer ternary operators when I write php so maybe I could get use to it. I never write nested ternary operators anywhere though.

localjo:
Yeah, it's totally a question of style, and I usually like to use brackets, except for quick checks where you're doing something simple like a return. I think this;
```
if (!value) return;
```
Is way more readable, and less noisy than;
```
if (!value) {
  return;
}
```
As a general rule of thumb, if a whole if statement fits in an 80 character width line and isn't cryptic, I leave the brackets out to make the code easier to read quickly.

With ternary statements, I try to make sure they're not too cryptic, and if they get longer than 80 characters, I try to break them up into more lines as well. I'd prefer;
````
id = value.match(regEx) ? value : null;
```
Rather than;
```
if (value.match(regEx) {
  id = value;
} else {
  id = null;
}
```
My general rule of thumb is what's easier to read and scan quickly. Too many lines and blocks for simple things can make simple code appear more complex than it needs to be. Of course, these are just style preferences, so I don't really have a strong opinion about them for this PR. Eventually we should come to a consensus and add linting rules for it, but we've got a long way to go before we can bring all of our code in line with any particular standard. When we do eventually do that, I'd recommend we stick to a standard styleguide rather than spending a bunch of time discussing preferences. https://standardjs.com/

minniewong:
Fixes #482

minniewong:
Connects to #482

Issue #477, opened 2017-08-21
Orbit track descriptions are incorrect in the search results and layer list


minniewong:
![screen shot 2017-08-21 at 4 02 49 pm](https://user-images.githubusercontent.com/10747923/29536366-87662542-868b-11e7-9669-0e63bcbe93a5.png)

When looking at the orbit track descriptions using search and in the layer list, the orbit tracks have the incorrect layer description (except for Suomi NPP). When using the product picker, the descriptions are correct for all of the orbit tracks.

ZachTRice:
The orbit tracks are actually attached to multiple sources so when it it looping through, it is only grabbing one of the source descriptions. I'll work with it more to see if I can attach each of the sources it is attached to but even then it will be hard to tell which source you selected orbit tracks within since all descriptions would be lumped together. 

Another separate issue I noticed is that if I check and add the Aerosol Optic Depth - Terra/Modis - Desc/Day orbit track, it is not checked in Cloud Top Temperature - Terra/Modis - Desc/Day; even though these are the same orbit tracks. Manually selecting both and then unselecting one does unselect both though so the expected results are inconsistent.

PR #484, opened 2017-08-24
Orbit description fix

ZachTRice:
Fixes the issue of orbit tracks displaying the wrong measurement descriptions.

I would recommend merging this after merging the [orbit-track-references](https://github.com/nasa-gibs/worldview-options-eosdis/pull/109) PR  in the options repo since some of the code references new parameters; though everything should gracefully fallback regardless.

self.addDescriptions is a big, not so nice looking function providing a lot of fallbacks at the moment. Eventually it can be removed once we have individual layer descriptions in place. Right now it specifically filters out Orbital Tracks to get the right measurement descriptions associated with the orbit layers, then it is also is providing a check to see if a description field is set in the layer file, if not then associate it with the measurement description.
Connects to #477

### DIFF
--- a/web/js/util/wv.util.js
+++ b/web/js/util/wv.util.js
@@ -807,6 +807,11 @@ wv.util = (function(self) {
     return value;
   };
 
+
+  self.capitalizeFirstLetter = function(string) {
+    return string.charAt(0).toUpperCase() + string.slice(1);
+  };
+
   return self;
 
 })(wv.util || {});
--- a/web/js/layers/wv.layers.modal.js
+++ b/web/js/layers/wv.layers.modal.js
@@ -508,8 +508,23 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
             .append($label);
 
           // If this is an orbit track.... put it in the orbit track list
-          if (layer.title.indexOf("Orbital Track") !== -1) {
+          if (layer.layergroup && layer.layergroup.indexOf("reference_orbits") !== -1) {
             var orbitTitle;
+
+            if (layer.time && layer.track) {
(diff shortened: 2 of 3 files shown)


Benjaki2:
glad to hear you are thinking about Destroying 💥  that `self.addDescriptions` function. All those nested loops make me dizzy

web/js/util/wv.util.js
```diff
@@ -807,6 +807,11 @@ wv.util = (function(self) {
     return value;
   };
 
+
+  self.capitalizeFirstLetter = function(string) {
```
Benjaki2: could also use lodash's `_.startCase(_.toLower(str));`
stolen from [stackoverflow](https://stackoverflow.com/a/38084493)

localjo: I don't think we should add this as a utility function. It adds more noise and complexity to this file, and makes the code harder to follow. This is a pretty simple operation. We can just do this in `wv.layers.modal.js`;
```
orbitTitle = _.startCase(layer.daynight)  + "/" + _.startCase(layer.track);
```
That's shorter and more clear anyway.

localjo: Also, what case is the string in to begin with?

ZachTRice: Agreed. This makes a lot of sense. I will remove the function and just use it inline since it is so short. <3 lodash

ZachTRice: It is start case. The Daynight and Track need to start with first letter uppercased. _.startCase should be the correct function to use here. Ben's function above makes sure everything is lowercase first before converting to the start case but I don't think all that is needed.

localjo: It's already start case to begin with? Why do we need to convert it?

ZachTRice: Negative. They are lowercase values to begin with. I meant that they need to be start case.

Issue #439, opened 2017-07-28
Replace Grunt scripts with NPM scripts


Benjaki2:
* Gruntfile is currently ~800 lines of code that does too much in one place
* We can break these tasks into separate bash scripts inside a task folder.
* Won't rely on grunt/gulp wrappers which will remove unneeded 3rd party dependancies

localjo:
Happy to help with this

ZachTRice:
Since we will be converting this soon, I'm going to add additional tasks here that we need which haven't been added to grunt.

- Clean up the web/config/metadata/** markdown files after they have been converted.
`remove:markdown: ["web/config/metadata/**/*.md"],`
This can be added to the config process after the options-repo markdown files have been converted to html.

localjo:
One thing that would be nice when we tackle this is to add a progress indicator to the GIB GetCapabilities request, since that can take quite a long time and it's not clear if progress is being made or not.

localjo:
I've spent a lot of time looking at our build scripts this week, and formed a clearer picture of what we've got going on here;

* We've got a complex mix of Grunt, Bash and Python scripts running build tasks. They're all interdependent but the meat of the work is done by the Python scripts.
* The build tasks are (sort of) divided into three stages;
  1) Build the options repo and make a request to GetCapabilities
  2) Build assets for the Worldview repo
  3) Combine the results of the previous two builds into one
* There's a lot of unnecessary storage of intermediate files on the file system that can be removed to simplify the process
* We're using several different interfaces to invoke the build scripts; sometimes we're running things from `npm` scripts, sometimes we're running the same thing via `grunt`, sometimes there's a custom bash script that runs the build scripts in strange orders. We should aim to simplify all of this to have a straightforward `npm run build` command that does everything simply.

I've done some of this work in https://github.com/nasa-gibs/worldview/pull/647 to move us in the right direction, but there's still a lot of work to do.

localjo:
We can also relocate `/web/build` to `/build` and serve that in development, which will serve to speed up our development process by allowing us to see the final results of the build with string replacements, and enable hot reloading of JavaScript, CSS and HTML changes.

localjo:
This may be the time to address #661 

After #503 is merged, we should rearrange all of our build scripts into a single directory, instead of having them spread between root, `/bin`, `/tasks`, and `/deploy`, etc.

localjo:
[...]

PR #488, opened 2017-08-28
Added npm task for each grunt task. Updated developing documentation.

ZachTRice:
Connects to #439
Added npm task for each grunt task. 
Updated developing documentation.
Added .npmrc file to supress npm ERR! on fail
Converted `eslint`, `stylelint`, `stylefmt`, `autoprefix` grunt commands into pure npm run commands.


ZachTRice:
This PR is to get the ball rolling on converting the grunt scripts. Basically for each grunt script I created a npm script with the same name and attached the current grunt command to it. This way we can start using `npm run [command_name]` now (even though they still point to the grunt command) and start updating each one incrementally to the pure npm method, removing the grunt dependencies as we go.

.npmrc
```diff
@@ -0,0 +1,3 @@
+{
+  loglevel=silent
```
localjo: Sometimes npm errors can be useful. Possible `loglevel` values are "silent", "error", "warn", "notice", "http", "timing", "info", "verbose", "silly". I think we should leave the loglevel at `warn` or higher to avoid suppressing useful information about problems. A user can always run an npm script with the `-s` or `--silent` flags to hide errors, and if there's a particular script that we want to suppress warnings for, we could do something like `"test":"npm run lint -s && npm run e2e -s"`.

ZachTRice: There were issues with using the silent flag which I can't remember now.  .npmrc is the most consistent way to control logging. I will need to revisit this and see what we can do because when running linters, it was really hard to see what the issue was because of all the npm error messages.

localjo: Which version of npm are you using? Updating to node v8.2.1 and npm v5.4.2 makes the default error messages much shorter.

There's a discussion about this on the npm repo with a lot of potential workarounds: https://github.com/npm/npm/issues/8821

The final solution was an update to npm which makes the log messages shorter as of npm v5.4.2: https://github.com/npm/npm/commit/65b9943e9424c67547b0029f02b0258e35ba7d26 

We should add an `npm` key with `5.4.2` in the `engines` field in package.json, and then update `version-check.js` to also check the npm version.

localjo: Here's an example after the update;
screen shot 2017-09-27 at 10 46 14 am

package.json
```diff
@@ -11,14 +11,35 @@
...
+    "e2e": "grunt e2e",
+    "default": "grunt default"
   },
   "devDependencies": {
     "autoprefixer": "7.1.1",
     "babel-polyfill": "6.16.0",
     "bluebird": "3.4.6",
     "buster": "0.7.12",
     "chromedriver": "2.25.1",
+    "cssnano": "^4.0.0-rc.2",
```
localjo: Where is this used?

ZachTRice: No where yet. It is intended to replace `grunt-contrib-cssmin`

.npmrc
ZachTRice: I am using npm 3.8.10 as this is the default that ships with the node.js long term stable installer. Yep, I read through that discussion as well. Those messages look better in npm 5. But, I think we should stick with supporting what is included in the default nodejs lts install.

localjo: LTS is the "Long Term Support" version, not "Long Term Stable", the current version of node, 8.6.0, is stable and unless we have a compliance reason forcing us to stick to the older version, I don't think we should limit ourselves to that.  https://stackoverflow.com/questions/33661274/what-are-the-differences-between-long-term-support-lts-and-stable-versions-of

localjo: Worldview barely uses Node, since it's all client side code, so we shouldn't worry too much about Node updates breaking things anyway, since it would mostly affect developers and not users.

ZachTRice: I personally wouldn't mind upgrading to 5. I wanted to use it to link the semver of the options repo but that wasn't a very good reason to upgrade since I can link it to the tagged version with 3. I just worry about upgrading npm and making it harder for users because of the fact that the default lts has 3 and most people will install that first and probably never upgrade.

ZachTRice: Keep in mind upgrading node and npm on Windows is not as straight forward as Mac. (Though it's not rocket science either). 
See: https://stackoverflow.com/questions/18412129/how-do-i-update-node-and-npm-on-windows

I personally use nvm for windows: https://github.com/coreybutler/nvm-windows
But that's also not as straight forward as using `nvm` on Mac.

localjo: Our `engines` field and `version-check.js` are just recommendations, and not a hard requirement. I don't see any reason why our app isn't backwards compatible for developers who have an older version of Node installed. But, just in case, we use the version check to warn developers on older versions that they may have to update Node *if* they run into any problems. Since our Node dependencies are minimal, I doubt anyone would run into problems. But asking a developer to update Node if they do run into a problem doesn't seem like a big deal to me.

localjo: The only real risk (and I'd argue that it's still very small) is that us even using a newer version of Node while developing (even without making explicit changes to the engines/version check) could end up with us writing code that depends on newer Node features without realizing it which then don't work for developers who have an older version of Node installed. The best way to make sure that we don't run into that is to make sure that the `engines` fields have the version numbers that we're actually using in development.

The engines fields and version checks are basically saying "Hey, these are the versions we're using, so if you have any problems, try using these too."

PR #498, opened 2017-09-05
Webpack Proof of Concept

localjo:
Some initial groundwork on adding support for module loaders using Webpack


localjo:
Other things we'd want to do to move this forward:
- [ ] Move legacy vendor dependencies to one `/vendor` directory
- [ ] Switch to dependencies in package.json instead of local, when possible
- [ ] Remove reliance on moving/copying Grunt scripts (two items above should move us toward this)
- [ ] Migrate dependencies from `script-loader` to a loader that supports `import` and ES2015+
- [ ] Clean up directory structure for build/dependencies
- [ ] Determine the best approach for adding and including new dependencies in an organized way

localjo:
The command to run Webpack is `./node_modules/.bin/webpack`, eventually we'd want to integrate that with the start script and take advantage of the Webpack dev server and hot reloading.

localjo:
May come back to this after we modularize our existing codebase.

Issue #267, opened 2017-05-05
Add test for removed layers


minniewong:
If something is removed from the layerorder.json and not removed from one of the json files in the measurements folder it causes the product picker to not work for the category the measurement is in.

ZachTRice:
I may have addressed this in my polar-picker branch when rewriting the layer modal functionality. I was seeing this happening when I was testing with the gibs-specific-options repo. I can do some testing and look into this more.

ZachTRice:
The polar-picker branch addresses this issue. If a layer is removed from the layerOrder file, it will be removed from the measurement & layer views. If all layers within a measurement are removed, i.e. all 3 Blue Marble layers are removed, then that (empty) measurement won't be output in the categories view or the measurements view.

PR #510, opened 2017-09-12
Change polar product picker to list of measurements

ZachTRice:
Connects to #327 
Connects to #332 
Connects to #267 
Arctic and Antarctic projections now display measurement list instead of layer list.
Refactored layers modal code. 
Removed calls to specific projections, checked current view instead.
Adjusted breadcrumb text
Better filtering for Orbit tracks. 
Updated orbital track filtering function to use new options repo flags.
Added modalView URL flag with `layers`, `measurements`, and `categories` parameter options to force changing the current projection view.
Added console warn when setting modalView
Added modalView to Testing/Debugging documentation


web/js/layers/wv.layers.modal.js
```diff
@@ -120,458 +115,406 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
 
-    var $dotContinueLink = $('<a></a>')
-      .attr('data-category', category.id)
-      .text('...');
+          $categoryTitle.append($categoryLink);
+          $categoryOpaque.append($categoryTitle);
+
+          var $measurements = $('<ul></ul>');
+          $i = 0;
+          _.each(category.measurements, function(measurement, index) {
```
Benjaki2: @ZachTRice looks like a similar nested loop is being reused a few times. I think you could throw this into a function and reuse.

ZachTRice: I have updated the redundant part of the function which is just inside the code snippet you highlighted. That beginning `_.each` category loop is only used once but the source setting check inside is used multiple time which is what I changed.

web/js/layers/wv.layers.modal.js
```diff
@@ -32,85 +32,105 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+
+  // Checks to see if the current view has sources with settings.
+  var checkSourceSettings = function(current) {
+    _.each( current.sources, function( source, sourceName ) {
+      checkSettings(current, source);
+    });
+  };
+
+  var checkModalView = function() {
+    if (config.parameters.modalView == 'categories') {
```
Benjaki2: Good opportunity for switch statement

web/js/layers/wv.layers.modal.js
```diff
@@ -32,85 +32,105 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+  //Create container for 'by interest' filters buttons
+  var $nav = $('<nav></nav>').attr('id', 'categories-nav');
 
-  // Create container for breadcrumb
-  var $breadcrumb = $('<nav></nav>')
-    .attr('id', 'category-breadcrumb');
+  //Create container for breadcrumb
+  var $breadcrumb = $('<nav></nav>').attr('id', 'category-breadcrumb');
+
+  var checkSettings = function(current, source) {
```
Benjaki2: 👍

web/js/layers/wv.layers.modal.js
Benjaki2: when using the same chained object many times within a function it can be more efficient to set it as a var at the top of the func `eg : var modalView  = config.parameters.modalview`, but if the switch was in place it wouldn't be necessary cause the chained value would only be referenced once.

web/js/layers/wv.layers.modal.js
```diff
@@ -32,85 +32,105 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
   var $allLayers = $(self.selector + " #layers-all");
   var gridItemWidth = 320; //with of grid item + spacing
+  var projection = models.proj.selected.id;
   var modalHeight;
   var sizeMultiplier;
   var searchBool;
+  var hasSetting;
 
-  // Visible Layers
+  //Visible Layers
```
localjo: It's a good idea to avoid whitespace changes when possible, since they make it harder to figure out when the last meaningful change was made using `git blame`.

ZachTRice: Definitely. Hm, this must have been caused by an autofixer somewhere; probably after the switch from 4 to 2 spaces. I prefer the space after the `//` so I wouldn't have changed this willingly. I'm going to do another round or two of code cleanup so I'll go through and revert these before my next commit.

Benjaki2:
Ideally, try to add comments whenever you add new methods but the `drawMeasurements` and `drawCategories` methods definitely need some comments as they are so long. If you can break those methods up to be smaller methods that are easier to read/understand, do so. Otherwise just add comments to help describe whats going on.

web/js/layers/wv.layers.modal.js
```diff
@@ -120,458 +140,350 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+          if (category.placement) {
+            if (category.placement === 'first') {
+              sortNumber = 1;
+            } else if (category.placement === 'last') {
+              sortNumber = 3;
+            }
+          } else {
+            sortNumber = 2;
+          }
+          var $category = $('<div></div>').attr('data-sort', sortNumber).addClass('layer-category layer-category-' + interestCssName(metaCategoryName)).attr('id', category.id);
```
localjo: You can create a new element with the attributes you need, rather than chaining extra function calls on the end;
```
$('<div />', {
  class: 'layer-category layer-category-' + interestCssName(metaCategoryName)),
  id: category.id,
  'data-sort': sortNumber
});
```
It's more readable, and more efficient that way.

web/js/layers/wv.layers.modal.js
```diff
@@ -120,458 +140,350 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+          }
 
-  var cssName = function(name) {
-    if (name === 'hazards and disasters') {
-      return 'legacy';
-    } else {
-      return name;
-    }
-  };
+          var $categoryOpaque = $('<div></div>').addClass('category-background-cover');
```
localjo: This can be `$('<div/>', {class: 'category-background-cover'});`

web/js/layers/wv.layers.modal.js
```diff
@@ -120,458 +140,350 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
 
-  var setCategoryOverflow = function(category, $measurements) {
-    var $dotContinueItem = $('<li></li>')
-      .addClass('layer-category-item');
+          var $categoryTitle = $('<h3></h3>');
 
-    var $dotContinueLink = $('<a></a>')
-      .attr('data-category', category.id)
-      .text('...');
+          var $categoryLink = $('<a></a>').text(category.title).attr('alt', category.title).addClass('layer-category-name').click(function(e) {
```
localjo:
```
$('<a/>', {
  text: category.title,
  alt: category.title,
  class: 'layer-category-name',
  click: drawMeasurements(category)
});
```

web/js/layers/wv.layers.modal.js
```diff
@@ -32,85 +32,105 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+  var $breadcrumb = $('<nav></nav>').attr('id', 'category-breadcrumb');
+
+  var checkSettings = function(current, source) {
+    _.each( source.settings, function( setting ) {
+      var layer = config.layers[setting];
+      var proj = layer.projections;
+      if(layer.id == setting && Object.keys(proj).indexOf(projection) > -1) {
+        if (layer.layergroup && layer.layergroup.indexOf("reference_orbits") !== -1) {
+          if(current.id === "orbital-track") {
+            hasSetting = true;
```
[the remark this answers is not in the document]
ZachTRice: Good call. I moved the variable out of the parent scope and into the function; looking here to determine whether to return true. This function loops through the passed in source's settings and detects if it has a setting. if it does then hasSetting is true and the function now returns true.

web/js/layers/wv.layers.modal.js
```diff
@@ -53,11 +53,8 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
     // Create tiles
     render();
 
-    $addBtn.click(function(e) {
-      $(self.selector).dialog("open");
```
localjo: Make sure the add layers button still works.

ZachTRice: I checked it out locally after the merge and it looks to be working fine.

web/js/layers/wv.layers.modal.js
```diff
@@ -62,14 +62,23 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+  // Create container for breadcrumb
   var $breadcrumb = $('<nav></nav>').attr('id', 'category-breadcrumb');
 
-  var checkSettings = function(current, source) {
-    _.each( source.settings, function( setting ) {
+  /**
+   * var checkMeasurementSettings - Checks the (current) measurement's source
+   *  for a setting and returns true if present.
+   *
+   * @param  {type} current The current config.measurements measurement.
```
Benjaki2: @ZachTRice `type` should be replaced by parameter object type `String, boolean, Array, etc`
```
* @param {Object} source The current measurement source.
```
Also looks like the `current` parameter isn't used

web/js/layers/wv.layers.modal.js
```diff
@@ -57,54 +58,117 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+              hasSetting = true;
+            }
+          // Don't output sources with only orbit tracks
+          } else {
+            hasSetting = true;
+          }
+        }
+      }
+    });
+    return hasSetting ? true : false;
```
localjo: Since this is already a boolean, you can just do `return hasSetting;`

web/js/layers/wv.layers.modal.js
```diff
@@ -57,54 +58,117 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+   *  when drawing categories.
+   *
+   * @param  {type} current The current config.measurements measurement.
+   * @return {type}         Return true if the measurement has sources with settings.
+   */
+
+  var hasMeasurementSource = function(current) {
+    var hasSource;
+    Object.values(current.sources).forEach(function(source) {
+      if(hasMeasurementSetting(current, source)) { hasSource = true; hasMeasurement = true; };
```
localjo: I think `hasMeasurement` is unused, so I'd simplify to;
`hasSource = hasMeasurementSetting(current, source);`

web/js/layers/wv.layers.modal.js
```diff
@@ -114,458 +178,443 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+                  if (layer.daynight && layer.track) {
+                    orbitTitle = _.startCase(layer.track) + "/" + _.startCase(layer.daynight);
+                  }
+
+                  $label.empty()
+                    .text(orbitTitle);
+                  $sourceOrbits.append($wrapper);
+
+                /**
+                 * @deprecated since version 1.8.0 If the data set is old and doesn't have
```
localjo: It's unclear what's "deprecated" here.

Issue #512, opened 2017-09-12
Add `name` to package.json


ZachTRice:
`name` is a required package.json field but it is missing from our file.

PR #513, opened 2017-09-12
Added name field to package.json

ZachTRice:
Connects to #512

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "worldview",
   "version": "1.8.0",
   "release": "1",
   "repository": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "worldview",
```
localjo: We're not installing this as a dependency, so it might not matter, but the package name [`worldview` is already taken on npm](https://www.npmjs.com/package/worldview). Could we use `nasa-worldview` instead?

ZachTRice: Gah, yea we better make it unique.

ZachTRice:
Changed has been added in #517

PR #526, opened 2017-09-21
E2E Testing Improvements [WIP]

### LINKED ISSUE #454: Add Browserstack automated testing
@localjo mentioned automated browserstack testing in conversation. I'm making a ticket to look at getting this integrated at some point. Might be something to  integrate when we transition away from busterjs #96 

Nightwatch, Bamboo and Travis CI are supported.

https://www.browserstack.com/automate/nightwatch
https://www.browserstack.com/automate/bamboo
https://www.browserstack.com/automate/travisci

localjo:
This is a work in progresss

- [x] Adds `nightwatch-cucumber` for more readable test specs
- [x] Adds HTML format test reports
- [x] Moves e2e testing out of Grunt and into npm scripts
- [x] Migrates all tests to use Cucumber
- [x] Adds support for running tests on Browserstack
- [x] Improves README/CONTRIBUTING docs
- [x] CORS policy problem resolved (see #454)
- [x] Adds tests for natural events feature
- [x] Use browserstack-capabilities

Resolves #454 

Known issues:
* Mouse-over based tests are notoriously unreliable. They're currently skipped with `--skiptags hover` on the npm scripts, but they can be run by removing the tag. Once https://github.com/cucumber/cucumber-js/issues/727 is resolved, we should remove the skip, and just have them retry a few times.
* The CORS policy issue prevents most tests from passing on Browserstack at the moment.


localjo:
This commit migrates one of our existing e2e tests to Cucumber:https://github.com/nasa-gibs/worldview/pull/526/commits/0f963241b521e278afb1ff1966ee9e3708c9ad12

The migrations will get progressively easier as redundant code has been implemented in Cucumber. Here's a screenshot of the report for the tests migrated in that commit:

screen shot 2017-09-21 at 5 34 04 pm

localjo:
Need to come up with a way to run tests for different browsers in parallel, while filtering some tests out for certain browsers and generating separate reports for each browser. See: https://github.com/mucsi96/nightwatch-cucumber/issues/256

localjo:
All tests are migrated to Cucumber, with only one failing in Chrome. screen shot 2017-09-26 at 2 33 36 pm

Tests are running in Firefox, but tests that involve `mouseover`/`hover` events are skipped due to the WebDriver in Firefox not fully supporting those events. We can find a workaround for this in the future.

localjo:
@Benjaki2 @ZachTRice We should be able to test a local install of Worldview via Browserstack fairly easily: https://www.browserstack.com/automate/nightwatch

I'm going to work on setting this up.

We could probably even make the tests run against SIT/UAT fairly easily as well. In the future, maybe we can even figure out how to automate a deploy to SIT/UAT every time we push a new commit to a PR, and then the whole process would be automated.

ZachTRice:
We can probably use something like Husky for the git hooks: https://github.com/typicode/husky. As far as deploying to sit/uat every time, I'm not sure how that might work since that requires VPN access and it would probably need to be hidden in our repo so others couldn't discover this.

localjo:
I added instructions for the End-to-end tests:
https://github.com/nasa-gibs/worldview/blob/events-e2e/CONTRIBUTING.md#end-to-end-tests

These should get simpler once we fix the CORS issue and if we figure out a solution for hover-based tests.

For now, if you just want to run the e2e tests locally so you can add new tests, the easiest thing to do is just to run `npm run e2e:chrome` which should work out of the box (after a fresh `npm install`).

localjo:
I'd like to use this to easily define which browsers we want to test: https://www.npmjs.com/package/browserstack-capabilities

localjo:
I finished adding tests for Natural Events that cover the equivalent of the Test Plan and a few tests for the new features.

Test coverage needs to be improved, but in order to do that, I think we need to mock the API responses we're dealing with, since it's hard to test precisely against a moving target. Mocking API responses will help us to create better tests for other features as well. However, I think we should approach that as a separate issue and circle back to it, since it's a significant amount of work that doesn't need to hold up the other things we're working on at the moment. @Benjaki2 @rboller @ZachTRice 

Aside from that, I need to clean up some of the code for running these tests on Browserstack and then this will be ready for review. That should happen in the first half of tomorrow, assuming nothing else distracts me in the meantime.

localjo:
A majority of Browserstack tests are failing, and it's not clear why. Looks like a problem with the runner, but I haven't figured it out yet.

The output it also a little bit too noisy to be useful, but I haven't figured out how to make that quieter yet either. It seems that the output options for `nightwatch-cucumber` don't work as documented, same for `nightwatch`.

ZachTRice:
Have you explored running chrome in headless mode for these tests? Looks like nightwatch supports it with:
```
desiredCapabilities: {
    "chromeOptions" : {
        "args" : ["headless"]
    }
}
```
https://github.com/nasa-gibs/worldview/blob/events-e2e/e2e/browserstack.conf.js#L33-L43

Also these would need to be updated.
https://github.com/nasa-gibs/worldview/blob/7c445d8b97ba60fe3654163254b37f39940a3942/doc/developing.md#unit-tests

PR #546, opened 2017-09-25
UI Bug fixes

localjo:
This fixes a bug where events were not selected when clicking on event markers in some browsers


localjo:
Also fixes sizing/scrollbars on data tab

PR #574, opened 2017-10-17
Search opt patch

Benjaki2:


web/js/layers/wv.layers.modal.js
```diff
@@ -28,24 +28,13 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
   var searchBool;
   var hasMeasurement;
   var copy = [];
+  var metaLoaded = false;
```
localjo: What do you think of renaming this to `metadataLoaded`?

web/js/layers/wv.layers.modal.js
```diff
@@ -59,6 +48,23 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
     $(window)
       .resize(resize);
   };
+  var loadMeta = function() {
```
localjo: What do you think of renaming this to `loadMetadata`?

web/js/layers/wv.layers.modal.js
```diff
@@ -59,6 +48,23 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+    if(metaLoaded) return;
+    Object.values(config.layers).forEach(function(layer) {
+      visible[layer.id] = true;
+      if(layer.description){
+        $.get('config/metadata/' + layer.description + '.html')
+          .success(function(data) {
+            self.metadata[layer.id] = data;
+          //$sourceMeta.html(data);
```
localjo: Can we remove these commented out lines completely?

Committing commented out code increases cognitive load and the removed lines can be recovered from the git history.

https://blog.kentcdodds.com/please-don-t-commit-commented-out-code-53d0b5b26d5f

PR #579, opened 2017-10-20
Added layer metadata load handling

Benjaki2:
Metadata loading is now triggered in `wv.main.js`
and metadataLoaded boolean prop is passed to layer list
react component

### DIFF
--- a/web/js/layers/wv.layers.modal.js
+++ b/web/js/layers/wv.layers.modal.js
@@ -75,13 +75,14 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
   };
 
   /**
-   * Returns props to be added to React component 
+   * Uses props the render react component to
+   *  modal
    * 
-   * @method getProps
-   * @return {Object} React component Props
+   * @method renderComponent
+   * @return {Object} React component
    */
-  var getProps = function() {
-    return {
+  var renderComponent = function() {
+    var props =  {
       config: config,
       metadata: self.metadata,
       model: model,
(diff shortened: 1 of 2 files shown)


web/js/layers/wv.layers.modal.js
```diff
@@ -58,8 +64,32 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
...
+    }
   };
 
+  /**
+   * Returns props to be added to React component 
+   * 
+   * @method getProps
+   * @return {Object} React component Props
+   */
+  var getProps = function() {
```
localjo: Why separate this into it's own function? Is it used in more than one place? This is a static object, so why not leave it inline where it's used?

Benjaki2: I almost used it in another place when I was implementing this fix. That is why I broke it out initially. I was thinking about rerendering as opposed to setting the state but that seemed like it would stray away to much from the app universal style right now. I kept it there because I think it is better in general to shorten run-on methods.

localjo: Since it's a static object, how about if we remove the function part, and use a variable instead;

```
var props = {
      config: config,
      metadata: self.metadata,
      model: model,
      width: modalWidth - 20, // modalWidth, minus padding
      height: modalHeight - $('#layer-modal > header').outerHeight() - 30,
      metadataLoaded: metadataLoaded
    };
self.reactList = ReactDOM.render(React.createElement(WVC.LayerList , props), $allLayers[0])
```

That way we don't have to trace through multiple assignments and function executions to figure out where those values come from when we're reading the code.

localjo: If we need to use the variable outside of that function scope, we could move it, but if we're only using it inside that function scope, I think we should keep the definition inside the scope as well. What do you think?

Benjaki2: To me, it would make sense to make that entire snippet it's own function, returning `self.reactList`

PR #582, opened 2017-10-23
Fix metadata preload

localjo:
Fix metadata preload

Because of the limit of parallel requests in browsers (6 in Chrome) there are still some situations on slow connections where the loading of metadata files might be blocking the loading of other resources like images.

This change waits to start loading metadata until all of the initially displayed resources (images, icons, etc) are downloaded, but still blocks some resources that are loaded later like CSS background images on elements that are not initially visible.


localjo:
Abandoned this approach

PR #583, opened 2017-10-24
Performance fix

localjo:
Improves metadata fetching. Relies on https://github.com/nasa-gibs/worldview-components/pull/25

Also removed some dead code.


localjo:
Some minor improvements I'd like to make;
- [ ] Render each layer's metadata icon as it comes in (rather than waiting for all of the requests for visible layers to finish)
- [ ] Improve CSS in IE

PR #584, opened 2017-10-25
Optimize descriptions

ZachTRice:
Optimize descriptions by moving them into WVC component and requesting on i btn click.


ZachTRice:
Abandoned this approach in favor of performance-fix solution

PR #586, opened 2017-10-30
Restore modal offset

ZachTRice:
Restore modal offset so scroll is always visible


localjo:
This was the line that caused the resizing bug with the android keyboard. What does this PR do?

web/js/layers/wv.layers.modal.js
```diff
@@ -689,6 +690,7 @@ wv.layers.modal = wv.layers.modal || function(models, ui, config) {
    */
   var drawAllLayers = function() {
     var projection = models.proj.selected.id;
+
```
localjo: Let's get rid of this whitespace insertion, and squash this PR into a single commit since it's a single change.

PR #590, opened 2017-10-30
Use semi-standard style

localjo:
Change linting style to use StandardJS, plus semicolons: https://standardjs.com 

The only rules overridden in this PR are setting a bunch of error-level rules to warning-level for our legacy code so that we don't have to fix all of our old code to get linting implemented. Aside from that, and requiring semicolons, this avoids adding custom linting rules;

> The whole point of `standard` is to save you time by avoiding [bikeshedding](https://www.freebsd.org/doc/en/books/faq/misc.html#bikeshed-painting) about code style. There are lots of debates online about tabs vs. spaces, etc. that will never be resolved. These debates just distract from getting stuff done. At the end of the day you have to 'just pick something', and that's the whole philosophy of `standard` -- its a bunch of sensible 'just pick something' opinions. Hopefully, users see the value in that over defending their own opinions.


localjo:
Maybe we should run `--fix` on legacy code before starting on module loaders. Will try first thing tomorrow to see the results.

.eslintrc
```diff
@@ -23,14 +23,45 @@
...
+        "curly": "off",
         "semi": ["error", "always"],
         "indent": ["error", 2, {"SwitchCase": 1}],
         "quotes": "warn",
         "no-octal": "warn",
         "no-unused-vars": "warn",
         "no-console": "warn",
         "no-undef": "warn",
-        "no-useless-escape": "warn"
+        "no-useless-escape": "warn",
```
localjo: All of these additions are just turning 'semistandard` errors into warnings for our legacy code so we don't have to fix everything right now.

localjo:
I've also run `standard --fix` and `eslint --fix` to automatically clean up some of the code. And I've cleaned up the dependencies in package.json.

e2e/step_definitions/utility.js
```diff
@@ -40,35 +39,35 @@ defineSupportCode(({Given, Then, When}) => {
...
   });
 
   // Close a tab by index
   Then('I close tab number {int}', (index) => {
     index--;
-    return client.windowHandles(function(tabs) {
+    return client.windowHandles(function (tabs) {
       client.switchWindow(tabs.value[index]);
-      client.closeWindow(function(){
+      client.closeWindow(function () {
```
ZachTRice: I like the spacing improvements but this one bugs me,

ZachTRice: The space between `function` and `()`. The space after is good and all the other rules I saw seem good.

localjo: It seems a little silly, but in cases where there are params there, it makes things a little easier to read;
```
function (arr, index) {
  // stuff
}
```

But the point of using `standard` is [to avoid bikeshedding](https://standardjs.com/#why-should-i-use-javascript-standard-style) over things like this. The library enforces consistency over personal preferences.

Benjaki2: 👍 @ZachTRice @localjo   I like that space as well

localjo: One rule that I don't really care for is always having `{` and `}` in if statements. I like to do clean little one-liners like;
```
if (!relevant) return;
```

But I'd rather have consistency and stick to the standard defaults than spending a bunch of time trying to agree on style preferences that at the end of the day aren't super important as long as we're consistent.

ZachTRice: I'd rather specify standards that work for us. 
`"space-before-function-paren": ["warn", "never"],`
And I think your rule above makes sense too.

These 2 rules are just adding a bit of bloat

localjo: Actually, that's a bad example, because I don't think that's a `standard` rule. But the point is that we should stick to a standard style rather than personal preferences.

localjo: Also, sticking to `standard` is good because we're being consistent with [all of these companies](https://standardjs.com/#who-uses-javascript-standard-style), which makes our code feel more familiar for potential contributors.

ZachTRice: @Benjaki2 what are your thoughts? I lean towards using standard as a guideline and making our own tweaks to that.

Benjaki2: I think that standard is a good place to start, but we can definitely make our own tweaks to that if we get a majority vote.

PR #592, opened 2017-11-01
Make preliminary build task work

localjo:
I think we should merge this and my two PRs in worldview-components before we continue with our branches, so that we all have a working build task to test our changes against.

Here are the changes in this PR:

* Use updated babel preset
* Include jquery as dependency
* Correct import statements
* Stub index.js
* Remove globals from linter rules
* White space fixes


tasks/build-js
```diff
@@ -15,20 +15,12 @@ const outputMin = './build/js.mins.js';
...
   basedir: './web/js'
-})
-  .transform('babelify', {
-    plugins: [ ],
-    presets: [
-      'es2015',
-      'react'
-    ]
-  });
+}).transform('babelify', { presets: ['env'] });
```
ZachTRice: What does `presets: ['env']` indicate and why was sourceMaps removed?

PR #594, opened 2017-11-02
Modularize utils

localjo:
All `wv.util` functions can now be referenced with `import { default as util } from './util/util.js'` and then replacing the `wv.util` reference with just `util` (or whatever you name it in a given file).


web/js/wv.main.js
```diff
@@ -75,6 +75,9 @@ import {debugConfig, debug} from './debug';
 import Brand from './brand';
 import Tour from './tour';
 
+import { default as polyfill } from './polyfill';
+polyfill(); // Polyfills some browser features
```
localjo: Had to move the polyfills out of `utils/browser` into their own module, which needs to be executed from main.

web/js/util/load.js
```diff
@@ -1,7 +1,9 @@
-var wv = wv || {};
-wv.util = wv.util || {};
+import $ from 'jquery';
+import { size } from 'lodash';
+import { default as uiUtils } from '../ui/ui.js';
+import { default as brand } from '../brand.js';
 
-wv.util.load = wv.util.load || (function () {
+export default (function () {
```
ZachTRice: Rename this function to `utilLoad`; then do imports as just `utilLoad` instead of `default as utilLoad`

web/js/util/util.js
```diff
@@ -1,16 +1,21 @@
...
- * General utilities
- *
- * @class wv.util
- * @static
- */
-wv.util = (function (self) {
+/* global ntptEventTag */
+
+import $ from 'jquery';
+import { each, isNull } from 'lodash';
```
Benjaki2: do we want to use a syntax for lodash like `_each` so we know that we are using a lodash method or do we want to just include the whole package to avoid the hassle? @localjo @ZachTRice 

I was working with one file where I was importing 4+ lodash methods so I decided to use the whole package. I could change that though...

I think we should definitely import  the whole `openlayers ` package

ZachTRice: Hm, I have been importing separate lodash functions, but agree we should probably make them unique by appending `_` to the function name if we are doing this.

localjo: We should be importing only the functions we need from `lodash` and any other dependencies that support it. That keeps our bundle size as small as possible.  Even if we're using 4+ lodash methods, importing the whole library blows up bundle sizes.

As far as naming goes, it's all within the scope of the file, so it's easy to see where the function names are defined at the top of the file. We could rename them like `_each`, but really, that's unnecessary.

localjo:
I updated how `lodash` functions are included to follow the lodash documentation;
https://github.com/lodash/lodash#installation

```
// Cherry-pick methods for smaller browserify/rollup/webpack bundles.
var at = require('lodash/at');
var curryN = require('lodash/fp/curryN');
```

localjo:
@ZachTRice @Benjaki2 Can I merge this?

PR #596, opened 2017-11-02
Module loaders work

ZachTRice:


web/js/wv.main.js
```diff
@@ -62,19 +61,21 @@ import NotificationsUI from './notifications/notifications.ui';
...
-import LinkUI from './link/link.ui';
-import LinkInfo from './link/link.info';
+import linkModel from './link/link.model';
+import linkUi from './link/link.ui';
 // Projections
-import projectionParser from './proj/proj';
-import ProjectionModel from './proj/proj.model';
-import ProjectionUI from './proj/proj.ui';
-import ProjectionChange from './proj/proj.change';
+import projectionParser from './projection/projection';
```
localjo: What do you think about combining each folder into a single object? That would match how the code is built right now where each file in the folder inherits and extends from the parent object. So rather than importing each file separately, we can import the whole module in one line;
```
import projection from './projection/projection';
```
And then references like this;
```
wv.projection.model
wv.projection.ui
wv.projection.change
```
Rather than giving them new names, we can drop the `wv.` from existing references and they become;
```
projection.model
projection.ui
projection.change
```

That's what I did with `util` and `ui`. Here's what I mean;

1) Named each module and submodule with the name of it's key on the original object. Parent: 
https://github.com/nasa-gibs/worldview/blob/53ef2980fa575135c68a2d53922c94cb89abd172/web/js/util/util.js#L11
Submodule: https://github.com/nasa-gibs/worldview/blob/53ef2980fa575135c68a2d53922c94cb89abd172/web/js/util/browser.js#L3

2) Imported the submodules into the parent module: https://github.com/nasa-gibs/worldview/blob/53ef2980fa575135c68a2d53922c94cb89abd172/web/js/util/util.js#L7-L9

3) Exported the sub-modules on the parent object: https://github.com/nasa-gibs/worldview/blob/53ef2980fa575135c68a2d53922c94cb89abd172/web/js/util/util.js#L15-L18

What do you think about doing that with projection, layers, and other modules as well? It might reduce the chance of crossing wires by changing so many references.

web/js/layers/layers.add.js
```diff
@@ -1,13 +1,10 @@
...
- * @class wv.layers.add
- */
-wv.layers.add = wv.layers.add || function (models, ui, config) {
+import $ from 'jquery';
+import each from 'lodash/each';
+import find from 'lodash/find';
+import throttle from 'lodash/throttle';
+import util from '../util/util';
+
+export function layersAdd(models, ui, config) {
```
localjo: What do you think about keeping the name `add`, importing this into the `layers` parent, and attaching it to the object that exports?

web/js/layers/layers.info.js
```diff
@@ -1,7 +1,7 @@
-var wv = wv || {};
-wv.layers = wv.layers || {};
+import $ from 'jquery';
+import wvui from '../ui/ui';
 
-wv.layers.info = wv.layers.info || function (config, models, layer) {
+export function layersInfo(config, models, layer) {
```
localjo: What do you think about keeping the name `info`, importing this into the `layers` parent, and attaching it to the object that exports?

web/js/layers/layers.js
```diff
@@ -0,0 +1,129 @@
...
+          });
+        }
+      });
+    }
+    lstates.push(lstate);
+  });
+  state.l = lstates;
+};
+
+export function parse(state, errors, config) {
```
localjo: In the past this function returned;
```
layers: {
  parse: func()...,
  validate: func()...
}
```
And it was referenced with `layers.parse` and `layers.validate`. Will this break existing references?

web/js/layers/layers.options.js
```diff
@@ -1,7 +1,14 @@
...
+import each from 'lodash/each';
+import indexOf from 'lodash/indexOf';
+import isEqual from 'lodash/isEqual';
+import isUndefined from 'lodash/isUndefined';
+import parseInt from 'lodash/parseInt';
+import util from '../util/util';
+import wvui from '../ui/ui';
+import palettes from '../palettes/palettes';
+
+export function layersOptions(config, models, layer) {
```
localjo: What do you think about keeping the name `options`, importing this into the `layers` parent, and attaching it to the object that exports?

web/js/layers/layers.sidebar.js
```diff
@@ -1,13 +1,7 @@
...
-wv.layers = wv.layers || {};
-
-/**
- * @class wv.layers.sidebar
- */
-wv.layers.sidebar = wv.layers.sidebar || function (models, config) {
+import $ from 'jquery';
+import util from '../util/util';
+
+export function layersSidebar(models, config) {
```
localjo: What do you think about keeping the name `sidebar`, importing this into the `layers` parent, and attaching it to the object that exports?

web/js/link/link.model.js
```diff
@@ -1,7 +1,11 @@
...
-wv.link = wv.link || {};
+import $ from 'jquery';
+import each from 'lodash/each';
+import map from 'lodash/map';
+import isArray from 'lodash/isArray';
+import isObject from 'lodash/isObject';
+import util from '../util/util';
 
-wv.link.model = wv.link.model || function (config) {
+export function linkModel(config) {
```
localjo: What do you think about keeping the name `model`, importing this into the `link` parent, and attaching it to the object that exports?

web/js/link/link.ui.js
```diff
@@ -1,12 +1,18 @@
...
+import throttle from 'lodash/throttle';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Clipboard from 'clipboard';
+import Share, {GA as googleAnalytics} from 'worldview-components';
+import util from '../util/util';
+import wvui from '../ui/ui';
 
-wv.link.ui = wv.link.ui || function (models, config) {
+export function linkUi(models, config) {
```
localjo: What do you think about keeping the name `ui`, importing this into the `link` parent, and attaching it to the object that exports?

web/js/palettes/palettes.legend.js
```diff
@@ -1,10 +1,11 @@
...
-wv.palettes = wv.palettes || {};
+import $ from 'jquery';
+import each from 'lodash/each';
+import first from 'lodash/first';
+import last from 'lodash/last';
+import util from '../util/util';
+import palettes from '../palettes/palettes';
 
-wv.palettes.legend = wv.palettes.legend || function (spec) {
+export function palettesLegend(spec) {
```
localjo: What do you think about keeping the name `legend`, importing this into the `palettes` parent, and attaching it to the object that exports?

web/js/palettes/palettes.model.js
```diff
@@ -1,26 +1,26 @@
...
+import each from 'lodash/each';
+import find from 'lodash/find';
+import cloneDeep from 'lodash/cloneDeep';
+import isNaN from 'lodash/isNaN';
+import isUndefined from 'lodash/isUndefined';
+import parseInt from 'lodash/parseInt';
+import util from '../util/util';
+import palettes from '../palettes/palettes';
+
+export function palettesModel(models, config) {
```
localjo: What do you think about keeping the name `model`, importing this into the `palettes` parent, and attaching it to the object that exports?

web/js/projection/projection.change.js
```diff
@@ -1,21 +1,16 @@
...
-wv.proj = wv.proj || {};
+import $ from 'jquery';
+import util from '../util/util';
+import wvui from '../ui/ui';
 
-/**
- * @class wv.proj.change
- */
-wv.proj.change = wv.proj.change || function (models) {
+export function projectionChange(models) {
```
localjo: What do you think about keeping the name `change`, importing this into the `proj` parent, and attaching it to the object that exports?

web/js/layers/layers.model.js
```diff
@@ -1,17 +1,16 @@
...
+import find from 'lodash/find';
+import findIndex from 'lodash/findIndex';
+import isUndefined from 'lodash/isUndefined';
+import util from '../util/util';
 
-/**
- * @class wv.layers.model
- */
-wv.layers.model = wv.layers.model || function (models, config) {
+export function layersModel(models, config) {
```
Benjaki2: When you are just exporting one function like this one. I would suggest exporting it as the default `export default function (models, config) { }`

web/js/projection/projection.js
```diff
@@ -0,0 +1,16 @@
+export function parse(state, errors, config) {
```
localjo: It looks like wv.proj.js was exporting `proj: { parse: func()... }`, and I think this is just exporting `parse: func()...`. Will that cause problems with existing references?

web/js/projection/projection.model.js
```diff
@@ -1,20 +1,16 @@
...
+import each from 'lodash/each';
+import util from '../util/util';
+import ol from 'openlayers';
+import proj4 from 'proj4';
 
-/**
- * @class wv.proj.model
- */
-wv.proj.model = wv.proj.model || function (config) {
+export function projectionModel(config) {
```
localjo: What do you think about keeping the name `model`, importing this into the `proj` parent, and attaching it to the object that exports?

web/js/projection/projection.ui.js
```diff
@@ -1,15 +1,8 @@
...
+import each from 'lodash/each';
+import wvui from '../ui/ui';
 
-/**
- * Undocumented.
- *
- * @class wv.proj.ui
- */
-wv.proj.ui = wv.proj.ui || function (models, config) {
+export function projectionUi(models, config) {
```
localjo: What do you think about keeping the name `ui`, importing this into the `proj` parent, and attaching it to the object that exports?

web/js/layers/layers.model.js
```diff
@@ -182,7 +181,7 @@ wv.layers.model = wv.layers.model || function (models, config) {
   };
 
   self.remove = function (id) {
-    var index = _.findIndex(self.active, {
+    var index = findIndex(self.active, {
```
Benjaki2: I still feel that it would be useful to use `_findIndex` just so someone doesn't have to click around and search for it until they find it in the list of imports. I would assume there is a local findIndex function if I saw this.

web/js/index.js
```diff
@@ -1 +1,22 @@
-export { default as stub } from './stub';
+export { stub, stub2 } from './stub';
```
localjo: I'm thinking maybe we should replace this file with the list of `import`s from the main.js file, and comment out the lines that aren't converted yet. That way we can make sure all of the references in the main.js imports are correct. I noticed a few that are named differently.

web/js/layers/layers.model.js
localjo: There *is* a local `findIndex`, if you're importing it. An `_` before a variable name typically means it's a private variable, and that could be confusing as well.

web/js/layers/layers.model.js
localjo: I was thinking if we leave out `default`, we have to reference the name of the function we're importing, which makes it easier to search the codebase for that function and know that it will have the same name. When we use `default`, we're naming variables whatever we want, which leaves room for error. We had this discussion in Slack and came to that conclusion; https://nasa-gibs.slack.com/archives/C6X6QDJ0J/p1509637754000232

web/js/layers/layers.model.js
Benjaki2: That's a good point. Maybe another symbol in that case. When I see `_.something` I know to look it up in the lodash docs to see exactly what parameters it takes and whatnot. can save time.

web/js/layers/layers.model.js
Benjaki2: 👍 I will update my branch

web/js/image/image.panel.js
```diff
@@ -137,7 +142,7 @@ wv.image.panel = wv.image.panel || function (models, ui, config) {
 
     // Auto-set default resolution to map's current zoom level; round it
     // for incremental zoom steps
-    var curZoom = Math.round(wvx.ui.map.selected.getView()
+    var curZoom = Math.round(ui.map.selected.getView()
```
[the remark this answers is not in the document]
Benjaki2: `wvx.ui` is the same as `ui`.

web/js/image/image.panel.js
```diff
@@ -259,10 +264,9 @@ wv.image.panel = wv.image.panel || function (models, ui, config) {
 
       // get layer transparencies (opacities)
       var opacities = [];
-      _(products)
```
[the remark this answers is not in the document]
ZachTRice: Yes, this tripped me up at first because the previous syntax is wrong. I pulled down a version of development to double-check that correcting this syntax didn't break anything else.

web/js/layers/layers.add.js
ZachTRice: Cool. I can do that, I like that structure.

web/js/image/image.panel.js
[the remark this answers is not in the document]
localjo: It seems like `wvx.ui.map.selected` is referring to the currently selected map, and `./ui/ui.js` has no `map`, so I'm not sure they're the same.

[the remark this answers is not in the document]
localjo: They're definitely different on production;

screen shot 2017-11-03 at 9 46 26 am

web/js/layers/layers.model.js
localjo: We could do something like `import { findIndex as lodashFindIndex } from 'lodash'`, but I don't think that's necessary.

web/js/image/image.panel.js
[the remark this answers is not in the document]
Benjaki2: @localjo we aren't talking about `wv.ui`. The `ui` in this case is the `ui` passed into the function.

https://github.com/nasa-gibs/worldview/blob/master/web/js/wv.main.js#L104-L107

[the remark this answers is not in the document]
localjo: Ah, got it 🙈 I was confusing `wvui` with `ui`. Sorry about that! 😅

PR #597, opened 2017-11-02
Modularize ui

localjo:
Modularize ui as `wvui`.

This PR also modularizes `/web/js/wv.feedback.js` since that was a standalone file referenced by `ui`.


web/js/ui/mouse.js
```diff
@@ -0,0 +1,112 @@
+import * as d3 from 'd3';
```
localjo: This seems to be the usual way to import `d3`, but it sounds like [there might be some problems with importing `d3.event`](https://github.com/d3/d3/issues/2733) which is used in this file. I think it's also possible to import just the functions we need from `d3`, but we should probably wait until we get our whole codebase modularized before tinkering too much with this.

web/js/feedback.js
```diff
@@ -8,7 +16,7 @@ wv.feedback = wv.feedback || (function () {
       'mailto:@MAIL@?subject=Feedback for @LONG_NAME@ tool');
 
     $element.click(function (event) {
-      if (!wv.util.browser.small && window.feedback) {
+      if (!util.browser.small && window.feedback) {
```
Benjaki2: I think `util.browser` is a separate file

web/js/ui/info.js
```diff
@@ -75,18 +74,18 @@ wv.ui.info = wv.ui.info || function (ui, config) {
...
+    wvui.positionMenu($menuItems, {
       my: 'left top',
       at: 'left bottom + 5',
       of: $label
     });
     $menuItems.hide();
 
     $about.click(function () {
-      if (wv.util.browser.small || wv.util.browser.touchDevice) {
+      if (util.browser.small || util.browser.touchDevice) {
```
Benjaki2: another util.browser bug

web/js/feedback.js
localjo: It is a separate file, but the `util` module imports the `browser` object to keep it all on the same parent object;

https://github.com/nasa-gibs/worldview/blob/53ef2980fa575135c68a2d53922c94cb89abd172/web/js/util/util.js#L7-L9

https://github.com/nasa-gibs/worldview/blob/53ef2980fa575135c68a2d53922c94cb89abd172/web/js/util/util.js#L16-L18

This is consistent with the behavior before where these objects were all inheriting from and extending the same `wv.util` object. It also makes updating the references easier and less error prone, and importing all of the submodules of `util` less verbose.

web/js/ui/info.js
localjo: That's intentional.

PR #607, opened 2017-11-13
Module loader debugging

ZachTRice:
* Shimmed d3
* Shimmed Jquery and it's plugins
* Removed `const` that were exported and replaced them with default executions
* Added build folder to place store transpiled code
* Removed uglify
* Removed jQuery plugin tags
* Adjusted `export` && `import` naming to make them correspond/work
* A number of bug fixes with help from the linter


tasks/build-js
```diff
@@ -5,10 +5,14 @@
 const fs = require('fs');
 const browserify = require('browserify');
 const watchify = require('watchify');
+const outputDir = './web/build/';
```
localjo: Why are we creating a new build directory in `/web` instead of using `/build`?

Gruntfile.js
```diff
@@ -613,7 +613,11 @@ module.exports = function(grunt) {
       }
     },
 
-    uglify: {
+    'build-js': {
+      command: 'npm run build-js'
```
localjo: Why are we calling npm scripts from Grunt? Aren't we trying to move things out of Grunt and use npm directly?

web/index.html
```diff
@@ -104,37 +104,7 @@
...
-
-    <!-- link.dev  -->      <script type="text/javascript" src="ext/node_modules/gifshot/build/custom/gifshot.custom.js"></script>
-    <!-- link.prod --> <!-- <script type="text/javascript" src="ext/node_modules/gifshot/build/custom/gifshot.custom.min.js"></script> -->
-
-    <!-- link.dev -->       <script type="text/javascript" src="ext/node_modules/clipboard/dist/clipboard.js"></script>
-    <!-- link.prod --> <!-- <script type="text/javascript" src="ext/node_modules/clipboard/dist/clipboard.min.js"></script> -->
-
-
+  
     <!-- link.dev  -->      <script type="text/javascript" src="ext/main/jquery-2.1.4/jquery.js"></script>
```
localjo: Why are we leaving the jQuery scripts here?

web/index.html
```diff
@@ -171,109 +136,11 @@
...
-
-    <!-- link.dev --> <script type="text/javascript" src="js/link/wv.link.model.js"></script>
-    <!-- link.dev --> <script type="text/javascript" src="js/link/wv.link.ui.js"></script>
-
-    <!-- link.dev --> <script type="text/javascript" src="js/wv.feedback.js"></script>
-    <!-- link.dev --> <script type="text/javascript" src="js/wv.tour.js"></script>
-    <!-- link.dev --> <script type="text/javascript" src="js/wv.debug.js"></script>
-    <!-- link.dev --> <script type="text/javascript" src="js/wv.main.js"></script>
+    <!-- link.dev --> <script type="text/javascript" src="build/wv.js"></script>
+    <!-- link.prod --> <!-- <script type="text/javascript" src="js/wv.min.js?v=@BUILD_NONCE@&mobile"></script> -->
```
localjo: Are `build/wv.js` and `js/wv.min.js` the same thing?

web/js/date/wheels.js
```diff
@@ -24,7 +25,7 @@ export function dateWheel(models, config) {
     $container
       .addClass('datespan')
       .html('<div id=\'wv-date-mobile-label\'></div><input type=\'hidden\' id=\'linkmode\' readonly>');
-
+    console.log(mobiscroll);
```
localjo: Any reason to have a `console.log` committed here?

web/index.html
Benjaki2: We aren't. We haven't gotten to this yet. This PR is not ready for review. Still work to be done.

tasks/build-js
Benjaki2: at this point... at least a few tasks depend on everything being located in the web directory

Benjaki2:
Let's squash this.

Benjaki2:
@ZachTRice @localjo we could add a `shim-config` file to clean up our `package.json` but I think that would make it so we can't use aliases. Which seems annoying.

PR #612, opened 2017-11-16
Fixed tab switching & "add layer" bugs

ZachTRice:
- [x] Change palettesLegend import to reference a function
- [x] Import googleAnalytics from worldview-components to fix tab error
- [x] Add previously undefined variables errors when adding layer.
- [x] Add check for null variable when passed into ui hide function.
- [x] Add previously undefined variables breaking the colorbar on hover
- [x] Fix layersInfo import naming preventing info window from opening
- [x] Fix layersOptions import naming preventing config window from opening
- [x] Fix `$colorBox is not defined` error in palettes/legend.js
- [x] Fix `palette is not defined` in palettes/model.js
- [x] Fix `ReferenceError: oldLegend is not defined` in palettes/model.js
- [x] Fix `cannot call methods on dialog prior to initialization; attempted to call method 'close'` error when attempting to use options button to open options dialog after info dialoag is open; or vice-versa

After latest updates to module-loaders base branch a few more bugs were introduced:
- [x] Fix nouislider dependency; `nouislider-browser` is v5, `nouislider` starts at v7. We were using v6.1.0
- [x] Fix jQuery dialog box styling affected by change to jQuery-ui-1.10.4
- [x] Fix `cache.removeWhere` "o.removeWhere is not a function" error -- relates to jscache package
- [x] Fix issue with opacity slider performing strangely onClick (slow to move around, and keeps moving after click release)

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "jscrollpane": "^2.0.22",
     "lodash": "4.15.0",
     "mobiscroll-jquery": "^3.2.6",
-    "nouislider-browser": "^5.0.0-1",
+    "nouislider": "7.0.10",
     "ol": "^4.4.2",
     "perfect-scrollbar": "^0.8.1",
     "proj4": "2.3.3",
@@ -110,6 +110,7 @@
   },
   "browser": {
     "d3": "./web/ext/d3/d3.js",
+    "jscache": "./web/ext/jscache-gitba01cdc/cache.js",
     "isotope-layout": "./node_modules/isotope-layout/dist/isotope.pkgd.js",
     "jquery": "./node_modules/jquery/dist/jquery.js"
   },
--- a/package.json
+++ b/package.json
@@ -110,7 +110,6 @@
   },
   "browser": {
     "d3": "./web/ext/d3/d3.js",
-    "jscache": "./web/ext/jscache-gitba01cdc/cache.js",
     "isotope-layout": "./node_modules/isotope-layout/dist/isotope.pkgd.js",
     "jquery": "./node_modules/jquery/dist/jquery.js"
   },
--- a/web/css/main.css
+++ b/web/css/main.css
@@ -10,7 +10,7 @@
 @import url("../../node_modules/jquery.joyride/jquery.joyride.css");
 @import url("../../node_modules/jquery-jcrop/css/jquery.Jcrop.css");
 @import url("../../node_modules/ol/ol.css");
-@import url("../../node_modules/nouislider-browser/jquery.nouislider.css");
+@import url("../../node_modules/nouislider/distribute/jquery.nouislider.min.css");
 
 /* App CSS */
 @import url("fonts.css");
--- a/web/css/main.css
+++ b/web/css/main.css
@@ -10,7 +10,7 @@
 @import url("../../node_modules/jquery.joyride/jquery.joyride.css");
 @import url("../../node_modules/jquery-jcrop/css/jquery.Jcrop.css");
 @import url("../../node_modules/ol/ol.css");
-@import url("../../node_modules/nouislider/distribute/jquery.nouislider.min.css");
+@import url("../../node_modules/nouislider/src/jquery.nouislider.css");
 
 /* App CSS */
 @import url("fonts.css");
--- a/web/js/layers/active.js
+++ b/web/js/layers/active.js
@@ -8,8 +8,8 @@ import loEach from 'lodash/each';
 import loEachRight from 'lodash/eachRight';
 import util from '../util/util';
 import wvui from '../ui/ui';
-import layersInfo from './info';
-import layersOptions from './options';
+import {layersInfo} from './info';
+import {layersOptions} from './options';
 import {palettesLegend} from '../palettes/legend';
 
 export function layersActive(models, ui, config) {
--- a/web/js/layers/options.js
+++ b/web/js/layers/options.js
@@ -147,8 +147,7 @@ export function layersOptions(config, models, layer) {
         }
       })
       .on('slide', function () {
-        models.layers.setOpacity(layer.id, parseFloat($(this)
-          .val()));
+        models.layers.setOpacity(layer.id, $(this).val());
       });
     var $label = $('<div></div>')
       .addClass('wv-label')
--- a/web/js/layers/options.js
+++ b/web/js/layers/options.js
@@ -273,6 +273,7 @@ export function layersOptions(config, models, layer) {
   };
 
   var onRangeUpdate = function () {
+    var current;
     updateRangeLabels();
 
     var palette = models.palettes.get(layer.id, index);
--- a/web/js/layers/options.js
+++ b/web/js/layers/options.js
@@ -2,7 +2,7 @@ import $ from 'jquery';
 import 'jquery-ui/button';
 import 'jquery-ui/dialog';
 import 'icheck';
-import 'nouislider-browser/jquery.nouislider';
+import 'nouislider/distribute/jquery.nouislider';
 import 'perfect-scrollbar/jquery';
 import loEach from 'lodash/each';
 import loIndexOf from 'lodash/indexOf';
--- a/web/js/layers/options.js
+++ b/web/js/layers/options.js
@@ -273,14 +273,12 @@ export function layersOptions(config, models, layer) {
   };
 
   var onRangeUpdate = function () {
-    var current;
-    updateRangeLabels();
-
     var palette = models.palettes.get(layer.id, index);
     var imin = (loIsUndefined(palette.min)) ? 0 : palette.min;
     var imax = (loIsUndefined(palette.max))
       ? palette.legend.tooltips.length - 1 : palette.max;
-    current = [parseFloat($range.val()[0]), parseFloat($range.val()[1])];
+    var current = [parseFloat($range.val()[0]), parseFloat($range.val()[1])];
+    updateRangeLabels();
     if (!loIsEqual(current, [imin, imax])) {
       $range.val([imin, imax]);
     }
--- a/web/js/map/ui.js
+++ b/web/js/map/ui.js
@@ -80,13 +80,13 @@ export function mapui(models, config) {
       .on('remove', removeLayer)
       .on('visibility', updateLayerVisibilities)
       .on('opacity', updateOpacity)
-      .on('update', updateLayer);
+      .on('update', updateLayerOrder);
     models.date.events.on('select', updateDate);
     models.palettes.events
-      .on('set-custom', updateLayer)
-      .on('clear-custom', updateLayer)
-      .on('range', updateLayer)
-      .on('update', updateLayer);
+      .on('set-custom', updateLookup)
+      .on('clear-custom', updateLookup)
+      .on('range', updateLookup)
+      .on('update', updateLookup);
     $(window)
       .on('resize', onResize);
     updateProjection(true);
@@ -339,13 +339,13 @@ export function mapui(models, config) {
   /*
    * Reloads layer on change
    *
-   * @method updateLayer
+   * @method updateLayerOrder
    * @static
    *
    *
    * @returns {void}
    */
-  var updateLayer = function () {
+  var updateLayerOrder = function () {
     reloadLayers();
   };
 
@@ -370,6 +370,21 @@ export function mapui(models, config) {
         .setAt(index, createLayer(def));
     });
     updateLayerVisibilities();
+  };
+
+  /*
+   * Update layers for the correct Date
+   *
+   * @method updateLookup
+   * @static
+   *
+   *
+   * @returns {void}
+   *
+   * @todo Check if this function can be combined with updateLayerOrder
+   */
+  var updateLookup = function (layerId) {
+    reloadLayers();
   };
 
   self.getCustomLayerTimeout = function (layer) {
--- a/web/js/map/ui.js
+++ b/web/js/map/ui.js
@@ -80,13 +80,13 @@ export function mapui(models, config) {
       .on('remove', removeLayer)
       .on('visibility', updateLayerVisibilities)
       .on('opacity', updateOpacity)
-      .on('update', updateLayerOrder);
+      .on('update', updateLayer);
     models.date.events.on('select', updateDate);
     models.palettes.events
-      .on('set-custom', updateLookup)
-      .on('clear-custom', updateLookup)
-      .on('range', updateLookup)
-      .on('update', updateLookup);
+      .on('set-custom', updateLayer)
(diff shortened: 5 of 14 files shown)


web/js/layers/active.js
```diff
@@ -5,7 +5,7 @@ import util from '../util/util';
 import wvui from '../ui/ui';
 import layersInfo from './info';
 import layersOptions from './options';
-import palettesLegend from '../palettes/legend';
+import {palettesLegend} from '../palettes/legend';
```
localjo: Just curious, why is this change being made? Don't both versions mean the same thing?

web/js/util/load.js
```diff
@@ -36,8 +36,7 @@ export default (function () {
           loading -= 1;
           if (loading === 0) {
             clearTimeout(indicatorTimeout);
-            indicatorTimeout = null;
```
localjo: Was this line causing problems? I wonder if it was in there for browser compatibility?

ZachTRice: After testing, the fix is related to the next line in the code. Passing in a null parameter causes issues with the qvui hide function.

web/js/layers/active.js
ZachTRice: Otherwise I get `Uncaught TypeError: (0 , _legend2.default) is not a function`. To be perfectly fair, I am not entirely sure why it is only working this way. There seems to be quite a few gotchas related to import and export naming which I need to review further to understand what the issue is in this particular case.
https://stackoverflow.com/questions/36795819/when-should-i-use-curly-braces-for-es6-import/36796281

web/js/util/load.js
localjo: The next line uses `indicatorId` not `indicatorTimeout`

web/js/layers/active.js
localjo: Ah, I understand now. If you're importing something by name, rather than the default, the `{}` are required.

localjo:
This looks good to me. Could you update the wording of the commit message to use imperative mood before merging?

localjo:
This change seems to cause some problems when adding a layer. When I check the box for a layer I see;
`Uncaught TypeError: Cannot read property 'hide' of undefined                 load.js:40`

localjo:
This looks good to me. Could you update the wording of the commit message to use imperative mood before merging?

ZachTRice:
Commit message has been updated

web/js/layers/options.js
```diff
@@ -273,6 +273,7 @@ export function layersOptions(config, models, layer) {
   };
 
   var onRangeUpdate = function () {
+    var current;
```
localjo: You can add the `var` keyword on line 283 instead, since that's in the middle of the list of vars being declared for this function scope. If you want the variables to be defined first in this function, you can move the `updateRangeLabels()` call below that list.

web/js/palettes/legend.js
```diff
@@ -59,6 +59,7 @@ export function palettesLegend(spec) {
   };
 
   var renderScale = function ($legendPanel, legend, index, layerId) {
+    var $container, $colorbar;
```
localjo: You can add the `var` keyword to the next two lines where these variables are defined, instead of breaking the variable definitions into a separate line like this.

web/js/palettes/legend.js
```diff
@@ -116,10 +117,10 @@ export function palettesLegend(spec) {
   };
 
   var updateClasses = function (legend, index) {
+    var $colorBox;
```
localjo: You can add the `var` keyword two lines down where `$colorBox` is actually defined.

web/js/palettes/model.js
```diff
@@ -27,6 +27,7 @@ export function palettesModel(models, config) {
   };
 
   self.getCustom = function (paletteId) {
+    var palette;
```
localjo: You can add the `var` keyword to the next line instead.

web/js/palettes/model.js
```diff
@@ -346,7 +347,7 @@ export function palettesModel(models, config) {
 
     loEach(state.l, function (layerDef) {
       var layerId = layerDef.id;
-      var minValue, maxValue;
+      var values, minValue, maxValue;
```
localjo: Why are you defining `values` outside of the scope it was originally used in? It looks like the line `values = util.toArray(attr.value.split(';'));` should have the `var` keyword added to it to keep the variable in the same scope. Moving it out of that scope could change the behavior.

web/js/palettes/model.js
```diff
@@ -479,12 +480,13 @@ export function palettesModel(models, config) {
   };
 
   var updateLookup = function (layerId) {
+    var legend, oldLegend, entries;
```
localjo: You're moving these variables outside of their original scope. Why?

web/js/palettes/legend.js
ZachTRice: There are two each methods containing `$colorbox`, so this var needs to be defined at the top in this function.

web/js/palettes/model.js
```diff
@@ -479,12 +480,13 @@ export function palettesModel(models, config) {
...
+    var lookup = {};
```
localjo: You're moving this declaration before the ["early return"](https://forum.freecodecamp.org/t/the-return-early-pattern/19364) on the next line. There would be no need to declare it if the function returns early.

web/js/palettes/legend.js
localjo: It looks like those are two different `$colorBox` variables that just happen to have the same name. Are they supposed to refer to the exact same element? Doesn't making them use the same variable increase the risk that one might overwrite the other?

localjo: That existing lodash each loop is a little strange. It looks like the same loop happens twice;
```
loEach(legend.colors, function (color, classIndex) {
```
That could probably be simplified to only happen once.

web/js/palettes/model.js
ZachTRice: Otherwise `values` is undefined. Seems to make sense to define it here.

web/js/palettes/model.js
ZachTRice: Otherwise these values are undefined and when you use these features in WV, the console throws a `ReferenceError: oldLegend is not defined` error.

web/js/palettes/legend.js
ZachTRice: Yea, this is weird. It seems like these two can be combined, I'll look at combining these.

web/js/palettes/model.js
localjo: Why do you need `values` to be defined outside of the scope it was originally defined in? This `var` declaration belongs inside of the `loEach` loop, on line 358 where it's being given a value for the first time, unless you're using it before that.

web/js/palettes/model.js
localjo: There are no lines referencing these values outside of their original scope. You should add the `var` keyword inside the `loEach` loop, and you can just add it to the lines where those variables are given values. No need to make it a separate line.

localjo:
Keep in mind that everything inside of a lodash.each loop is inside of a callback function, which means [it's in a new scope](http://2ality.com/2011/02/javascript-variable-scoping-and-its.html). We should be careful about changing variable scopes unintentionally because it could have side effects that very difficult to debug during a refactor.

ZachTRice:
Thanks. I have moved variables back into the each loops for now. Layer sidebar info, option and colorbar functions are all working. There is still a bug when hovering the map where the colorbar label doesnot appearing but this code is related to the map (code in map/runningdata.js) so I am going to document that in the main module loaders branch.

web/js/palettes/legend.js
```diff
@@ -118,11 +118,13 @@ export function palettesLegend(spec) {
...
     $panel.empty();
+    var $detailPanel = $('<div></div>');
     loEach(legend.colors, function (color, classIndex) {
-      var $colorBox;
+      var label = legend.tooltips[classIndex];
+      label = (legend.units) ? label + ' ' + legend.units : label;
       var $runningDataPointLabel = $('<span></span>')
         .addClass('wv-running-category-label');
-      $colorBox = $('<span></span>')
+      var $colorBox = $('<span></span>')
```
localjo: It looks like the `$colorBox` in the first loop and the `$colorBox` in the second loop are different HTML elements, so I think they should use two different variables. You could either give the two elements different variable names, which is probably easiest, or refactor a bit more and eliminate the need to store the "color box" in an intermediate variable at all. What do you think?

ZachTRice:
This PR is a complete mess now after merging in the latest changes to `module-loaders` which includes the addition of imported modules. One of the big issues is the inconsistencies with the packages we were using before.

`jquery-ui-1.11.4` was used before, now we are imprting `jquery-ui-1.10.4`. This has affected the styles of the dialog box, not to mention any other js issues that may arise that aren't evident.

`noUiSlider` was using 6.1.10, now I am trying to import `noUiSlider` 7.0.10 because it is the closest version that is semi-working. NPM linking and importing 6.1.10 fixes some issues but then I am left debugging additional (new) issues related to `cache.removeWhere` function which I am trying to track down.

localjo:
It might be best to track down these problems one at a time. Since we don't have testing in place to confirm that our changes aren't breaking things we need to be very careful about reviewing each other's changes and looking for problems. The version mismatches should have been caught before they were merged. But the nature of trying to refactor an app with years of technical debt and no testing is that we're going to have a lot of bugs to track down.

ZachTRice:
The cache issue and slider issues appear to be only used in this area of the app so I am documenting this issues here. If they end up being larger in scope, then I will create a task in the main branch. The version mismatches are just going to be that way since there aren't equivalent npm packages.

ZachTRice:
@localjo could I get you to look over this PR when you get a chance. I would like to get it merged into the base module-loaders branch. It is currently up to speed with the latest changes in that branch.

web/css/main.css
```diff
@@ -10,7 +10,7 @@
 @import url("../../node_modules/jquery.joyride/jquery.joyride.css");
 @import url("../../node_modules/jquery-jcrop/css/jquery.Jcrop.css");
 @import url("../../node_modules/ol/ol.css");
-@import url("../../node_modules/nouislider-browser/jquery.nouislider.css");
+@import url("../../node_modules/nouislider/distribute/jquery.nouislider.min.css");
```
localjo: If there's a non-minified version of CSS, you should prefer that, since the build script minifies and generates a source map that points to the original file. If we include the minified version, we lose the ability to use the source map.

web/js/layers/options.js
```diff
@@ -147,8 +147,7 @@ export function layersOptions(config, models, layer) {
         }
       })
       .on('slide', function () {
-        models.layers.setOpacity(layer.id, parseFloat($(this)
-          .val()));
+        models.layers.setOpacity(layer.id, $(this).val());
```
localjo: Why did you remove the `parseFloat` here?

web/js/map/ui.js
```diff
@@ -337,15 +337,15 @@ export function mapui(models, config) {
...
    *
-   * @method updateLayerOrder
+   * @method updateLayer
    * @static
    *
    *
    * @returns {void}
    */
-  var updateLayerOrder = function () {
+  var updateLayer = function () {
```
localjo: Why did you rename this function? `updateLayerOrder` and `updateLayer` mean different things. Is this function doing something different than it was before?

web/js/map/ui.js
```diff
@@ -80,13 +80,13 @@ export function mapui(models, config) {
...
-      .on('update', updateLayerOrder);
+      .on('update', updateLayer);
     models.date.events.on('select', updateDate);
     models.palettes.events
-      .on('set-custom', updateLookup)
-      .on('clear-custom', updateLookup)
-      .on('range', updateLookup)
-      .on('update', updateLookup);
+      .on('set-custom', updateLayer)
+      .on('clear-custom', updateLayer)
```
localjo: Why did you change which function is passed as a callback on these four lines?

web/js/map/ui.js
localjo: If this function is just calling `reloadLayers`, why not use that directly, instead of having it wrapped inside another function?

web/js/palettes/legend.js
```diff
@@ -136,14 +135,15 @@ export function palettesLegend(spec) {
...
       $colorBox.on('mouseout', hideUnitsOnMouseOut);
     });
+    // TODO: Review this each loop. It can probably be removed.
     var $detailPanel = $('<div></div>');
     loEach(legend.colors, function (color, classIndex) {
       var label = legend.tooltips[classIndex];
       label = (legend.units) ? label + ' ' + legend.units : label;
       var $row = $('<div></div>')
-        .addClass('wv-palettes-class-detail')
+        .addClass('$colorBox')
```
localjo: Why are you changing the class on each row to `$colorBox` instead of `wv-palettes-class-detail`?

package.json
```diff
@@ -110,6 +110,7 @@
   },
   "browser": {
     "d3": "./web/ext/d3/d3.js",
+    "jscache": "./web/ext/jscache-gitba01cdc/cache.js",
```
localjo: Can this be removed?

web/js/layers/options.js
ZachTRice: `parseFloat` was affecting the click dragging of the slider. There were performance issues / bugginess. Removing it fixes these issues. The step, min and max are defined so I don't see why it is needed.

web/js/map/ui.js
ZachTRice: When we found out that `cache.removeWhere` was broken, I removed it from `updateLookup`. This only left a call to `reloadLayers` within `updateLookup`. But there was another function only calling `reloadLayers`, `updateLayerOrder` so I removed `updateLookup` and renamed `updateLayerOrder` to `updateLayer` since that is a more accurate description of what is happening.

web/js/map/ui.js
ZachTRice: I tried using it directly, but it wasn't working.

web/js/palettes/legend.js
ZachTRice: This was accidentally introduced. Good catch.

web/js/map/ui.js
ZachTRice: Also, see my comment above about the reason behind renaming this function.

ZachTRice:
The iCheck issue was related to a change in the "patched" version of iCheck we were running before. See:
https://github.com/fronteed/icheck/blob/1.x/icheck.js#L157
is not the same as
https://github.com/nasa-gibs/worldview/commit/ba80f504d313b0d5b87cfe848ad17ee723e2eb49#diff-dc550be1377c6cbd4583cac3ad4af6d5R157

PR #503, opened 2017-09-07
Module loaders

Benjaki2:
Resolves #494 
Fixes #528
Fixes #673 

Depends on https://github.com/nasa-gibs/worldview-components/pull/32

# Outstanding Items

## Features to be modularized and re-enabled
- [x] Core
- [x] Tour - @ZachTRice #639
- [x] Animation - @Benjaki2 #640 
- [x] Data Download - @Benjaki2 #654 
- [x] Image Download - @Benjaki2 #645
- [x] Natural Events - @localjo #648 
- [x] Notifications - @Benjaki2 #644
- [x] URL Shortening (done along with other features)
- [x] Custom Palettes  (done along with other features)
- [x] What's New - @Benjaki2 (#644)
- [x] Feedback (done along with #639)
- [x] Arctic Projection Change - @localjo #650 
- [x] Google Analytics (done along with other features)
- [x] Alert (done along with other features)
- [x] Timeline mobile wheels - @localjo #651 
- [x] Debug.js - @localjo #653 

## Update Buster Tests 
All tests should be modularized, and the code should be tested against them before that section is marked complete)
- [x] Core files (fixtures.js, wv.brand-test.js)
- [x] Data
- [x] Date
- [x] Layer
- [x] Link
- [x] Map
- [x] Palettes
- [x] Proj
- [x] UI
- [x] Util

## Other Tasks
- [x] Clean up build scripts
- [x] Remove converted dependencies (wv.js.json, dependencies, etc.)
- [x] Move all dependencies from `/ext` to package.json
- [x] Make sure all required CSS files are included (some have been removed)
- [x] Shim or include jscache library in package.json
- [x] Enable watch mode for JS and CSS build tasks - @localjo #647 
- [x] Make build script work for dev and production environments - @localjo #647 

## Bugs
- [x] Fix projection button (Getting the error "Cannot read property 'close' of undefined in projection/projection/ui.js")
- [x] Fix "Add layer" errors (in layer module)
- [x] Fix error related to WVC on tab switching
- [x] The URL state is not updating
- [x] iCheck boxes are missing `iCheck iCheck-checkbox` classes on div which is affecting the styling
- [x] Hovering a layer on the map the includes a colorbar does not update the label in the sidebar. (`.wv-running` / `.wv-active` classes not being added from `map/runningdata.js`)
- [x] `jQuery.easing[this.easing] is not a function` error when opening projection switcher.
- [x] Styling of projection dropdown is off (more spacing and not highlighted)
[...]

PR #613, opened 2017-11-17
Move ext scripts to dependencies key

localjo:
* Removed browser shim where not needed
* Imported dependencies where they are used
* Added dependencies to the `dependencies` key of package.json so they are part of our dependency graph
* Use npm installed version of dependencies instead of /ext directory

Still need to:
- [x] Remove `d3` global var
- [x] Figure out the correct way to include `isotope`
- [x] Move cache.js to dependencies
- [x] Import dependency CSS files from `node_modules`
- [x] Replace `browserify-css` with `postcss` npm script
- [x] Delete unneeded files from /ext directory
- [x] Remove dead code from Gruntfile


Benjaki2:
@localjo the `browser`key does not work with files inside of `node_modules` directory (strange, I know). Also, remember we talked about keeping the that `ext/` version of `d3` for the time being as we are dependent on a forked version for the time being.

localjo:
Looks like there's an issue with the `d3` version we're using depending on `jsdom` which depends on `contextify` which breaks `npm install`. So I guess we're going to have to stick with `d3` in `/ext` until we can update to not use the fork anymore.

localjo:
For reference, the d3 version we're using is basically 3.4.11 plus the changes in this commit: https://github.com/d3/d3/pull/2019/files

Benjaki2:
I'm seeing console errors with `jQuery.easing` not being a method (I assume this is a jQuery-ui thing) along with `icheck`

localjo:
@Benjaki2 Thanks. I'll look into that. Can you give me more details about what line you're seeing that error on and what actions cause it? 

Also, don't spend too much time debugging this yet. I'll tag you and Zach when it's ready for review, but it's only half done right now.

localjo:
I think this branch, in it's current state is an improvement and all of the changes work. However, I want to make some changes to the CSS bundling to align this more with our goals.

I used `browserify` to bundle the CSS, but when I wanted to add some more CSS transforms (auto-prefixing, specifically), I realized I wanted to use `browserify-postcss`, which could essentially be swapped in as-is, but then I realized that we talked about wanting to use libraries directly via npm scripts instead of using wrappers.

That should also be a pretty easy change at this point. I've got it mostly working locally with this script;

```bash
postcss ./web/css/main.css --use postcss-import --output ./web/build/wv.css
```

The only thing that's not working is relative paths for `font-awesome` font files. I had to resolve that problem for `browserify-css` as well, and I can use [`postcss-assets-rebase`](https://github.com/devex-web-frontend/postcss-assets-rebase) in this case. I just need to figure out how to configure that plugin. I think I'll need to move away from using the CLI in npm scripts to using the JS api like we're doing with browserify. Fairly straightforward, but I might not finish it before the holiday weekend.

localjo:
See the top comment for all of the items this PR accomplishes. The two unfinished items can probably be saved for later on in #503, since they're related to the final build process.

The only changes should be related to the JS and CSS bundles, but I'm not sure if any of that broke the string replacement tasks. We'll also need to figure out how to enable "watch" mode at the right times so that it doesn't block other build scripts from running. And finally, we'll want to figure out exactly where/how the final built assets should be located and if they should be in the same place on both dev and production.

All of the outstanding items could be figured out later in a separate PR, and I'd like to move forward with getting this merged so we're all working from the same base.

@Benjaki2 and @ZachTRice go ahead and review this and let me know what you think.

localjo:
To get this running, you'll probably have to do these steps;
```
rm -rf node_modules
npm install
npm link worldview-components
npm run build
npm start
```

localjo:
I agree with those thoughts about CSS. Another idea worth considering is that we can actually `import` CSS files in JS. A common pattern that builds on the idea of renaming CSS files to have a 1:1 correlation with JS modules is to get rid of the single entry point CSS file and move the CSS files into the same folder as the JS module they apply to and then `import`ing them in the JS module. The bundler (whether Browserify, Webpack or something else) can then extract them and group them into a single bundle or inject them into the document using a `<style>` tag. I think that change and some of the others you suggested should be saved for another project when we can focus on our ideal CSS organization, but the changes in this PR enabling CSS bundling lay the foundation for us to be able to do that easily later.

PR #615, opened 2017-11-27
Remove dependency artifacts and update docs

localjo:
Just some cleanup of some outstanding items on #503


localjo:
This should be ready to merge if it looks good to you guys.

localjo:
@ZachTRice @Benjaki2 @minniewong @rboller I'd like your feedback on my changes to developing.md in this PR. If there is anything lacking in my descriptions of the Grunt tasks, let me know. The intention is to keep the descriptions brief while explaining why someone would use that task and what the results of the task will be. Here's a link to view my changes to the file: https://github.com/nasa-gibs/worldview/pull/615/files#diff-e0bc0ebf70515a5abdf1ae720b279282

ZachTRice:
I'm not in favor of completely removing the style_guide document. Not everyone is going to have a linter to view how their code should be formatted. Without this documentation there is not mention of how to format CSS and the modifications we have made to standardjs. This is also a good place to document any other kind of styles we want to maintain consistency with such as creating elements. If someone wanted to add a UI element, eventually, it would be good to have a place to say all modals should have 3px radius border, use this set of colors, etc.

doc/style_guide.md
```diff
@@ -1,151 +0,0 @@
-# Developer Guide
```
ZachTRice: This document should not be removed in my opinion. There is a lot of useful information here.

localjo:
@Benjaki2 I'm in favor of moving `eslint` and `stylelint` out of Grunt. Should be pretty easy to do, but I can do it in another PR since it's a separate change.

doc/style_guide.md
localjo: This moved to CONTRIBUTING.md

localjo:
@ZachTRice I moved the style guide information here: https://github.com/nasa-gibs/worldview/pull/615/files#diff-6a3371457528722a734f3c51d9238c13R27

localjo:
A visual style guide is a completely separate issue from a code style guide. I don't think we need to be concerned about that in this PR.

doc/style_guide.md
ZachTRice: Seems like we are loosing a lot of important information in the process.

doc/developing.md
```diff
@@ -1,156 +1,53 @@
...
 
-* ``dist/<name>-<version>-<release>.git<revision>.tar.bz2``
+`grunt config`* : Updates branding and configuration options.
 
-Where ``name``, ``version``, ``release`` are the values found in the
-``package.json`` file and ``revision`` is the value from:
+`grunt site`* : Generates build assets for deployment.
 
-    git rev-parse --short HEAD
+`grunt fetch`* : Makes a request to the GIBS `GetCapabilities` API and updates layer configurations.
```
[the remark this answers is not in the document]
minniewong: agreed

doc/style_guide.md
localjo: No, this entire style guide document has been replaced and is irrelevant now.

localjo: The problem with trying to maintain a `.stylelintrc`/`.eslintrc` and a separate description of those rules is that there are then two sources of truth and the two will never stay in sync. It's best to have people refer to once source of truth, which is the place where the rules are configured, `.stylelintrc` and `.eslintrc`.

ZachTRice: True, updating two locations for one change doesn't make sense. I still feel like we are loosing something by not having real examples but if I am in the minority on this then I'm not going to hold up the PR over it.

PR #617, opened 2017-11-28
Fixed condition for running-data retrievel limit

### LINKED ISSUE #616: onMouseMove: dataRunner is not triggered if naturalEvent is disabled
It took me some time to figure why the cursor was not displayed on the legend despite the fact that I created my own color map in the GIBS format. 

It seems that the condition is wrong in onMouseMove function:
 https://github.com/nasa-gibs/worldview/blob/6cbe782049c69571305b8402e41d949eee4bffd5/web/js/map/wv.map.ui.js#L878-L882

I think that the condition should be:
``` javascript
      if ((typeof models.naturalEvents != 'undefined' &&
            models.naturalEvents.active) ||
          (typeof models.data != 'undefined' && models.data.active)) {
        return;
      }
```

PacoDu:
Fixes #616


web/js/map/wv.map.ui.js
```diff
@@ -875,9 +875,9 @@ wv.map.ui = wv.map.ui || function(models, config, components) {
       if (self.mapIsbeingDragged || wv.util.browser.small) {
         return;
       }
-      if (typeof models.naturalEvents == 'undefined' ||
-        typeof models.data == 'undefined' ||
-        models.naturalEvents.active || models.data.active) {
+      if ((typeof models.naturalEvents != 'undefined' &&
```
localjo: This logic looks correct. It took a little bit of time for me to understand the original intention of the code, since there's not a lot of documentation, so let's take the opportunity to make things a little bit more readable here too. What do you think about doing something like this?

```
// Don't add data runners if we're on the events or data tabs
var isEventsTabActive = (typeof models.naturalEvents !== 'undefined' && models.naturalEvents.active);
var isDataTabActive = (typeof models.data !== 'undefined' && models.data.active);
if (isEventsTabActive || isDataTabActive) return;
```

@Benjaki2 does this look right to you too?

localjo:
Thank you for finding this issue, and thank you for the contribution! 🙂 Let me know what you think about my feedback.

web/js/map/wv.map.ui.js
PacoDu: Yes I think this would be much more readable, we could do the same for the map animation condition.

localjo:
It looks like this logic was originally added by @Benjaki2 in this commit: https://github.com/nasa-gibs/worldview/commit/bf4d24cb7367246d821739518f68f3e151020b97 with the intention of disabling data runners when the layers tab is not selected.

It was later updated by @rboller to safely check those conditions;
https://github.com/nasa-gibs/worldview/commit/49ec49615226c605ff55c0509a1be372ec1a56f7#diff-50be1ce8995e932907ce7380a2091738R842
and https://github.com/nasa-gibs/worldview/commit/ddfff2992928b7e8a4e3714106aec73928c65348

@PacoDu If you want to also update the map animation condition at the same time for consistency, I think that would be a valuable contribution. :)

localjo:
Oh! You're one step ahead of me on the map animation updates. :) Looks great!

localjo:
Could you consolidate all of the comments into a single comment at the top of the changed lines, and then group all of the lines together @PacoDu?

PacoDu:
:smile: Yes I just pushed the modification based on your advice. Also modified the `!=` as you noticed in your review (I needed some refresh on javascript :stuck_out_tongue:)

I'll edit the comments for consistency as you proposed

localjo:
@PacoDu Our workflow is to merge into `development` first, and then do frequent releases of `development` into `master`. I've changed the target of this PR to `development`. Could you rebase your changes on the `develoment` branch? I don't think there will be any conflicts when you do that.

localjo:
Never mind, I merged in changes from the `development` branch, and then merged this into development. Sorry about the confusion.

The `development` branch with your changes should be released some time this week.
tumblr_m8x3rg72gy1rpdtxz

PacoDu:
Thanks for the rebase, next time I'll fork from development if I need to develop new fix ;)

PR #629, opened 2017-12-06
Fix dependency errors in core

ZachTRice:
Changes in this pull request:

- Import jquery/effect-slide to resolve jquery easing console errors
- Remove jquery-ui menu li css and override menu ul padding to maintain consistent look since dependency css is different than our previously used version.
- Define `Share` inside of import brackets to fix breaking link button

### DIFF
--- a/web/js/projection/ui.js
+++ b/web/js/projection/ui.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import 'jquery-ui/effect-slide';
 import 'jquery-ui/button';
 import 'jquery-ui/menu';
 import loEach from 'lodash/each';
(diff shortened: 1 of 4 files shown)


web/js/projection/ui.js
```diff
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import 'jquery-ui/effect-slide';
```
localjo: Good catch! I missed that when I `import`ed jquery UI.

ZachTRice: It was a tough one to find tbh.

localjo: Yeah, I went through the entire jquery UI API docs and searched for each function and `import`ed the ones I found we were using, but I guess I missed one. 😬 Thanks for catching it!

PR #631, opened 2017-12-07
Fix jQuery UI and iCheck style changes

ZachTRice:
Changes in this pull request:
 - Fix text appearing darker in events & date tab and measurement setting list
- Fix "Squash Layer" button missing border and padding
- Fix color palettes radio "check" buttons missing thick borders
- Fix "Add Layers" button font changed by default jQuery UI font stack

### DIFF
--- a/web/css/sidebar.css
+++ b/web/css/sidebar.css
@@ -42,6 +42,8 @@
 }
 
 #productsHolder footer .action {
+  font-family: 'Open Sans', sans-serif;
+  font-weight: 700;
   display: inline-block;
   border-color: transparent;
   color: #fff;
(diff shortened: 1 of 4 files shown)


web/css/sidebar.css
```diff
@@ -42,6 +42,8 @@
 }
 
 #productsHolder footer .action {
+  font-family: 'Open Sans', sans-serif;
```
localjo: What was the font stack for this element before? If this is what was being applied before I don't want to worry too much about it, but it seems a little short. Should it have "Arial", "Helvetica" or "Verdana" fallbacks for browsers that don't support web fonts or users who have web fonts disabled? Just a thought, but let's stick with whatever is being applied to this element in production for now.

ZachTRice: I agree, we should add 'Arial' in there. We are using this font stack consistently throughout so we'll keep it like this for now until we can change all references at the same time.

PR #635, opened 2017-12-07
Fix Buster tests

localjo:
This PR updates our Buster tests so that we can run them against our modularized app code to check for regressions. All of the tests have been updated except the ones in `./test/data` since the Data feature hasn't been modularized yet. When we modularize that feature, we can update the tests for it in the same PR.

For each of the tests I updated, I renamed the test file from the `wv.*-test.js` naming convention to the `*-spec.js` naming convention. That was partially in order to remove the `wv` prefix to be consistent with the app code, and the change from `-test` to `-spec` was just a way to tell Buster to run the updated tests, but not the tests which haven't been updated (since Buster has limited glob-matching/ignoring support).

I made a few small changes to the app files, apart from updating the tests, to fix the tests that were broken. We didn't really break too many of the tests with our changes so far. Good job @Benjaki2 and @ZachTRice !


localjo:
Note, to review this PR, run the following;
```
rm -rf node_modules
rm package-lock.json
npm install
npm run build
npm test
```

localjo:
This doesn't depend on `npx`, that was just to try out the commands to see why things weren't working for you.

ZachTRice:
The "test:run" line in package.json needs to be:
`"test:run": "sleep 5 && buster-test && npm run test:end",`
to work properly in Windows as we found experimenting with running these tests on my machine.

test/main.js
```diff
@@ -0,0 +1,58 @@
...
+    parse: projectionParser,
+    model: projectionModel,
+    change: projectionChange
+  },
+  ui: ui,
+  util: util
+};
+
+window.fixtures = fixtures;
+window._ = underscore;
```
Benjaki2: @localjo Is it not possible to use lo-dash here?

web/js/map/ui.js
```diff
@@ -41,7 +41,7 @@ export function mapui(models, config) {
   var createLayer;
   var precache = mapPrecacheTile(models, config, cache, self);
 
-  var dataRunner = self.runningdata = new MapRunningData(models);
+  var dataRunner = self.runningdata = new mapRunningData(models);
```
Benjaki2: @localjo Should this remain Capitalized as it uses a constructor pattern? If we decide against keeping it capitalized, `MapRotate` should probably be lowercase as well.

test/main.js
localjo: No, the Buster tests rely on Underscore, which is a different library. https://github.com/nasa-gibs/worldview/pull/635/commits/89ad4367b94a69ecf92b032cf310d9ddd333d5c7

web/js/map/ui.js
localjo: Good catch.

ZachTRice:
Just to keep this thread up to date with the latest testing...
To end the tests on Windows, we would need to use:
```
    "test:end": "run-script-os",
    "test:end:darwin:linux": "kill $(ps aux | grep '[b]uster-server' | awk '{print $2}')",
    "test:end:win32": "tskill node",
```
The caveat to this approach is that `tskill node` will kill all node processes. I think this should suffice but a more ideal approach would target the specific node PID running the test and only kill that process.

localjo:
@ZachTRice The latest commit should mostly fix the tests on Windows, with the caveat you mentioned that running tests will have the side effect of killing any other node processes. If you come up with a better command to kill the test server on Windows, feel free to add a commit to this branch. Otherwise, I think this is good to go.

@Benjaki2 Do you want to give my latest changes one last look too?

ZachTRice:
@localjo I agree. Good work, coming together on this to get this working on Windows!

PR #638, opened 2017-12-07
Fix measurement description button not showing full text

ZachTRice:
Changes in this pull request:
- Replace legacy jQuery toggle (event) function

### DIFF
--- a/web/js/layers/modal.js
+++ b/web/js/layers/modal.js
@@ -467,6 +467,7 @@ export function layersModal(models, ui, config) {
 
         // Begin source level
         Object.values(current.sources).forEach(function (source) {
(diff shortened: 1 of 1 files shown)


web/js/layers/modal.js
```diff
@@ -467,6 +467,7 @@ export function layersModal(models, ui, config) {
 
         // Begin source level
         Object.values(current.sources).forEach(function (source) {
+          var showFullDescription;
```
localjo: I might name this variable `isExpanded`, but that's subjective.

ZachTRice: Yea I tried to name this logically but I think `isExpanded` is more concise.

PR #639, opened 2017-12-08
Enable tour features

ZachTRice:
This is a WIP...

Changes in this pull request:

- [x] Enable tour config feature in main.js
- [x] Add jquery-ui dialog
- [x] Change name of Google analytics import name to be consistent with codebase
- [x] Add `autoStart: true` to initiate joyride v2.2.0
- [x] Change feedback.js to export default
- [x] Fix stylesheet differences between joyride 2.0 and 2.2 
- [x] Re-implement page counter.*
Note*: The previous version of joyride was customized directly to add this feature:
https://github.com/nasa-gibs/worldview/search?utf8=%E2%9C%93&q=wv-tour-page-count 
- [x] Restore Previous buttons


ZachTRice:
@bking, @localjo I can't seem to get the previous buttons to show up. If you get a chance to break away and look at this feature, here is where I am at:

I see joyride examples with previous buttons here: http://zurb.github.io/joyride/ 

But then I see references to the functionality not working for others, see:
https://github.com/zurb/joyride/pull/184

And it is nowhere mentioned on the official page:
https://zurb.com/playground/jquery-joyride-feature-tour-plugin

Looking at the version we are running on production, it looks like there was significant modification to the previous button code:
https://github.com/nasa-gibs/worldview/blob/e8c796d012ada1f7e9118c27230391dd58042963/web/ext/tour/joyride-2.0.3-3/joyride.js
(If you go through and search for the comment "//BETH" all the code around it seems to have been modified at some point from the original.

Also a few commits related to previous button in the file blame suggest the same
https://github.com/nasa-gibs/worldview/blame/e8c796d012ada1f7e9118c27230391dd58042963/web/ext/tour/joyride-2.0.3-3/joyride.js

---------------------------
If there is no obvious solution here, then it might be best to test the fix linked above, fork the repo, add that fix and then link to our fork in package.json.

Benjaki2:
I took a look and agree that a fork is probably necessary.

The working example (http://zurb.github.io/joyride/) isn't using Jquery-joyride so maybe the non-jquery version would work for us if it isn't a lot of work to migrate?

web/css/tour.css
```diff
@@ -22,6 +22,10 @@
   width: 450px;
 }
 
+.joyride-close-tip {
```
localjo: We didn't have custom `.joyride` CSS before, and [we're including the Joyride stylesheet in this branch already](https://github.com/nasa-gibs/worldview/blob/module-loaders-tour/web/css/main.css#L10).

We shouldn't have to add custom CSS to our stylesheets to achieve style parity with production. @ZachTRice @Benjaki2 I might have missed something when I included the dependency stylesheets. Maybe the paths aren't correct, or maybe they're in the wrong order or something. Both of you have added dependency styles to stylesheets in your PRs, but I think maybe there's a problem with how I did the CSS originally. Have either of you looked into that?

web/pages/tour.html
```diff
@@ -119,6 +119,7 @@ <h3>Timeline</h3>
         <li>Change the interval to "Days" and explore the imagery for that month.</li>
       </ul>
     </div>
+    <p class="wv-tour-page-count">(page 1 of 6)</p>
```
localjo: How were these `<p>` tags being added before?

web/js/feedback.js
```diff
@@ -30,4 +29,4 @@ export const feedbackModal = (function () {
   };
 
   return self;
-})();
+})({});
```
localjo: Why did you move `self` from being defined inside the function to being passed in as a parameter?

web/js/feedback.js
```diff
@@ -7,8 +7,7 @@
 
 import util from './util/util';
 
-export const feedbackModal = (function () {
-  var self = {};
+export default (function (self) {
```
localjo: Why go from a named export to an unnamed `default` export?

web/pages/tour.html
ZachTRice: The plugin was a complete hack before. These were added into the core plugin before. There is no native support for this type of functionality.

web/css/tour.css
ZachTRice: The plugin we were using before was modified to add css directly to the plugin. Here is the css blame: https://github.com/nasa-gibs/worldview/blame/master/web/ext/tour/joyride-2.0.3-3/joyride.css

web/js/feedback.js
ZachTRice: Discussed in slack chat; export constant does not work in this case. We can look at changing this to a named function when we do a review. Util functions are setup in a similar manner as this.

web/js/feedback.js
ZachTRice: Discussed in slack chat; export constant does not work in this case. We can look at changing this to a named function when we do a review. Util functions are setup in a similar manner as this.

web/js/feedback.js
localjo: Does a naming the function change how parameters are passed in? Before, you could call this function with no parameters, `feedbackModal()`, and it would work fine. Now we have to call the function with an empty object for it to work, `feedbackModal({})`. Why can't we do `var self = {}` inside the function like it was before?

ZachTRice: Good call, I have made this change and tested it.

PR #642, opened 2017-12-20
Remove unused Grunt config and document the rest

localjo:
Part of #503 

- [ ] Clean up build scripts
- [ ] Update both development and production build scripts
- [ ] Re-enable watch tasks for build scripts and tests


localjo:
Build scripts need to do the following;
- [ ] Make `GetCapabilities` call
- [ ] Bundle JS
  - [ ] Combine modules into bundle
  - [ ] Convert ES6 to ES5
  - [ ] Shim legacy scripts
  - [ ] Replace string placeholders
  - [ ] Cachebusting + copy referenced assets to build
  - [ ] Minify
- [ ] Bundle CSS
  - [ ] Resolve Imports
  - [ ] Add vendor prefixes
  - [ ] Cachebusting + copy referenced assets to build
  - [ ] Minify
- [ ] Build HTML
  - [ ] Convert Markdown to HTML
  - [ ] Replace string placeholders
  - [ ] Cachebusting + copy referenced assets to build
- [ ] Copy remaining assets and config
- [ ] Build `.tar`
- [ ] Build `.rpm`

localjo:
I ended up going on much more of a rabbit trail of trying to rewrite out build scripts than I thought I would. (We're using Grunt to wrap Bash to wrap Python, which is running in a Virtual Environment...) Long story short, I think I'm going to give up, and add something to the Gruntfile that copies the new bundles into the existing build scripts.

I think when we tackle the overhaul of our build scripts (#439) later, we're going to need to just wipe out the scripts that we have (Grunt, Bash, Python) and start over from scratch. It's probably way easier that way.

PR #647, opened 2017-12-28
Update build scripts

localjo:
- [x] Makes new build scripts work in production
- [x] Enables watch tasks for development (command: `npm run watch`)
- [x] Other build script improvements

Also, fixes #402

PR #648, opened 2018-01-02
Modularize natural events

localjo:
This is based on `update-build-scripts` (#647) which should be merged first.

This PR modularizes the natural events feature.

### DIFF
--- a/web/js/naturalEvents/request.js
+++ b/web/js/naturalEvents/request.js
@@ -1,17 +1,18 @@
-var wv = wv || {};
-wv.naturalEvents = wv.naturalEvents || {};
-/**
- * @module wv.naturalEvents.request
- */
-wv.naturalEvents.request = wv.naturalEvents.request || function (models, ui, config) {
+import $ from 'jquery';
+import orderBy from 'lodash/orderby';
+import uniqBy from 'lodash/uniqby';
+
+import util from '../util/util';
+
+export default function naturalEventsRequest (models, ui, config) {
   var self = {};
-  self.events = wv.util.events();
+  self.events = util.events();
 
   self.EVENT_QUERY_RESULTS = 'queryResults';
   self.EVENT_SELECT = 'select';
 
   self.apiURL = config.features.naturalEvents.host;
-  var querySuccessFlag = false;
+  self.querySuccessFlag = false;
   var model = models.naturalEvents;
   self.ignored = config.naturalEvents.skip || [];
   model.data = {};
@@ -23,7 +24,7 @@ wv.naturalEvents.request = wv.naturalEvents.request || function (models, ui, con
 
   var onQueryResults = function () {
     if (model.data.sources && model.data.types && model.data.events) {
-      querySuccessFlag = true;
+      self.querySuccessFlag = true;
 
       // Remove types and events for ignored event categories
       var removeIgnoredItems = function (item) {
@@ -43,16 +44,16 @@ wv.naturalEvents.request = wv.naturalEvents.request || function (models, ui, con
 
       // Sort event geometries by descending date
       model.data.events = model.data.events.map(function (e) {
-        e.geometries = _.orderBy(e.geometries, 'date', 'desc');
+        e.geometries = orderBy(e.geometries, 'date', 'desc');
         // Discard duplicate geometry dates
-        e.geometries = _.uniqBy(e.geometries, function (g) {
+        e.geometries = uniqBy(e.geometries, function (g) {
           return g.date.split('T')[0];
         });
         return e;
       });
 
       // Sort events by descending date
-      model.data.events = _.orderBy(model.data.events, function (e) {
+      model.data.events = orderBy(model.data.events, function (e) {
         return e.geometries[0].date;
       }, 'desc');
 
--- a/web/js/naturalEvents/request.js
+++ b/web/js/naturalEvents/request.js
@@ -1,6 +1,6 @@
 import $ from 'jquery';
-import orderBy from 'lodash/orderby';
-import uniqBy from 'lodash/uniqby';
+import lodashOrderBy from 'lodash/orderby';
+import lodashUniqBy from 'lodash/uniqby';
 
 import util from '../util/util';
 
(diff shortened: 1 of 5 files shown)


web/js/naturalEvents/ui.js
```diff
@@ -1,12 +1,21 @@
-var wv = wv || {};
-wv.naturalEvents = wv.naturalEvents || {};
-
-wv.naturalEvents.ui = wv.naturalEvents.ui || function (models, ui, config, request) {
-  var self = {}, eventVisibilityAlert, $footer, view;
+import $ from 'jquery';
+import find from 'lodash/find';
```
ZachTRice: We should use `loFind`, `loEach`, etc. as the rest of the codebase is using the `lo` notation. `each`, `find`, etc. are ambiguous keywords that are used by many libraries (lodash, jquery, underscore, etc). It is hard to tell when looking at the code whether a jQuery `each` is being used or a lodash `each` is being used. So regardless of what we call these, it is important to namespace to differentiate them.

localjo: I strongly disagree with that naming convention for a few reasons;
1) It gets confusing in many cases, like [when we have `loValues`](https://github.com/nasa-gibs/worldview/pull/651/files#diff-cd08152820a4d6135bd017b13a0f6f9aR101). Is the function determining low values of it's input?
2) The `lo` abbreviation is ambiguous. It's not clear at all that `lo` is referring to `lodash` except because it's in our recent memory. Future developers, and probably even future us, won't remember what the abbreviation stands for. If we're going to use a prefix, we should use the whole prefix `lodashFind`, etc. Again, what is a future developer going to think when they see a line that says `loValues(someObject)`?
3) Using the same function name that the library uses makes it easier to find/replace that function across the code base, making it easier to refactor and track down where a dependency is used.
4) A function name should describe what the function does, so that the intention of the code is clearly communicated without the developer needing to look up a function definition or library documentation. I can read a line of code that says `find(layers, { 'projection': 'geographic' })` and understand that it's going to find layers with a geographic projection without looking up the function definition or documentation. But if I see a line that says `loMap(layers, id)` it's really a lot more difficult to intuit that it's mapping layers to their IDs. Again, when a future developer or potential open source contributor sees this, they're not going to intuitively understand that `lo` refers to `lodash`.
5) It's really easy to look up a function definition in a file. When you see a function, and you don't know what it does, you probably do `ctrl+f` and search for the function name or right click and choose "Go to Declaration" and find the function definition instantly. If you see `each(layer, renderLayer)` and you aren't sure what `each()` does, but it would take you two seconds to find out that `import each from 'lodash/each';` is at the top of the file.

If you really think it's confusing, we can create a parent object at the top of the file and do something like this;
```
import each from 'lodash/each';
import find from 'lodash/find';
var _ = {
  each: each,
  find: find
};
```

I'd even settle for `lodashEach` and `lodashFind` if you guys insist, but cryptic abbreviated prefixes do not make good variable names.

Either way, our PRs up to this point are using inconsistent naming patterns, so I think we should address this in a cleanup pass on the main module loaders PR.

ZachTRice: 1) I don't think #1 is really an issue but I see the point you are trying to make.  
2) Counter argument to your point here is that `ol` is ambiguous naming but we know that this is openlayers. Is it only ok to use this abbreviated naming when the original author decides to use it? 
3) I disagree completely with this statement. It is much harder to find and replace `each` or `find` in the codebase because it is used by many libraries and can be found within other words like `forEach`, search `loEach`, I can get every instance of that. As I said, it doesn't matter what we namespace this as, whether that is `lodashEach` or `loEach`, we should namespace it somehow so that we can search the entire codebase for exactly what we want.
4) The function is being described what it does but it needs to be namespaced somehow to make it unique.. ideally it would be `_.[name]`, `lo.[name]`, or `lodash.[name]`.
5) "it would take you two seconds to find out that `import each from 'lodash/each';` is at the top of the file.", as would `import loEach from 'lodash/each';` but it is much harder to do what you first suggested and ctrl-f `each` compared to `loEach` because there will potentially be many other matching artifacts throughout the file since `each` is a generalized word that can be used all over the place.

How are these naming conventions any different? Do you consider these cryptic? It is not as though openlayers wants your to just import `feature` or `style`, they namespace. And they name space using a shorter version of `openLayers` so you aren't typing something like `openLayersGeomPolygon`.
+import olExtent from 'ol/extent';
+import OlFeature from 'ol/feature';
+import OlStyleStyle from 'ol/style/style';
+import OlStyleStroke from 'ol/style/stroke';
+import OlLayerVector from 'ol/layer/vector';
+import OlSourceVector from 'ol/source/vector';
+import OlGeomPolygon from 'ol/geom/polygon';

Also, I don't think our PRs have much inconsistency as it stands; just this one library has been a pain point and I'd like to come to a consensus on it.

I would compromise and use `lodashEach`. My main concern is making the names unique in some way so you can search the whole codebase and find what you need. simply using `each` or `find`, doesn't allow you to do that.

localjo: I'm not a big fan of the `ol` prefix either. I'd prefer to just use the native names of `Polygon`, `Vector`, etc. But at least `ol` is the name of the package (you do `npm install ol`), which I think makes it a little more intuitive.

Let's use `lodashEach`, etc. I think that's a reasonable compromise. @Benjaki2 What do you think?

ZachTRice: "But at least ol is the name of the package (you do npm install ol), which I think makes it a little more intuitive.", 

I agree with that statement. And that pushes me towards using `lodash`. I think that is a good argument. I like `lo` because it is short and intuitive _to me_ but 4 extra characters isn't much more to type if it is going to make the code more clear for other developers.

web/js/naturalEvents/ui.js
```diff
@@ -251,12 +259,11 @@ wv.naturalEvents.ui = wv.naturalEvents.ui || function (models, ui, config, reque
 
     if (event.geometries.length > 1) {
       var eventIndex = 0;
-      _.each(event.geometries, function (geometry) {
-        eventIndex++;
+      lodashEach(event.geometries, function (geometry) {
+        eventIndex = eventIndex + 1;
```
ZachTRice: Keep looking at this line wondering why this changed?

web/js/naturalEvents/request.js
```diff
@@ -1,17 +1,18 @@
...
   var self = {};
-  self.events = wv.util.events();
+  self.events = util.events();
 
   self.EVENT_QUERY_RESULTS = 'queryResults';
   self.EVENT_SELECT = 'select';
 
   self.apiURL = config.features.naturalEvents.host;
-  var querySuccessFlag = false;
+  self.querySuccessFlag = false;
```
ZachTRice: Why is this set to `self.querySuccessFlag` now?

ZachTRice:
Approving this but added some questions about a couple changes you made.

web/js/naturalEvents/ui.js
localjo: The linter didn't like `eventIndex++`, but `eventIndex = eventIndex + 1` is the same thing, and the linter doesn't complain.

web/js/naturalEvents/request.js
localjo: The linter was complaining that `querySuccessFlag` wasn't being used, even though it's used around line 26/27. Attaching it to the `self` object instead fixes the linter error and behaves the same way.

PR #651, opened 2018-01-02
Modularize mobile timeline wheel

localjo:
Based on `update-build-scripts` (#647) 

Modularizes mobile timeline wheel. Had to revert `mobiscroll` to 2.6.x and put it back in the `/web/ext` directory. The newer version of mobiscroll requires a license to access full functionality, so we need to stick with the old version until we can replace this dependency.

This PR also fixes various mobile bugs.

### DIFF
--- a/web/js/date/wheels.js
+++ b/web/js/date/wheels.js
@@ -1,5 +1,5 @@
 import $ from 'jquery';
-import 'mobiscroll-jquery';
+import 'mobiscroll';
 import util from '../util/util';
 
 export function dateWheels(models, config) {
--- a/web/js/date/wheels.js
+++ b/web/js/date/wheels.js
@@ -2,7 +2,7 @@ import $ from 'jquery';
 import 'mobiscroll';
 import util from '../util/util';
 
-export function dateWheels(models, config) {
+var dateWheels = function(models, config) {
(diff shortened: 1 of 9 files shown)


localjo:
I'll look into the errors on mobile. Are you seeing the timeline wheel appear on desktop? I wouldn't think it works on desktop.

ZachTRice:
Yeah, the timewheel appears on the responsive desktop view when you click the date in the corner.

ZachTRice:
Note: I am only able to reproduce this when the source is minified.

web/js/date/wheels.js
```diff
@@ -89,3 +89,5 @@ export function dateWheels(models, config) {
   init();
   return self;
 };
+
+export default dateWheels;
```
Benjaki2: @localjo what's the reason for making this a default export instead of exporting the `function`?

localjo: `dateWheels` is a function. These two are more or less equivalent;

```
export default function(){};
```

And

```
var dateWheels = function(){};
export default dateWheels;
```

I don't remember exactly why I did it this way here. Probably just to be clearer by giving a name to what I was exporting.

PR #654, opened 2018-01-08
Modularize Data Download

Benjaki2:
Changes in this pull request:

- [x] Modularize data download code
- [x] Adjust naming of map function exports to match universal styling
- [x] Modularize EDSC files
- [x] Include data download tests


ZachTRice:
Actually, I am seeing the following error:
screenshot_16
To reproduce to this link, click some granules to download, then switching projection. 
http://localhost:3000/?p=geographic&l=VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor,VIIRS_SNPP_Fires_375m_Night(hidden),VIIRS_SNPP_Fires_375m_Day(hidden),MODIS_Fires_Aqua(hidden),MODIS_Fires_Terra,Reference_Labels,Reference_Features,Coastlines(hidden)&t=2015-01-02&z=1&v=-142.86539783640285,-126.84694971129696,110.54085216359715,94.49680028870304&ab=off&as=2008-01-02&ae=2015-01-02&av=3&al=false&download=MOD14

web/js/data/results.js
```diff
@@ -1,45 +1,59 @@
...
+import OlGeomPoint from 'ol/geom/point';
+import olExtent from 'ol/extent';
+import loClone from 'lodash/clone';
+import loEach from 'lodash/each';
+import {dataCmrRoundTime, dataCmrGeometry} from './cmr';
+import util from '../util/util';
+import {dataHelper} from '../edsc';
+import {
+  CRS_WGS_84,
+  mapIsPolygonValid,
```
localjo: What's the difference between prefixing all of these with `map` and prefixing all of the Worldview JavaScript files with `wv.`? Does the extra prefix provide some value? Or are we being more verbose than we need to be?

web/js/data/results.js
```diff
@@ -1,45 +1,59 @@
...
+import OlGeomMultiPoint from 'ol/geom/multipoint';
+import OlFormatGeoJSON from 'ol/format/geojson';
+import OlGeomPolygon from 'ol/geom/polygon';
+import OlGeomPoint from 'ol/geom/point';
+import olExtent from 'ol/extent';
+import loClone from 'lodash/clone';
+import loEach from 'lodash/each';
+import {dataCmrRoundTime, dataCmrGeometry} from './cmr';
+import util from '../util/util';
+import {dataHelper} from '../edsc';
```
localjo: Why not rename the whole `edsc` directory to `web/data/helpers` or something and eliminate the edsc name completely?

web/js/edsc.js
```diff
@@ -1,9 +1,14 @@
+import edscArc from './edsc/arc';
```
localjo: What's the value of the `edsc` prefix? Didn't we agree that `edsc` isn't really relevant to this code anymore?

web/js/data/results.js
```diff
@@ -269,7 +283,7 @@ wv.data.results.dateTimeLabel = function (time) {
   return self;
 };
 
-wv.data.results.densify = function () {
+export function dataResultsDensify() {
```
localjo: What's the difference between prefixing all of the data functions with `data` and prefixing all of the Worldview JavaScript files with `wv.`? Does the extra prefix provide some value? Or are we being more verbose than we need to be?

web/js/data/cmr.js
```diff
@@ -1,15 +1,17 @@
...
- * @module wv.data
- */
-var wv = wv || {};
-wv.data = wv.data || {};
-wv.data.cmr = wv.data.cmr || {};
-
-wv.data.cmr.client = wv.data.cmr.client || function (spec) {
+import $ from 'jquery';
+import util from '../util/util';
+import OlPolygon from 'ol/geom/polygon';
```
localjo: Should this be `OlGeomPolygon` for consistency with the other places you imported it?

web/js/data/ui.js
```diff
@@ -274,47 +283,47 @@ wv.data.ui = wv.data.ui || function (models, ui, config) {
...
+    wvui.notify(
       'No results received yet. This may be due to a ' +
       'connectivity issue. Please try again later.'
     );
   };
 
   var updateSelection = function () {
     var $button = $('#wv-data-download-button');
-    var selected = _.size(model.selectedGranules);
+    var selected = loSize(model.selectedGranules);
```
localjo: I think `loSize` is a very confusing function name to find in a context like this. Wouldn't someone's first intuition here be to think that we're selecting the granule with the lowest size from the model? I think @ZachTRice and I are in favor of spelling out the whole prefix for lodash functions; `lodashSize`, since `lo` can be kind of ambiguous.

web/js/data/results.js
```diff
@@ -337,29 +351,29 @@ wv.data.results.dividePolygon = function () {
...
+      new OlGeomMultiPolygon([resultMultiPoly]);
     return granule;
   };
 
   return self;
 };
 
-wv.data.results.extentFilter = function (projection, extent) {
+export function dataResultsExtentFilter(projection, extent) {
   var self = {};
```
localjo: When a non-geographic projection is selected, it looks like this function gets called with `extent` being undefined.

web/js/map/map.js
```diff
@@ -4,6 +4,10 @@ import loIsUndefined from 'lodash/isUndefined';
 import loFind from 'lodash/find';
 import OlGeomPolygon from 'ol/geom/polygon';
 
+export const CRS_WGS_84 = 'EPSG:4326';
```
localjo: Where was this constant defined before? I suspect that maybe the bug we're seeing is related to this constant.

Benjaki2: here within`self`: https://github.com/nasa-gibs/worldview/blob/master/web/js/map/wv.map.js#L8

web/js/data/data.js
```diff
@@ -0,0 +1,15 @@
+export function dataParser(state, errors, config) {
```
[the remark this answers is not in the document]
Benjaki2: A number of the 'parser' files in other features have other methods in them.

web/js/data/results.js
Benjaki2: I'd say we essentially got rid of the`wv` in `wv.map.something` and are left with the `map.something` as camel case. The prefix stands for the folder in which something is in. We can't just call a method `parser`, we need `dataParser` if we want to export unique names across the board. `mapIsPolygonValid` would be fine as `isPolygonValid` but I'm just trying to stick with a convention.

web/js/data/cmr.js
Benjaki2: Good catch!

Benjaki2:
@localjo @ZachTRice 
I believe #c4d3a68 addresses the error with polar views.

localjo:
Looks good! Let's merge!

PR #659, opened 2018-01-09
Clean Up Build Scripts

localjo:
- [x] Fixes build scripts to work with GIBS options repo
- [x] Changes `NODE_ENV` check to production instead of development

This makes development setup easier for users, since they don't need to set a `NODE_ENV` variable to get started


ZachTRice:
Maybe this logic should be changed so that it doesn't say "Development Version" in the console, even after changing /building with NODE_ENV set to production:
https://github.com/nasa-gibs/worldview/blob/change-node-env-check/web/js/main.js#L335-L341

localjo:
@ZachTRice Let's open a separate PR to handle that line. That's not related to the `NODE_ENV` directory, but it probably doesn't work as intended anymore after our other module loaders changes. I'm not sure how important it is, we could probably remove that block entirely.

I added a PR to this commit that should allow passing in an options subdirectory with the following;
`npm run build -- subdirectory_name`
`npm run build:config -- subdirectory_name`
`npm run getcapabilities -- subdirectory_name`

I'm not 100% sure if it will work on Windows though, so let me know.

ZachTRice:
Adding the directory flag doesn't appear to change the subdirectory for me.

ZachTRice:
*Running `npm run build -- subdirectory_name` doesn't appear to detect the subdirect flag, the other two commands work on my machine.

localjo:
This should work with the GIBS options repo. The command to continue despite errors is;
```
IGNORE_ERRORS=true npm run build -- sit-cache1
```

ZachTRice:
All commands appear to be working as expected on my machine now. 👍

PR #665, opened 2018-01-11
Fix babel polyfills not working in IE and older browsers

ZachTRice:
- Add babel-polyfill import to main.js


web/js/main.js
```diff
@@ -1,3 +1,4 @@
+import 'babel-polyfill';
```
localjo: Maybe we should add a comment on this line that explains why we need to import it;
```
import 'babel-polyfill'; // Needed for worldview-components in IE and older browsers
```

PR #688, opened 2018-01-17
Fix debug -/+ buttons not working

### LINKED ISSUE #681: Layer Debugger Is Broken [layer.render]
In #503, when `debug=layers` is in the URL, the plus and minus buttons to cycle through layers aren't working. Tested in Safari.

ZachTRice:
Fixes #681

Changes in this pull request:

- Change jQuery's `.attr` to `.prop`


ZachTRice:
@localjo, try testing this in Safari and let me know if the debugger works for you. This fixes the issue in Chrome.

PR #697, opened 2018-01-17
Fix style regressions

### LINKED ISSUE #670: Style regressions introduced
In #503 some style regressions were introduced;

*  The "Add Layers" button is now slightly larger and the top padding is off, in all browsers.
* The Tour modal has extra vertical padding added.

### LINKED ISSUE #671: Invalid attribute in meta viewport tag
index.html has a meta viewport tag with a `minimal-ui` attribute. Safari complains that this attribute is invalid.

ZachTRice:
Fixes #670, Fixes #671 

Changes in this pull request:
- Add "Segoe UI" default font back to "Add Layers" button
- Change tour step 1 to be a wide window
- Add `bordered` styles to tour windows which were removed from our version of joyride; 
this requires using an after element on the `nub` element whereas before there was a second element.
- Fix margins in tour windows
- Remove `minimal-ui` viewport attribute

### DIFF
--- a/web/css/jquery-ui-override.css
+++ b/web/css/jquery-ui-override.css
@@ -1,6 +1,7 @@
 /*
  * Default jquery-ui button
  */
+
 a.ui-button.ui-state-default {
   color: #eee;
   font-weight: 700;
@@ -11,6 +12,7 @@ a.ui-button.ui-state-default {
   background-image: -o-linear-gradient(#555, #666);
   font-family: "Segoe UI", "Arial", sans-serif;
 }
+
 a.ui-button.ui-state-default.ui-state-hover {
   border-color: #eee;
   background-color: #eee;
--- a/web/css/tour.css
+++ b/web/css/tour.css
@@ -14,10 +14,17 @@
   background-color: rgb(175, 181, 196);
 }
 
+.joyride-tip-guide p,
+.joyride-tip-guide ul {
+  margin: 0 0 14px;
+}
+
+.joyride-tip-guide.bordered,
 .joyride-bordered {
   border: #808080 solid 1px;
 }
 
+.joyride-tip-guide.wide,
 .joyride-wide {
   width: 450px;
 }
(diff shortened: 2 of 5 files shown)


web/css/jquery-ui-override.css
```diff
@@ -1,6 +1,10 @@
 /*
  * Default jquery-ui button
  */
+.ui-button-text {
+  font-family: "Segoe UI", "Arial", sans-serif;
```
localjo: It looks like the original "Segoe UI" declaration is still there, it's just overridden by [these two lines](https://github.com/nasa-gibs/worldview/blob/module-loaders/web/css/sidebar.css#L45-L46) added in https://github.com/nasa-gibs/worldview/commit/8345cb170710cbe9d5dfaa07358a8cdbb7a18c55 that we could probably remove to be more efficient.

web/css/tour.css
```diff
@@ -431,34 +438,79 @@ p.wv-tour-page-count {
...
 }
 
+.joyride-tip-guide.toolbarTip span.joyride-nub.top {
+  left: 250px;
+}
+
+.joyride-tip-guide.bordered span.joyride-nub.top::after {
+  top: -16px;
+  left: -15px;
+  border: 15px solid #808080 !important;
```
localjo: Are all of these `!important` declarations necessary? These are pretty specific selectors already.

web/pages/tour.html
```diff
@@ -5,7 +5,7 @@
     <p class="splashwelcome">
       This application allows you to interactively browse global satellite imagery within hours of it being acquired. Use the features described below to find interesting imagery, save and share what you find, and download the underlying data.
     </p>
-    <br>
+    <br><br>
```
localjo: I'm not quite sure what caused the extra spacing in production, but I think a single space is fine, and fits the rhythm of the page better. A double `<br><br>` seems unnecessary. What do you think about removing it?

ZachTRice: I agree, I added it back in for parity with what we had before.

web/css/tour.css
ZachTRice: Actually, that's a good point. On the after elements, `!important` should not be needed. I will double-check and remove.

PR #705, opened 2018-01-18
Clean up documentation

localjo:
Update instructions to match recent changes
Fix typos
Clean up formatting


doc/apache_setup.md
```diff
@@ -1,106 +1,18 @@
...
-### Apache Server Configuration
-*These instructions work with OS X El Capitan. Directories and steps may differ with other OS X versions*
-
-Open httpd.conf
-```bash
-sudo vi /etc/apache2/httpd.conf
-```
-Make sure the following modules are enabled:
-```bash
-LoadModule expires_module libexec/mod_expires.so
```
ZachTRice: I think this is important information to retain. Without some documentation about the apache modules needed locally, it would be hard for a user to discern why their apache instance is not properly serving WV.

doc/features.md
```diff
@@ -1,44 +1,28 @@
...
-If you understand the implications of this feature and wish to enable it,
-edit the ``options/features.json`` file and change to:
-
-    "urlShortening": true
-
-Rebuild Worldview and the CGI script should be included in the
-``service`` directory. If your system administrator has disabled
-``.htaccess`` files, configure apache to allow the execution of this
-CGI script.
+  > Caution: Do not commit this file to a public repo, and make sure that the `build/options` directory is not publicly accessible on your web server since your API key should remain secret.
```
ZachTRice: Good note to add here. How about, in the future, we create a githook which checks this directory and prevents committing code if there is sensitive information here. `pre-commit` hooks are normally something an individual has to setup on their machine but maybe we could look into using something like https://github.com/observing/pre-commit, to run an npm script on `git commit`, thoughts?

doc/apache_setup.md
localjo: My two cents is that we shouldn't officially support Apache. We're migrating to Node/Express officially. Worldview will run on any web server, Apache, Express, etc. Pat would like to remove CGI dependencies, and I agree, which would mean we could even host Worldview in a static environment like S3 or the NGAP equivalent. I think providing support for Apache or any other environment than what we'll be using in development and production means spending time and resources that could be better spent on new features. Anyone who insists on using Apache is likely familiar with how to configure an Apache server to serve static assets and execute CGI files. It's a fairly standard Apache configuration and I think we should leave the documentation for how to do it to the Apache community. That being said, I'm flexible on this. One possibility is having a link on this page to Apache docs for how to configure a web app. We can discuss more on Monday.

ZachTRice: I agree that Apache is not something we should focus on. With that being said, Worldview can run Apache currently and it is a configuration that could be in use in some local environments. 

I would argue your point that "anyone who insists on using Apache is likely familiar with how to configure an Apache server to serve static assets and execute CGI files". LAMP stack applications like WAMP, XAMPP, etc. make it extremely easy to get an Apache environment setup locally without having much familiarity with how Apache works. I think it would be a disservice to our users to remove this information as of now until we fully remove support for Apache.

Agree, let's discuss more the future of supporting Apache and come up with a game plan to either keep supporting it our gracefully remove it.

doc/features.md
localjo: How would we detect sensitive information?

From our side, I don't think we need an automatic check for this, since we've already got safeguards in place. We know better than to commit sensitive information. Even if we commit it locally, it's only a problem if that commit gets shared publicly, so we have time to catch it. And if it gets shared publicly we'll catch it during our code review process. In the event that a secret key got shared, we can always invalidate the key and generate a new one.

That being said, it could be a helpful safeguard to add for our users, if there's an easy way to implement detection of sensitive information.

ZachTRice: Your last paragraph sums up exactly what I was thinking; it'd be a nice catch for end-users if it is something we can easily add. Otherwise, there is no point in jumping through a bunch of hoops since WV doesn't contain any other sensitive data.

My idea would be to search for strings that would indicate private information.
See: https://github.com/jonDowdle/git-hooks/blob/master/pre-commit/reject-private-keys.sh

Here are some pre-built hooks using the pre-commit framework I linked above:
https://github.com/pre-commit/pre-commit-hooks

localjo:
I added links to Apache docs, reorganized some of the sections, improved formatting, added a screenshot, and linted for filler words to try to make things clearer. Let me know if this looks good.

PR #707, opened 2018-01-19
Fix sortable overriding layer controls on Android tablets

ZachTRice:
Changes in this pull request:

 - Target `a` elements with sortable's cancel attribute within the layer `li`.


ZachTRice:
On a real Android tablet, I was unable to open the info, options, or close the layer in the sidebar because jQuery sortable was taking precedence. Targeting the a tags within the layer li with sortable's cancel attribute, disables sortable when clicking these areas preventing the issue from occurring.

PR #710, opened 2018-01-19
Prevent tour window from appearing in incorrect position

ZachTRice:
Fixes a bug found during testing where some devices with 768px widths (iPad, Galaxy Tab) were still showing the tour. Because the tour could not fit properly at this size, the first tour step was dropping below the timeline. This PR increases the width detection to `> 800`

Changes in this pull request:

- Increase validScreenSize width detection to prevent the tour window from appearing on devices with too small of an area for the window to fit correctly.

### DIFF
--- a/web/js/tour.js
+++ b/web/js/tour.js
@@ -236,7 +236,7 @@ export default function (models, ui, config) {
       .width();
     var viewHeight = $(window)
       .height();
-    return viewWidth >= 768 && viewHeight >= 680;
+    return viewWidth > 800 && viewHeight > 680;
   };
 
   init();


web/js/tour.js
```diff
@@ -236,7 +236,7 @@ export default function (models, ui, config) {
       .width();
     var viewHeight = $(window)
       .height();
-    return viewWidth >= 768 && viewHeight >= 680;
+    return viewWidth > 800 && viewHeight > 680;
```
localjo: Is this for iPad specifically?

ZachTRice: @minniewong was seeing this on iPad 1st Gen portrait view 😮, I checked the resolution of this device and made the change to greater than 800 to prevent the dialog from occurring on portrait modes where the width is more 768px (iPad) or 800px on some other devices.

PR #713, opened 2018-01-22
Fix toolbar modal positioning

ZachTRice:
Changes in this pull request:

- Change `left` position direction to `right` as per documentation
- Import `jquery-ui/position` to wvui to prevent unexpected results

### DIFF
--- a/web/js/ui/ui.js
+++ b/web/js/ui/ui.js
@@ -1,6 +1,7 @@
 import $ from 'jquery';
 import 'jquery-ui/dialog';
 import 'jquery-ui/menu';
+import 'jquery-ui/position';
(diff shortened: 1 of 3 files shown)


ZachTRice:
This is the issue I was seeing after #503 changes:

screenshot_21

web/js/ui/ui.js
```diff
@@ -1,6 +1,7 @@
 import $ from 'jquery';
 import 'jquery-ui/dialog';
 import 'jquery-ui/menu';
+import 'jquery-ui/position';
```
localjo: This adds 15.5k to the bundle. I think that's ok.

Issue #701, opened 2018-01-17
Update OpenLayers to latest version


ZachTRice:
The following features break when updating to the latest version (4.6.x). We need to address each before we can upgrade.
- [x] Vector layers throw errors when zooming out past maximum zoom level. This error seems to have been fixed here: https://github.com/openlayers/openlayers/pull/7125 which was merged into 4.4.0, but 
Error: `Uncaught TypeError: Cannot read property 'getUnits' of undefined`
Note: This error occurs in **4.3.0-4.3.4** but not in 4.4.0+
Fixed in PR #722 
- [x] Granule labels incorrectly positioned #683 
Note: breaks when upgrading to **4.4.x+**
Fixed in PR #722 
- [x] Can not animate custom colormaps
Note: breaking in **4.5.x+**; 
Fixed in PR #699 (Will need to test again with 4.6.x+)
- [x] Zoom indicator (km & mi) are showing the incorrect numbers
Note: This occurs in **4.6.x+**
Pending fix in https://github.com/openlayers/openlayers/pull/7700 (waiting for a release with this change)

localjo:
Updating also causes a bug with the map scale:

![screen shot 2018-01-18 at 12 58 31 pm](https://user-images.githubusercontent.com/516388/35114408-97155aca-fc52-11e7-9a6a-a43b2200d439.png)

PacoDu:
I was wondering: What made you choose OpenLayer over Leaflet ? Just for curiosity.

rboller:
Hi @PacoDu, we made that decision awhile ago after some careful consideration - Leaflet would've likely also worked for us, though OpenLayers was a better fit in general for our needs.  From what I recall, one of the main points of consideration was better map projection support in OL since we needed to support polar stereographic projections along with geographic.  There also appeared to be a trend of [increasing contributions to OL](https://github.com/openlayers/openlayers/graphs/contributors) and [decreasing contributions to Leaflet](https://github.com/Leaflet/Leaflet/graphs/contributors) during that time.

That said, Leaflet is of course still a great library and we have set of demos with a legacy version of it (v0.7.3) with GIBS here:
https://github.com/nasa-gibs/gibs-web-examples#live-examples

If you're interested in Leaflet and GIBS, you're certainly welcome to bring those examples up to date!

ZachTRice:
[...]

Issue #683, opened 2018-01-15
Granule labels incorrectly positioned [dd.ams.2]


localjo:
In #503 this bug was introduced. Labels on some granules are not correctly positioned when hovering over the granule. Tested in Safari.

<img width="1115" alt="screen shot 2018-01-15 at 12 07 55 pm" src="https://user-images.githubusercontent.com/516388/34954041-f0dd99a4-f9ec-11e7-88e9-4d55da642915.png">

Benjaki2:
Broken in Firefox as well.

ZachTRice:
I am seeing this issue as well on Chrome, Edge and IE. 

Also, when compared to production; production is showing "Large "+" buttons in Data Download on initial load" #528 but the newest version on testing does not have this issue.

localjo:
Looks like this is probably related to [changes in `web/data/map.js`](https://github.com/nasa-gibs/worldview/commit/129035edaa7b935008f543a68501754d8feff226#diff-29ebb84669d5dfc2f0281a7726cc7f6e) but I still haven't tracked it down completely.

localjo:
Thanks @ZachTRice for discovering that reverting to Openlayers v4.3.x fixes the issue. Let's look into what changed in Openlayers 4.4.x and see if we can update something to make this work with 4.4.x.

localjo:
Fix has been merged into `module-loaders` branch.

PR #722, opened 2018-01-27
Fix granule labels incorrectly positioned in OpenLayers 4.4.x+

### LINKED ISSUE #528: Large "+" buttons in Data Download on initial load [dd.sel.1]
When zoomed out, the data download buttons do no scale properly. 
[Issue Example](https://worldview.earthdata.nasa.gov/?p=geographic&l=MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor,MODIS_Terra_Aerosol,Coastlines&t=2013-09-28&z=3&v=-347.0625,-226.40625,347.0625,226.40625&ab=off&as=2013-09-28&ae=2013-10-05&av=3&al=true&download=MOD04_L2)
Internal Test `dd.sel.1`. 
Affects IE11, Chrome, FF. Edge works as expected.

ZachTRice:
Connects to #701, Connects to #683, Fixes #528

Changes in this pull request:

- Upgrade OpenLayers to 4.5.0
- Move hover labels from buttonStyle function to hoverStyle & set exceedLength attribute to true
- Arrange function calls so hover styles are called last resulting in labels overlaying buttons rather than appearing being them

@nasa-gibs/worldview

### DIFF
--- a/web/js/data/map.js
+++ b/web/js/data/map.js
@@ -39,7 +39,7 @@ export function dataMap(model, maps, config) {
   };
 
   var buttonStyle = function (feature) {
-    var dim = getButtonDimensions();
+    var dimensions = getButtonDimensions();
     var image;
     if (model.isSelected(feature.granule)) {
       image = 'images/data.minus-button.png';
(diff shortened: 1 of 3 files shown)


web/js/data/map.js
```diff
@@ -47,37 +47,30 @@ export function dataMap(model, maps, config) {
...
+    return [new OlStyleStyle({
+      image: new OlStyleIcon({
+        src: image,
+        scale: dim.scale
+      })
+    })];
   };
 
   var hoverStyle = function (feature) {
+    var dim = getButtonDimensions();
```
localjo: Since you're creating this variable in a new place, it might be a good time to take the opportunity and rename all of the `dim` variables in this file to `dimensions` so it's easier to read in the future. Your call.

localjo:
Looks good. Let's merge into `develop`! :)

PR #724, opened 2018-01-27
Fix console error thrown when switching projections

### LINKED ISSUE #723: Console error when switching projections
When switching projections the following console error occurs:
`Cannot read property '0' of null`

The issue is that during loading the `coords` variable used in the `onMouseMove` function does not have a default value set.

screenshot_2

ZachTRice:
Fixes #723

Changes in this pull request:

- Set the coords variable to [0,0] by default so that it is not null.

@nasa-gibs/worldview

### DIFF
--- a/web/js/map/ui.js
+++ b/web/js/map/ui.js
@@ -854,6 +854,7 @@ export function mapui(models, config) {
       }
       pixels = map.getEventPixel(e.originalEvent);
       coords = map.getCoordinateFromPixel(pixels);
+      if (!coords) return;
       if (!olExtent.containsCoordinate(extent, coords)) {
         outside = true;
       }


web/js/map/ui.js
```diff
@@ -853,7 +853,7 @@ export function mapui(models, config) {
         return;
       }
       pixels = map.getEventPixel(e.originalEvent);
-      coords = map.getCoordinateFromPixel(pixels);
+      coords = map.getCoordinateFromPixel(pixels) || [0, 0];
```
localjo: I don't fully understand this code, but I think it might be better to add a line after this like;
```
if (!coords) return;
```
Rather than setting a coordinate of `[0,0]`.

ZachTRice: Good eye, thanks

PR #769, opened 2018-03-05
Optimize requests

### LINKED ISSUE #149: Optimize tile requests for multi-day and sub-daily products
Subtasks:
- [x] Create build step in processTemporalLayer.py to detect "monthly" and "yearly" temporal type from GC doc
Pending Fix in: #750 
- [x] Create build step in processTemporalLayer.py to output array of all date ranges from GC doc
Pending Fix in: #750 
- [x] Create branch based on `develop` for optimizing monthly/yearly layer requests
- [x] Create build step to determine timeInterval. (i.e. `P8D`, would be `8`)
- [x] Monthly: Request tiles only for active layers when timeline date matches a date within the dateRanges object and perform request only at what is specified in the dateInterval object.
- [x] Yearly: Request tiles only for active layers when timeline date matches a date within the dateRanges object and perform request only at what is specified in the dateInterval object.
- [x] Daily: Check for odd intervals and only request daily layers when the specified interval has passed (should tie in with the subdaily work below)
- [x] Fix issues with precached events (animations and image download)
- [x] Investigating enabling persistent cache via LocalStorage
- [x] Rebase `subdaily-timeline` on `optimize-requests` branch
- [x] Subdaily: Ensure that daily layers are only requested when needed, based on initial "Daily" work above
- [x] Bug fixes

Overview:
For our new multi-day products (e.g., monthly and annual), Worldview currently requests imagery for every day within a given period even though it only needs to retrieve it once.  This can be optimized by deeper parsing of the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) string from the [GIBS GetCapabilities document](https://gibs.earthdata.nasa.gov/wmts/epsg4326/best/1.0.0/WMTSCapabilities.xml) to better handle these non-daily products.

As also described in #610, we'll need to optimize for sub-daily products (e.g., hourly) so that time slider movements which don't cross a day/month/year boundary won't trigger continuous tile requests for daily/monthly/annual layers.

ZachTRice:
Fixes #149
Changes in this pull request:

- Create build steps to optimize add dateInterval to layer object
- Only request layers when outside of startDate/endDate range + dateInterval bounds
- Ensure precache works as expected with cached dates
- Very amount of precache based on the animation speed

@nasa-gibs/worldview

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -207,7 +207,7 @@
     "ansi-escapes": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.0.0.tgz",
-      "integrity": "sha512-O/klc27mWNUigtv0F8NJWbLF00OcegQalkqKURWdosW08YZKi4m6CnSUSvIZG1otNJbTWhN01Hhz389DW7mvDQ==",
+      "integrity": "sha1-7D6LTp+AZPwCw6ybZfHCdb2o75I=",
       "dev": true
     },
     "ansi-regex": {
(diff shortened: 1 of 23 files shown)


ZachTRice:
The `subdaily-timeline` branch was rebased on top of this work. This work will be included once subdaily has been merged..

PR #803, opened 2018-03-21
Fix build scripts for handling single layer wmts capabilities

### LINKED ISSUE #802: Build scripts can't handle wmts capabilites with a single layer
Build scripts assume that the WMTS service description contains multiple layer. This is not always true, some WMTS endpoint have a single layer configured.

Running `npm run build` with a single layer WMTS configured gives:
```
extractConfigFromWMTS.py: 0 error(s), 0 warning(s), 71 layers for EXWEXs:geographic
Traceback (most recent call last):
  File "tasks/../tasks/extractConfigFromWMTS.py", line 196, in <module>
    error_count, warning_count, layer_count = process_entry(entry, colormaps)
  File "tasks/../tasks/extractConfigFromWMTS.py", line 169, in process_entry
    id = gc_matrix_set["ows:Identifier"]
TypeError: string indices must be integers
```

PacoDu:
Fixes #802 .

Changes in this pull request:

- Check if layer is an OrderedDict (if it's the case GetCapabilities returned a single layer)
- Reduicing code duplication as mush as possible (not possible for tasks/extractConfigFromWMTS.py:141 due to error counts)

@nasa-gibs/worldview

Btw I saw multiple declaration of "id" variable in some build script (extractConfigFromWMTS.py), this is a reserved name in python this should not be used for var name.

### DIFF
--- a/tasks/extractConfigFromWMTS.py
+++ b/tasks/extractConfigFromWMTS.py
@@ -9,6 +9,7 @@ import urllib2 as urllib
 import xmltodict
 import isodate
 from processTemporalLayer import process_temporal
+from collections import OrderedDict
 
 prog = os.path.basename(__file__)
 base_dir = os.path.join(os.path.dirname(__file__), "..")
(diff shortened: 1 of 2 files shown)


PacoDu:
With this fix I'm able to build configuration for both single and multi layer WTMS endpoints

ZachTRice:
@PacoDu This looks great! I did some testing with our normal options repo and some internal/custom options repo we use and all perform as expected. Also, good catch with the `id`'s.

PR #900, opened 2018-04-20
Fix year jump

### LINKED ISSUE #761: Year jumps unexpectedly after guitar pick drag
Steps to reproduce: 
Start app, click on down arrow for the Year, the timeline does not update/change/move to correct year and the imagery does not update either.

Benjaki2:
Fixes #761  .

Changes in this pull request:

- [x] Combine various reverse/forward functions into one `animateByIncrement` function
- [x] Isolate timeouts from one global timeout for all date-change operations to two more specific timeouts.

@nasa-gibs/worldview

### DIFF
--- a/web/js/date/timeline-input.js
+++ b/web/js/date/timeline-input.js
@@ -23,6 +23,226 @@ export function timelineInput(models, config, ui) {
 
   var $incrementBtn = $('#right-arrow-group');
   var $decrementBtn = $('#left-arrow-group');
+
+  var init = function () {
+    $incrementBtn
+      .mousedown(function (e) {
+        e.preventDefault();
+        switch (ui.timeline.config.currentZoom) {
+          case 1:
+            animateByIncrement(1, 'year');
+            break;
+          case 2:
+            animateByIncrement(1, 'month');
+            break;
+          case 3:
+            animateByIncrement(1, 'day');
+            break;
+          case 4:
+            animateByIncrement(10, 'minute');
+            break;
+          default:
+            animateByIncrement(1, 'day');
+        }
+      })
+      .mouseup(stopper);
+
+    $decrementBtn
+      .mousedown(function (e) {
+        e.preventDefault();
+        switch (ui.timeline.config.currentZoom) {
+          case 1:
+            animateByIncrement(-1, 'year');
+            break;
+          case 2:
+            animateByIncrement(-1, 'month');
+            break;
+          case 3:
+            animateByIncrement(-1, 'day');
+            break;
+          case 4:
+            animateByIncrement(-10, 'minute');
+            break;
+          default:
+            animateByIncrement(-1, 'day');
+        }
+      })
+      .mouseup(stopper);
+
+    $(document)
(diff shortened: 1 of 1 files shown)


Benjaki2:
@minniewong Could you test this locally and see if the date-Jump issue is fixed?

web/js/date/timeline-input.js
```diff
@@ -10,177 +11,297 @@ import lodashParseInt from 'lodash/parseInt';
...
-      self.stop();
-    }
-  };
+    $(document)
+      .mouseout(stopper)
+      .keydown(function (event) {
+        if (event.target.nodeName === 'INPUT' || keyDown === event.keyCode) return;
+        switch (event.keyCode) {
+          case util.key.LEFT:
+            if (models.date.selectedZoom === 4) {
```
ZachTRice: This section of the code looks like it needs some latest changes from develop. There was an addition added to v2.3.0 which changed the left / right arrow functionality to reflect the current zoom level.
```
switch (models.date.selectedZoom) {
      case 1:
        animateReverse('year', -1);
        break;
      case 2:
        animateReverse('month', -1);
        break;
      case 3:
        animateReverse('day', -1);
        break;
      case 4:
        animateReverse('minute', -10);
        break;
}
```

web/js/date/timeline-input.js
```diff
@@ -10,177 +11,297 @@ import lodashParseInt from 'lodash/parseInt';
...
+              animateByIncrement(-1, 'hour');
+            } else {
+              animateByIncrement(-1, 'day');
+            }
+            event.preventDefault();
+            break;
+          case util.key.RIGHT:
```
ZachTRice: Same as the above except with `animateForward`

web/js/date/timeline-input.js
```diff
@@ -283,10 +407,10 @@ export function timelineInput(models, config, ui) {
       if ((selectedDateObj > tl.data.start()) &&
         (selectedDateObj <= util.today())) {
         var parent = selected.parent();
-        var sib = parent.next('div.input-wrapper.selectable')
```
ZachTRice: I am not able to tab from the day field back around to the year field. This is due to removing the `selectable` class logic here on this line (and the line below) + the toggleSubdaily function is no longer applying those classes (here: https://github.com/nasa-gibs/worldview/blob/fix-year-jump-3/web/js/layers/active.js#L428-L444)

web/js/date/timeline-input.js
Benjaki2: Good find!

Benjaki2:
@ZachTRice Thanks for the feedback. The requested changes are up. Could you take a again?

ZachTRice:
@Benjaki2 I am still not able to tab correctly on the day time input (it should bring me back to the year input). Also, using the left / right arrow keys on the keyboard while in month / year zoom levels does not increment by month and year. It still increments by 1 day.  *Note, minute increments by 1 hour when I use the left and right arrow which is correct.

PR #981, opened 2018-05-18
Update RPM to build on el7

### LINKED ISSUE #883: Build on RHEL7
Ensure that the RPM builds and deploys properly on RHEL7. Fix any issues with the spec file if not.

mike-mcgann:
## Description

Fixes #883.

Update RPM build to build on Red Hat Enterprise Linux 7 and CentOS 7.

- Use systemctl instead of service
- Fix bug where release was not being filled in. Set to zero. 
- Split out buildRPM.sh from buildAll.sh 
- Use npx for grunt instead of using a global version 
- Root Worldview at /worldview instead of /labs/worldview
- Remove CGI handler
- Update apache config to use "Require all granted" syntax

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,10 +28,10 @@
   "scripts": {
     "browserstack": "node e2e/browserstack.js -c e2e/browserstack.conf.js -e browserstack --skiptags wip",
     "build": "rm -rf build && run-s \"getcapabilities -- {1}\" \"build:config -- {1}\" --",
-    "postbuild": "run-p build:css build:js build:tests && grunt build site",
+    "postbuild": "run-p build:css build:js build:tests && npx grunt build site",
     "build:css": "node ./tasks/buildCSS",
     "build:config": "bash -c 'tasks/buildOptions.sh ${1}' 0",
-    "postbuild:config": "grunt config",
+    "postbuild:config": "npx grunt config",
     "build:js": "node ./tasks/buildJS",
     "build:tests": "node ./tasks/buildJS --tests",
     "e2e": "run-s --continue-on-error --silent e2e:*",
--- a/tasks/buildAll.sh
+++ b/tasks/buildAll.sh
@@ -8,14 +8,5 @@ rm -rf build dist node_modules .python web/build
 npm install --production=false # Install dependencies, set production false so devDependencies are installed even though NODE_ENV is set to production
 npm run build # Build the app
 
-# Prepare rpm sources
-mkdir -p build/rpm/{SOURCES,SPECS}
-cp worldview.spec build/rpm/SPECS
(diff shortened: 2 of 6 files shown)


tasks/buildAll.sh
```diff
@@ -8,14 +8,5 @@ rm -rf build dist node_modules .python web/build
 npm install --production=false # Install dependencies, set production false so devDependencies are installed even though NODE_ENV is set to production
 npm run build # Build the app
 
-# Prepare rpm sources
-mkdir -p build/rpm/{SOURCES,SPECS}
-cp worldview.spec build/rpm/SPECS
-cp dist/*.tar.bz2 build/rpm/SOURCES
-cp *.conf build/rpm/SOURCES
+tasks/buildRPM.sh
```
ZachTRice: 👍

package.json
```diff
@@ -28,10 +28,10 @@
   "scripts": {
     "browserstack": "node e2e/browserstack.js -c e2e/browserstack.conf.js -e browserstack --skiptags wip",
     "build": "rm -rf build && run-s \"getcapabilities -- {1}\" \"build:config -- {1}\" --",
-    "postbuild": "run-p build:css build:js build:tests && grunt build site",
+    "postbuild": "run-p build:css build:js build:tests && npx grunt build site",
     "build:css": "node ./tasks/buildCSS",
     "build:config": "bash -c 'tasks/buildOptions.sh ${1}' 0",
-    "postbuild:config": "grunt config",
+    "postbuild:config": "npx grunt config",
```
ZachTRice: Do we need to `npm i --save-dev npx` (Add it to the package devDependencies)? and/or do we need to add npx install globally in the travis.yml file?

worldview.spec
```diff
@@ -1,6 +1,6 @@
 Name: @WORLDVIEW@
 Version: @BUILD_VERSION@
-Release: @BUILD_RELEASE@.@BUILD_NUMBER@%{?dist}
+Release: 0.@BUILD_NUMBER@%{?dist}
```
ZachTRice: I'm curious why "0." here?

package.json
mike-mcgann: It comes packaged with node.js but this is only something I've discovered recently.

worldview.spec
mike-mcgann: Since we are both the developers and the packagers, it can stay at zero. It is a separate version number that can get bumped if the packaging needs an update when the version of the application hasn't changed. Current scripts were pulling that value from package.json but it appears that value was removed at some previous time. If we aren't going to use it, might as well be zero.

PR #1066, opened 2018-06-13
Fix dual colorbar units overlapping

### LINKED ISSUE #1065: Colorbar units on Cloud Effective Radius layers are unreadable in water phase
### Description

The colorbar units on the Cloud Effective Radius - Water Phase are unreadable. I think the range is supposed to be 4.0um to 30.0 um. It looks like the first number is on top of another number. The intervening units are readable when you hover over the colorbar.

screen shot 2018-06-13 at 3 09 56 pm

### Environment

Version: 2.5.1 (production)
Browser: Chrome Version 67.0.3396.87 (Official Build) (64-bit) | Firefox 52.8.0 (64-bit

### LINKED ISSUE #1067: Colorbars on dual legend layers are not working independently
### Description

Colorbars on dual legend layers like Cloud Effective Radius and Cloud Water Path should work independently of each other. The colorbar units are missing for the top legend and the cursor is not querying the layer independently of the bottom legend, instead they are pointing to the same place on the legends when you hover over the map.

If you hover over the blue/greens on the map, the Ice Phase colorbar pointer should point to that area on the color bar, and display the units. If you hover over the orange/red area, the Water phase colorbar pointer should point to the right place on the lower legend. 

screen shot 2018-06-13 at 3 24 44 pm

### Environment

Version: 2.5.1  
Browser: Chrome Version 67.0.3396.87 (Official Build) (64-bit) | Firefox 52.8.0 (64-bit

ZachTRice:
## Description

Fixes #1065, Fixes #1067 

- Adds position relative to wrapping div (for #1065)
- Target specific IDs to add/remove running data values

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/palettes/legend.js
+++ b/web/js/palettes/legend.js
@@ -58,7 +58,7 @@ export function palettesLegend(spec) {
 
   var renderScale = function ($legendPanel, legend, index, layerId) {
     var $container = $('<div></div>')
-      .attr('id', legend.id)
+      .attr('id', layerId + '-' + legend.id)
       .addClass('wv-palettes-legend')
       .attr('data-index', index);
     var $colorBarCase = $('<div></div>')
--- a/web/js/palettes/legend.js
+++ b/web/js/palettes/legend.js
@@ -58,6 +58,7 @@ export function palettesLegend(spec) {
 
   var renderScale = function ($legendPanel, legend, index, layerId) {
     var $container = $('<div></div>')
+      .attr('id', legend.id)
       .addClass('wv-palettes-legend')
       .attr('data-index', index);
     var $colorBarCase = $('<div></div>')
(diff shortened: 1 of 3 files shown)


web/js/palettes/legend.js
```diff
@@ -58,6 +58,7 @@ export function palettesLegend(spec) {
 
   var renderScale = function ($legendPanel, legend, index, layerId) {
     var $container = $('<div></div>')
+      .attr('id', legend.id)
```
Benjaki2: @ZachTRice You might want to mix the layer-Id with the legend-Id to create a unique identifier. I think some layers use the same legend and when more than one of these layers are present in the sidebar there can be problems selecting the correct one.

ZachTRice: @Benjaki2 the legend.id is a unique identifier. This value is what is in the config/palettes/[layer].json id field + an index count starting from 0 + '_legend', so even layers with the same layerId will use this as a way to uniquely identify each legend. 
screenshot_17

ZachTRice:
@Benjaki2 I have updated the ID's to append the layerId

PR #1116, opened 2018-06-26
Fix 0px sidebar padding-bottom causing layer button to move on load

### LINKED ISSUE #1114: Add layers button jumps down on page load
### Description

On page load, the '+ Add Layers' button jumps down.

### Steps to Reproduce

1. Load page and watch '+ Add Layers' button - compare to production

### Environment

Version: 2.6.0 UAT
Browser: [ Windows Chrome ]

ZachTRice:
## Description

Fixes #1114

Add padding-bottom to the sidebar so that the Add Layers button doesn't jump down on the initial load. Also adjust the timeout so that the padding-bottom is applied quicker when switching tabs.

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/css/products.css
+++ b/web/css/products.css
@@ -1,5 +1,5 @@
 #productsHolder .ui-tabs-panel {
-  padding: 0;
+  padding: 0 0 37px;
   position: relative;
 }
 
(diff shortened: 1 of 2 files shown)


ZachTRice:
@minniewong I think the mobile jump will need to be addressed in a follow up PR as this must fire a function to resize the sidebar rather than the fix I implemented which set a default height on load.

PR #1513, opened 2018-11-21
Show/select first valid available layer measurement in product picker

### LINKED ISSUE #1511: Layer picker incorrectly showing unavailable layers in polar views
**Describe the bug**
Brightness Temperature in the Arctic and Antarctic views incorrectly show the AMSU-A layer description when the AMSU-A layers are not available in the polar views. 

**To Reproduce**
1. Go to Arctic view
2. Click on Add Layers
3. Scroll down to Brightness Temperature and open it (same issue with Freeze/Thaw)
4. No satellite/sensor is highlighted and it shows the layer description for AMSU-A layers which are not available in the polar views

**Expected behavior**
Layer picker should only show layers and layer descriptions that are available in the polar projections

**Screenshots**
screen shot 2018-11-20 at 2 20 53 pm

 - OS: Mac OSX
- Browser Chrome
 - Version 70.0.3538.102 (Official Build) (64-bit)
Production (v2.10.0)

edplato:
## Description

Fixes issue that happened when a layer measurement subgroup item was available in the Geographic projection but not available in the Arctic projection. This issue caused the unavailable Geographic measurement description to be displayed in Arctic and the correct Nav items to not be highlighted.

Fixes #1511  .

This fix first checks for and then uses a valid matching index of the measurement subgroup items while rendering the source data.

## Checklist

## Further comments

The subtitles for the measurements in the product picker are hardcoded in the `wv.json` and do not differ from projection to projection. We may want to consider changing/updating this in the future.

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/layer/measurement-row.js
+++ b/web/js/components/layer/measurement-row.js
@@ -266,9 +266,9 @@ class LayerRow extends React.Component {
     const { activeSourceIndex } = this.state;
     const sources = lodashValues(measurement.sources);
 
-    // handle activeSourceIndex values not valid after projection change
-    let validIndexSelected = false;
+    // set first valid index to handle invalid activeSourceIndex indexes after projection change
     let minValidIndex = -1;
+    let validActiveIndex = activeSourceIndex;
 
     const Tabs = (
       <Nav vertical className="source-tabs col-md-3 col-sm-12">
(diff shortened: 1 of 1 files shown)


web/js/components/layer/measurement-row.js
```diff
@@ -266,27 +266,65 @@ class LayerRow extends React.Component {
...
+                }
+                // only set minValidIndex once
+                if (minValidIndex < 0) {
+                  minValidIndex = index;
+                }
+              }
+              return [source, hasSetting];
+            }
+          )
+          .map(
```
Benjaki2: @edplato Looks good. Only feedback is that this could probably be all in one `map()` to prevent the double looping. Though that shouldn't make a visible difference in this case since the array shouldn't ever be very large.

edplato: Good recommendation, I slimmed down my changes to only one `map()` and removed some extra logic.

Benjaki2: Nice! looks good

PR #1521, opened 2018-11-27
Add data download e2e tests

edplato:
## Description

- Add data download e2e tests to automate some of that part of the Test Suite.

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/e2e/features/data-download/data-download-test.js
+++ b/e2e/features/data-download/data-download-test.js
@@ -1,0 +1,99 @@
+const reuseables = require('../../reuseables/skip-tour.js');
+const layersTab = '#layers-sidebar-tab';
+const dataDownloadTabButton = '#download-sidebar-tab';
+const zoomInButton = '#wv-map-geographic > div.wv-map-zoom-in.wv-map-zoom.ui-button.ui-corner-all.ui-widget';
+const zoomOutButton = '#wv-map-geographic > div.wv-map-zoom-out.wv-map-zoom.ui-button.ui-corner-all.ui-widget';
+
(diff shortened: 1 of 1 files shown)


e2e/features/data-download/data-download-test.js
```diff
@@ -0,0 +1,99 @@
...
+
+    // On Valid Data Select, 'No Data Available' indicator disappears and data layer is in sidebar
+    client.clearValue('#year-animation-widget-main');
+    client.setValue('#year-animation-widget-main', ['2013', client.Keys.ENTER]);
+    client.expect.element('#indicator').to.not.be.visible;
+    client.expect.element('#wv-data-MOD04_L2MODIS_Terra_Aerosol').to.be.visible;
+
+    // On Valid Data Select, 'No Data Available' indicator appears
+    client.clearValue('#year-animation-widget-main');
+    client.setValue('#year-animation-widget-main', ['2022', client.Keys.ENTER]);
```
Benjaki2: Looks like this `setValue()` is breaking `IE` and `Safari`.

edplato: It appears that the `inputValue` array with `client.Keys.ENTER` is causing the issue on some browsers. I also tried changing the input values and pressing `TAB` after setting the value, but that didn't work either. I refactored with a naive solution that clicks the up/down arrows on the timeline to change dates. Let me know if you think this will work for us.

Benjaki2: Seems like a good solution  👍

Issue #101, opened 2016-12-22
Catalog Of Worldview Examples


Benjaki2:
Thanks for the feedback @mstuenkel

Comment by @ZachTRice:
Here is a list of potential events to be story-boarded. Adding a check beside the event means the event has been completely story-boarded out. This list will need to be narrowed down to a 9 events initially.

Event Types
---
- Default (No Event Type)
- Wildfire
- Volcano
- Snow
- Sea and Lake Ice
- Iceberg
- Water Color
- Dust and Haze
- Severe Storm
- Man Made

Top 9 Latest Events
---
1. **Hurricane Florence - 2018**
Type: Hurricane
Reason for showcasing (optional):
Features to highlight:
2. **California Wildfires & British Columbia Wildfires - 2018**
Type: Wildfire
Reason for showcasing (optional):
Features to highlight:
3. **Snow in South Africa - August 2018**
Type: Snow
Reason for showcasing (optional):
Features to highlight:
4. **Saharan Dust Storms - 2018**
Type: Dust and Haze
Reason for showcasing (optional):
Features to highlight:
5. **Sinabung Volcano - February 2018**
Type: Volcano
Reason for showcasing (optional):
Features to highlight:
6. **Hurricane Maria - September 2017**
Type: Hurricane
Reason for showcasing (optional):
Features to highlight:
7. **Larsen C Ice Shelf / Iceberg A68A - July 2017**
Type: Iceberg
Reason for showcasing (optional):
Features to highlight:

and 2 more events of different event types

Other Potential Events:
---
10. **Shrinking Aral Sea - 2001-2018**
Type: Man Made (?) 
Reason for showcasing:
Features to highlight:
11. **Eyjafjallajökull Volcano Eruption - 2010**
Type: Volcano
Reason for showcasing:
Features to highlight:

rboller:
A few more thoughts after discussing this idea again today.

We should consider providing a set of use cases for new users of the app since many not be familiar with what's possible. This would be distinct from the tour but related to it - so rather than showing how the app works, it would show how the imagery can be used for science and NRT applications.

We may also want to unobtrusively (i.e., not a modal) provide this kind of guidance to first-time users who arrive via permalink.

ZachTRice:
Updating the first post of this ticket with a checklist of events to be storyboarded

ZachTRice:
Action for 9/13/2018 Meeting:
- Create final list of 9 events for initial tour creation
[...]

PR #1618, opened 2019-01-29
New Feature: Tour Overhaul

### LINKED ISSUE #1354: Implement startup tour modal & steps
Implement the tour and stories, (outlined in ticket #101 and internal excel sheet), using bootstrap/reactstrap modals with css transitions.

Add any relevant comments regarding the tour functionality in this ticket.

To Do:
- [x] Split stories.json out into separate files and compile into 1 file on build.
- [x] Add: When an old tour is being accessed via a permalink, display a modal saying Sorry, this tour is no longer supported"
- [x] Limit tours to 9
- [x] Add documentation
- [x] Allow pasting the full link into the steps description parameter
- [x] When an old tour is being accessed via a permalink, display a modal saying `"Sorry, this tour is no longer supported".`
- [x] Add GTM tags & test on UAT
- [x] Add/test unit tests
- [ ] Add/test end to end tests
- [x] Add manual test plan pages

======================================================
Major Bugs:
- [x] Comparison feature breaks when two layers of the same type are loaded in A and B states
- [x] Tour permalink parameter is not saving
- [x] If a step has a custom palette and the next step has no custom palette info, the palette does not revert back to default palette

======================================================
Additional Suggestions / Fixes (1/8/2019)
- [x] Make the move from one part of the world to another smoother (like events)
- [x] Change color of links within the dark blue in-progress modal steps (default category)
- [x] Adjust number of start up stories based on vertical window height
- [x] Ensure in-progress modal does not block +/- window when vertical window height is decreased
- [x] Change extra step of in-progress modal to not account for last window (When there is 8 steps it says 9)
- [x] When you have a tour step and scroll down, the next step will be scrolled down to the bottom instead of starting at the top of the text.
- [x] "Read more" links on final dialog should open in new window
- [x] Create a Worldview features category

======================================================
Additional Suggestions / Fixes (1/10/2019)
- [x] Destroy animation widget when starting a new tour (and any other transitional-type elements)
- [x] Destroy comparison mode when a new story has been selected
- [x] Change link text to regular and not bold
- [x] Squashed, min/max, and palettes are not being applied to layers

======================================================
Additional Suggestions / Fixes (1/24/2019)
- [x] Comparison mode persists once opened and then you go back a step

======================================================
Additional Tour Enhancements: https://github.com/nasa-gibs/worldview/issues/1613

ZachTRice:
## Description

Fixes #1354, #101

Note: A couple of tours still need to be added by @minniewong 

## Misc suggestions:
- [x] Check tour on 13" Macbook 2560px x 1600px OS 10.14.2 FF/Chrome
- [x] Change Intro text to:
"Click an event below to analyze the event in great detail within the application. These guides will walk you through new and create ways to use Worldview."
- [x] Change checkbox text to:
"Check the box to hide this modal until a new story has been added."
- [x] Change “Start Tour” in “i” menu to something else “Explore Worldview”?
- [x] Use the first step back arrow to access the startup modal (Change icon)?

## Checklist

## Further comments

@nasa-gibs/worldview


web/js/sidebar/ui.js
```diff
@@ -35,7 +35,7 @@ export function sidebarUi(models, config, ui) {
       trailing: true
     });
     var layerAdd = function(layer) {
-      if (!ui.tour.resetting && !ui.naturalEvents.selecting) {
```
Benjaki2: @ZachTRice It looks like `sidebar` component is rendering many times on each step change, slowing things down. I would suggest something like this being applied on step change, just allowing the component to render one time after the state of everything else has been updated.
screen shot 2019-01-31 at 11 24 55 am
`Redux` implementation should prevent us from needing to this kind of hacky event-listener workaround 🤞

ZachTRice: I think this is going to need to re-render each time because it is loading a new link (state) with a group of layers. So, each step, the layers/model/load method is called with that currentState passed in which includes the group of layers. When there are, for example, 8 layers in a link, self.add is ran for each and updateLayers() is then called 8 times. I'm not sure how I would get around this, as it needs to make sure each layer is in the sidebar with the correct properties enabled, hidden, opacity, colormaps etc.

PR #1700, opened 2019-04-30
Fix tour breaking when using manipulating comparison mode b-state

### LINKED ISSUE #1685: Manipulating B-layers causes unwanted side-effects while loading tour steps
**Describe the bug**
While in comparison mode, if the B state layers are maniupulated (removed etc) and then a tour is started or the next tour step is loaded; the layers load the incorrect layers.

**To Reproduce**
1. Start A|B Mode
2. Remove layers from B-side
3. Click Explore Worldview (Tour link)
4. Load a tour
5. Observe incorrect layers loaded

**Expected behavior**
Changing the b-layer state should have no effect on what is loaded in the tour.

ZachTRice:
## Description

Fixes #1685

- If mode has switched to B state by the user, switch back to a state before parsing the tour step.
- Add some validation if there are no layers in the layer list (additional bug breaking in current production)

## Checklist

## Further comments

@nasa-gibs/worldview

### DIFF
--- a/web/js/map/ui.js
+++ b/web/js/map/ui.js
@@ -565,7 +565,8 @@ export function mapui(models, config) {
         id: def.id
       }
     });
-    if (!layer && layers[0].get('group')) {
+
+    if (!layer && layers.length && layers[0].get('group')) {
       let subGroup, olGroupLayer;
       lodashEach(layers, layerGroup => {
         if (layerGroup.get('group') === layerGroupStr) {
--- a/web/js/components/tour/modal-tour-in-progress.js
+++ b/web/js/components/tour/modal-tour-in-progress.js
@@ -241,7 +241,8 @@ class ModalInProgress extends React.Component {
           selectedB = new Date(selectedB);
           models.date.selectedB = selectedB;
         }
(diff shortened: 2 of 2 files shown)


web/js/map/ui.js
```diff
@@ -565,7 +565,8 @@ export function mapui(models, config) {
         id: def.id
       }
     });
-    if (!layer && layers[0].get('group')) {
+
+    if (!layer && layers.length && layers[0].get('group')) {
```
Benjaki2: Might want a `layers.length > 0` as well to ensure layers[0] exists

ZachTRice: `layers.length` and `layers.length > 0` are the same. The former being slightly faster.
ref: https://stackoverflow.com/questions/32911424/array-length-vs-array-length-0

Benjaki2: Whoops. Did not know. 👍

ZachTRice: All good 😉 let me try rebasing on the new tour story and see if I can reproduce

ZachTRice:
Hey @Benjaki2 I just realized the reason you are seeing an error with the new tour is because I added a fix in Minnie's branch. I just added that fix over in this branch so if you check this PR again with the new tour, you should not see it breaking. 

Really good catch!

PR #1859, opened 2019-07-16
Account for legacy parameters and add unit tests

### LINKED ISSUE #1834: URL parameter time is ignored (SIT v3.0)
**Describe the bug**
This may not be necessary if we don't feel the need to support it, but the `time` url parameter is not recognized and ignored. We use `t` for time, however I encountered a test link that uses `time`, so we may want to make sure both work (as in production) for older links that still use `time` instead of `t`.

Benjaki2:
## Description

Fixes #1834 URL parameter `time` is ignored

Additional changes
- [x]  legacy `switch` parameters + unit tests
- [x]  legacy `dataDownload` parameters + unit tests
- [x] legacy `products` parameters + unit tests

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/fixtures.js
+++ b/web/js/fixtures.js
@@ -70,52 +70,52 @@ fixtures.getState = function() {
     vectorStyles: {
       custom: {
         OrbitTracks_Aura_Ascending: {
-          'version': 8,
-          'name': 'Orbit Tracks',
-          'sources': {
-            'OrbitTracks_Aqua_Ascending': {
-              'type': 'vector',
-              'tiles': [
+          version: 8,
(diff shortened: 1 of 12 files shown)


mike-mcgann:
Its more than the "time" parameter, "switch" doesn't seem to work either. All of the previous parameters need to be supported or existing links out there will break--check the former unit tests to see how they were handled. The full query string that Ed provided in #1834 should work as that is still in use.

Benjaki2:
I fixed the legacy parameters that I found to be missing and added unit tests. Please review again when you all get the chance. Thanks!

PR #1921, opened 2019-07-26
Update gif component to store crop boundaries state 

### LINKED ISSUE #1908: Animated GIF bounding box should remain in the same place/same size as the previous time you used it (UAT v3.0)
**Describe the bug**
When creating an animated gif and you draw your bounding box, if you exit the "Create Animated GIF" window or if you create a GIF then return to create another GIF, the bounding box you drew earlier should remain (this is what image download/camera does)

**To Reproduce**
1. Go to UAT
2. Click on video camera icon
3. draw a larger bounding box
4. Exit "create an animated gif" by clicking on the x
4. Click on "Create an animated GIF" icon again and See error - the previous bounding box should have remained.

**Expected behavior**
The bounding box should remain the same within that session and not return to the default. If someone is trying to make multiple animated GIFS of the same area over different time periods, they should be able to use the same bounding box.

 - OS: OSX
 - Browser: Chrome, Firefox
 UAT v3/0

jasontk19:
## Description
Hook up bounding box state to redux store so we can restore position when component is re-opened

Fixes #1908.


jasontk19:
@minniewong I noticed this too and was about to make a separate ticket but I can spend a little while seeing if it's an easy fix to include here.

jasontk19:
Ok @minniewong @Benjaki2, the modal properly handles positioning when the crop tool is left near the right side edge of the screen.  There are still some issues to fix if we are comparing other behaviors of this tool to what's in production though:

1. Modal is not draggable anymore 
2. If you drag the crop area to fill the entire horizontal width, the modal goes entirely offscreen and can't be accessed.  

I can keep looking into fixing these other cases if we think this should be fixed for v3.0

web/js/containers/gif.js
```diff
@@ -266,15 +266,28 @@ class GIF extends Component {
     if (top + height > screenHeight) {
```
Benjaki2:
```suggestion
   if (left + width > screenWidth && x - 20 - width < 0) {
      left = left - width;
    }
  if (top + height > screenHeight) {
```
I think preventing the modal from going outside the viewport is a good call for this release, Having the modal be draggable doesn't seem as urgent.

jasontk19: Sounds good.  I added a little bit more handling for cases when the crop area was too wide to leave room for the modal.   Now the modal will move to the top/bottom of the crop area instead of overlap and only overlaps if there's absolutely nowhere else to go.

Benjaki2: Nice!

PR #1997, opened 2019-08-08
Subdaily polish

### LINKED ISSUE #1941: Sub-daily polish
Verify that sub-daily requests work with image download, animation, and animated GIF generation and apply fixes as needed. Ticket WVS-40 has a link to a sub-daily layer that can be used for testing.

jasontk19:
## Description

To test with sub-daily layers, you can checkout [this branch](https://github.com/nasa-gibs/worldview/compare/master...subdaily-test-config) and run `npm run build`, then switch to this PR branch and run the app without building.

* Update image download feature to include date with the timestamp when making a snapshot request to support sub-daily layers
* Update gif download feature to  include date with the timestamp when making a snapshot request to support sub-daily layers
* Re-use the interval selection widget that the timeline uses in the animation widget in order to provide a more consistent UX between the two components.
* Fix the customSelected property on location state to be parsed properly as a Boolean instead of a string.
* Remove unused animation-widget component

Fixes #1941.

## Checklist

@nasa-gibs/worldview


web/js/containers/animation-widget.js
```diff
@@ -1,23 +1,23 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import {
-  without as lodashWithout,
   find as lodashFind,
   get as lodashGet
 } from 'lodash';
+import googleTagManager from 'googleTagManager';
```
edplato:
```suggestion
```

jasontk19:
> Looks great!
> 
> I am seeing one error where the custom widget one allow you to make sub-daily custom selections
> (e.g. 10 min)
> Screen Shot 2019-08-09 at 3 16 00 PM

Yup, good catch!

> Looks great!
> 
> I am seeing one error where the custom widget one allow you to make sub-daily custom selections
> (e.g. 10 min)
> Screen Shot 2019-08-09 at 3 16 00 PM

Good catch! Should be fixed now.

web/js/containers/animation-widget.js
jasontk19: Looks like we are still using this to track gif creation.  I just moved it up in the list of imports by a few lines.

jasontk19:
@edplato Thanks for taking a close look at this!

> Nice changes and code cleanups as you added subdaily support to these features. Here are my thoughts and/or recommended changes:
> 
> I'm not sure about having two Custom Interval Selector widgets. I'm partial to the implementation where if the user clicks `CUSTOM` in the animation widget, the Custom Interval Selector over the timeline date selector opens. If you do go the two Custom Interval Selector widget route, there needs to be consistent updating in the event a user has both widgets open and makes a delta change in one widget where currently the other widget delta doesn't update to reflect that change, but I still lean towards to only use the original Custom Interval Selector widget located below the sidebar.

I felt that having the custom selection widget open above the timeline when you clicked `CUSTOM` above the animation widget felt disjointed to me. If you can select between minute/hour/month/day/year intervals from this widget, you shouldn't have to go elsewhere to set a custom delta/interval when you select `CUSTOM` from the list IMO.  In playing around some more with both of them open I see that there are some inconsistencies in keeping them in sync which I will take a look at.

What are your thoughts on only allowing one custom selection widget to be open at once?  For example if you open the one above the timeline then, select "CUSTOM" from the animation widget the one above the timeline would close.  There's no practical reason to have both open at the same time anyway as far as I can tell.

> If you set the custom interval for `23 DAYS` and refresh the page, `23 DAYS` isn't selected after page load. It's still available in the interval list, but it should be selected.

Good catch, I'll take a look.

> I see `subDailyMode` and `hasSubdailyLayers` are slightly different, but in some cases it makes more sense to leave the `hasSubdailyLayers` naming for consistency (`TimeScaleIntervalChange` for example).

Yeah I initially added this prop in the animation widget because I was determining the value based on whether subdaily layers were enabled /and/ a time interval of >=3 was selected.  Then I started changing the name on a few of the child components because to me it seemed like some of these components don't care about or, have anything to do with, map layers  and so a name like `subDailyMode` was more intuitive. This makes the component feel "cleaner" and more re-usable to me.  It does end up kind of confusing between components though in the current state. The prop we pass to `TimeScaleIntervalChange` and `CustomIntervalSelectorWidget` should probably be renamed back to `hasSubdailyLayers`.

> If I have a subdaily layer is in sidebar and `15 MINUTE` custom interval is selected in the Animation Widget, then I remove the subdaily layer from sidebar, I am still seeing `15 MINUTE` selected and available in the Animation Widget interval list. The increment should change from minute/hour to a non subdaily interval. 

Yup, good catch.  Will need to fix this!

> Minor style change:
> The current interval text in the Animation Widget is slightly larger and pushes other elements down which most noticeably removed the empty space below the down arrows use in the date selectors. It looks like the line-height and/or font size may be causing this, but might have to play around with it.

I noticed this looked a little cramped.  I will probably just add some height to the modal to give some space below the date selectors.  I like the current spacing around the increment selector and would like to keep the font size consistent with this component in the timeline.

jasontk19:
Should be a quick fix... standby!

PR #2022, opened 2019-08-13
headless e2e testing

jasontk19:
## Description

Add configs to the `nightwatch.conf` for Chrome and Firefox headless so we can optionally run e2e tests in the background if we want.  This is helpful when running e2e locally as you don't have to deal with the browser window constantly reopening in your desktop space.

Just run `npm run e2e:headless:chrome` or `npm run e2e:headless:firefox`

Both seem to complete faster than non-headless versions, though the chromeHeadless run is significantly faster. 

Unfortunately, a few additional tests are failing when running either headless browsers.  A follow on task should be to investigate and fix these tests.


jasontk19:
@ZachTRice That's weird... do you also get those errors when running the tests in gui mode?

ZachTRice:
For bonus points... It could be useful to write a node script that checks if port 3000 is in use, if it is not in use, i.e. WV isn't being ran, then run `node ./tasks/start.js` first before running the e2e test.

jasontk19:
@ZachTRice I'll keep looking into that in between other tasks; going to merge this for now since there's no difference to how the current e2e tests rely on the app being up and running locally.

PR #2113, opened 2019-09-10
Update timeline scale with current time every minute

### LINKED ISSUE #2106: Update timeline scale as time progresses while browser is open.
Currently, the a date/time beyond the one during which the app was loaded cannot be selected from the timeline.

Allow the user to advance the date if they kept the browser open long enough for time to progress.

jasontk19:
## Description

Fixes #2106 .

These fixes should allow the current time (down to the minute) to always be selectable on the timeline.  To do this, we are polling every 5 seconds to see if the current time minute value differs from the minute value in `appNow`.  Only if there is a difference do we update `appNow`.

To test, build with some sub-daily layers enabled.  Add at least one sub-daily layer, set the time scale to minute, and select the latest time possible (which should be the current time).  Then, wait until time advances to the next minute, and within 5 seconds the forward arrow should be enabled to allow you to select the new current time.  You should also be able to drag the slider forward to the current time. 

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/nasa-gibs/worldview/blob/master/.github/CONTRIBUTING.md) doc

@nasa-gibs/worldview


jasontk19:
Need to fix some broken tests before it's worth reviewing...

PR #2202, opened 2019-10-03
Animations layer caching 

### LINKED ISSUE #2166: Looped animations over 10 steps do not use cached layer tiles
**Describe the bug**
Looping animations with more than 10 steps enabled causes layer tiles to be re-requested on each play through.  

**To Reproduce**
1. Open animation tool
2. Set interval to 1 day, select 11 total days of range
3. Play

**Expected behavior**
We should be caching these layer tiles so they are only loaded once.

jasontk19:
## Description

Fixes #2166.

This change prevents the removal of a layer form the cache prior to adding it during animation.  Presumably this was implemented intentionally to refresh this item in the cache so that it was not automatically removed but, in my testing I had a hard time reproducing a scenario where this was an issue.  Looped animations performed significantly better with this change and I was able to see animations of up to ~90 days play smoothly without re-requesting layer tiles each time.

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/modules/map/selectors.js
+++ b/web/js/modules/map/selectors.js
@@ -5,40 +5,22 @@ import { promiseLayerGroup } from './util';
  * @return {object}      Promise.all
  */
 export function promiseImageryForTime(date, layers, state) {
-  var viewState;
-  var frameState;
-  var pixelRatio;
-  var promiseArray;
   const map = state.map;
-  var cache = map.ui.cache;
-  var mapUi = map.ui;
-  var selectedMap = map.ui.selected;
-  frameState = selectedMap.frameState_; // OL object describing the current map frame
+  const cache = map.ui.cache;
+  const mapUi = map.ui;
+  const selectedMap = map.ui.selected;
+  const { pixelRatio, viewState } = selectedMap.frameState_; // OL object describing the current map frame
 
-  pixelRatio = frameState.pixelRatio;
-  viewState = frameState.viewState;
-  promiseArray = layers.map(function(def) {
-    var key;
-    var layer;
+  const promiseArray = layers.map(def => {
+    const key = mapUi.layerKey(def, { date }, state);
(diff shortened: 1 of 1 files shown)


web/js/modules/map/selectors.js
```diff
@@ -5,40 +5,22 @@ import { promiseLayerGroup } from './util';
...
-    key = mapUi.layerKey(
-      def,
-      {
-        date: date
-      },
-      state
-    );
-    layer = cache.getItem(key);
-    if (layer) {
-      cache.removeItem(key);
```
Benjaki2: It's a rare case and I could be wrong but I think you actually want to recreate the layer even if it is in the cache:

e.g: If you were to look at a layer at some time before starting an animation it would be stored in our  `cachai cache` and the `tileCache`. But at some point it might be popped from the `tileCache`. Removing it here means `createLayer()` will actually recreate the layer (instead of skipping it's creation) `shifting` it to the top of the tileCache.

jasontk19: Yeah, conceptually this makes sense to me... I just had a hard time finding a scenario where I could see this causing any problems.  From what I can tell, I _think_ that if the item exists in the `cachai` cache but has been cleared from the tile cache, OL just re-requests the tiles.  I haven't dug into their code to confirm this but, I saw instances in really long animations where an item existed in the `cachai` cache yet tile requests were still being made. I assumed this was an indicator that that they were purged from the `tileCache`.

jasontk19: ... and if OL just re-requests the tiles when they've been purged from the `tileCache` then I'm guessing that's essentially not that different from recreating the layer to force it to the top of the cache.  

Let me know what you think.

Benjaki2: Thats seems about right. The only difference is that the re-request won't go out until you role over that date.

PR #2293, opened 2019-10-17
Disable thresholding and palette adjustment for the clean infrared layers

### LINKED ISSUE #2274: Fix/disable palette adjustments for Clean Infrared layers
Adjusting the palette for our upcoming geostationary Clean Infrared layers doesn't behave as expected.  Only the grayscale areas of the original palette are updated while the rainbow portion remains unchanged. In this example, I changed to the `Orange 1` palette but the layer isn't completed updated:

CleanIR-palette

Given our near-term release, this ticket is to at least disable palette adjustments for this set of layers but to also keep the original colorbars visible.  And if it's easy enough to fully fix, then let's of course go ahead and do that.

Note that there are three Clean Infrared layers which need to be addressed: GOES-East, GOES-West, and Himawari-8.

jasontk19:
## Description

Fixes #2274 

@nasa-gibs/worldview

### DIFF
--- a/config/default/common/config/wv.json/layers/goes/GOES-East_ABI_Band13_Clean_Infrared.json
+++ b/config/default/common/config/wv.json/layers/goes/GOES-East_ABI_Band13_Clean_Infrared.json
@@ -11,6 +11,9 @@
               "goes"
             ],
             "product":  "",
+            "palette": {
+                "immutable": true
+            },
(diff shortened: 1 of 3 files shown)


jasontk19:
@Benjaki2 awesome, that's way cleaner! updated.

PR #2308, opened 2019-10-18
E2E in Travis

jasontk19:
## Description

E2E tests will now run on every build!  If your build isn't passing, see if you broke any tests.

**Travis**
* Update travis config to run e2e tests in headless mode:
  * One OSX build with Node LTS will run with Chrome
  * One Linux build with Node LTS will run with Firefox
* Tests will run on every build
* See ticket comments for more details on why we aren't running browserstack at this time
* Allow Windows builds to fail for now since they are in beta and fairly often will fail
* Fix recent issue with Windows builds that was causing an older version of Python to be installed which meant they would _always_ fail
* Fix a few broken E2E tests so that everything _should_ be passing on build

**Browserstack**
* attempt to address an issue where process exited before logs could be received by parent process
* reduce the build configures to only three OS/browser combos
* give the build a name with timestamp so each run can be differentiated in the Browserstack dashboard

### DIFF
--- a/e2e/browserstack.conf.js
+++ b/e2e/browserstack.conf.js
@@ -23,7 +23,8 @@ const nightwatchConfig = {
     build: 'wv-nightwatch-' + timeStamp,
     applicationCacheEnabled: false,
     webStorageEnabled: false,
-    marionette: true
+    marionette: true,
+    project: 'Worldview'
   },
   test_settings: {
     default: {},
--- a/e2e/browserstack.conf.js
+++ b/e2e/browserstack.conf.js
@@ -1,6 +1,8 @@
 const environments = require('./environments.js');
 const glob = require('glob');
 const files = glob.sync('./e2e/features/**/*-test.js');
+const now = new Date();
+const timeStamp = `${now.getMonth() + 1}/${now.getDate()} @ ${now.getHours()}:${now.getMinutes()}`;
 
 const nightwatchConfig = {
   output_folder: './e2e/reports',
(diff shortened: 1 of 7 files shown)


e2e/browserstack.conf.js
```diff
@@ -18,7 +20,7 @@ const nightwatchConfig = {
     'browserstack.local': true,
     'browserstack.debug': true,
     'browserstack.localIdentifier': ('wvtester19234' + process.env.BROWSERSTACK_USER).replace(/[^a-zA-Z0-9]/g, ''),
-    build: 'nightwatch-browserstack',
+    build: 'wv-nightwatch-' + timeStamp,
     applicationCacheEnabled: false,
     webStorageEnabled: false,
     marionette: true
```
edplato: Random addition, now that browserstack changed their dashboard around our 'Project: Untitled Project' jumped out at me. This should change that:
```suggestion
    marionette: true,
    project: 'Worldview'
```

Benjaki2:
@jasontk19 it doesn't look like the travis check is showing up in this PR.

jasontk19:
@Benjaki2 I'm wondering if it's because I had a build going off this branch when I opened the PR?  I am seeing it show now that I just committed Ed's addition:
Screen Shot 2019-10-18 at 4 23 33 PM

Benjaki2:
That was probably it. I am the checks in action now!

jasontk19:
Node 12.x went LTS today so had to make some changes.  The latest commit will:

* Keep Node LTS builds, but allow them to fail
* run `npm run build` on each build job
* Set a flag to not wait for jobs that are allowed to fail before determining overall build status
* Run e2e on OSX/Linux jobs that explicitly set Node 10.16.3, instead of on the LTS builds

jasontk19:
Updated to use `npm ci` over `npm install` since we were running into an issue where `node-sass` couldn't find the right binding for the version of node we were building with:
```
Error: Missing binding /home/travis/build/nasa-gibs/worldview/node_modules/node-sass/vendor/linux-x64-57/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 8.x
Found bindings for the following environments:
  - Linux 64-bit with Node.js 10.x
```

jasontk19:
Latest build with this config can be seen here: https://travis-ci.org/nasa-gibs/worldview/builds/600917008

PR #2322, opened 2019-10-24
Make animation panel draggable, collapsable

### LINKED ISSUE #2081: Minimize the animation widget
The animation widget covers up valuable map real estate.

Be able to minimize the animation widget for a cleaner interface. Minimizing the controls to have only two buttons that play/pause and maximize (return to normal size) would be a clean interface that retains necessary functionality. There is also the option to allow clicking `X` to close the widget and continue playing, but that seems less intuitive as the range draggers need to be removed.

### LINKED ISSUE #2206: Animation panel should be moveable
Per user suggestion and dev discussion, we should make the animation panel moveable so it doesn't block content of interest.

jasontk19:
## Description

Fixes #2206 
Fixes #2081

### DIFF
--- a/web/css/anim.widget.css
+++ b/web/css/anim.widget.css
@@ -1,14 +1,16 @@
+.wv-animation-widget-wrapper {
+  position: absolute;
+  bottom: 125px;
+  left: 40%;
+}
 .wv-animation-widget {
   width: 334px;
   border-radius: 8px;
-  height: 134px;
+  height: 136px;
   background: #212121;
-  position: absolute;
-  transform: translateX(-50%);
-  -webkit-transform: translate(-50%);
-  bottom: 125px;
-  left: 50%;
   color: #fff;
+  padding: 1px 0;
+  cursor: move;
   & .wv-animation-widget-header {
     display: block;
     text-align: center;
@@ -81,10 +83,12 @@
     position: relative;
     width: 152px;
   }
-  & i.wv-close {
+  & i.fa {
     position: absolute;
-    top: 5px;
-    right: 5px;
+    &.wv-close {
+      top: 5px;
+      right: 7px;
+    }
     &:hover {
       color: #999;
       cursor: pointer;
--- a/web/css/anim.widget.css
+++ b/web/css/anim.widget.css
@@ -4,7 +4,7 @@
   z-index: 100;
   &.minimized {
     bottom: 80px;
-    left: 242px;
+    left: 10px;
     &.subdaily {
       left: 347px;
     }
--- a/web/css/anim.widget.css
+++ b/web/css/anim.widget.css
@@ -1,9 +1,6 @@
 .wv-animation-widget-wrapper {
   position: absolute;
   bottom: 88px;
-  /* left: 50%;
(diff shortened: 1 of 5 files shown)


jasontk19:
@ZachTRice Makes sense... this is where I had it originally but thought it looked cleaner when placed all the way to the left.  I might see if I can easily get it to conditionally move to that location when the custom interval selector is open.  If not, will move it back to above the animation button in the timeline.

ZachTRice:
I thought about this as well and I almost added an addendum to my comment to say, conditionally it could move when that selector is open. But, imo, I think it still looks clean above the animation icon even when it is the only thing open since the two icons (play + animation) are related. It feels a bit disconnected all the way to the left but that could just be me. I would be interested to hear what others think, I don't feel strongly either way.

jasontk19:
What if we centered the custom interval selector above the interval selection component itself?  IMO this makes more sense than having it off-centered. And...
Screen Shot 2019-10-25 at 12 28 37 PM

... it allows us to squeeze the minimized animation widget in the corner.  
Screen Shot 2019-10-25 at 12 28 44 PM


Agree that this placement feels _somewhat_ disconnected from the animation button (that opens the widget) but, I like this placement better than directly above that button.  It just looks cleaner to me when it's tucked away off to the side.  Here's a preview of that in the context of the rest of the viewport:
Screen Shot 2019-10-25 at 12 34 06 PM



Open to majority rule on this though.  @edplato let us know how you feel when you get back!

ZachTRice:
@jasontk19 that layout could work as well

edplato:
Looking good. I'm fine with the UI changes and have a couple minor style suggestions. I would conditionally align the custom interval selector with the right side of the `timeline-header` section so it lines up nicely in subdaily mode as well:
image

Also, the non-subdaily animation widget header text looks a little off centered now. The subdaily header looks pretty centered, but could you try to tweak the non-subdaily header to sit more centered. I know large custom intervals makes this difficult, so it may not be even worth it to have the current quasi-centered text and just go for a left text align. Idk - play around with a fix if it's possible.

Production:
image

Current branch:
image

Potential left align:
image


One thing I'm noticing with widget functionality is if I add a geostationary layer, the minimized widget jumps to the right, and if I remove a geostationary layer, the minimized widget jumps to the left.

jasontk19:
@edplato Good catch.  That's how it should be but I forgot to update the position for the subdaily condition.  Now the two modes look like this:
Screen Shot 2019-10-29 at 8 11 35 AM
Screen Shot 2019-10-29 at 8 11 25 AM

I updated the header text on the widget to be centered again, but now it takes into account the space for the collapse/close buttons and centers within the remaining space:
Screen Shot 2019-10-29 at 8 18 58 AM

I also pushed a few minor updates to keep the interval tooltip centered above the interval selection in the animation widget and, to keep the widget itself centered when switching between subdaily/regular mode if the user has not yet manually moved the widget.

jasontk19:
Yeah... think of the interval selector as an extension of the panel and the animation widget as a standalone draggable widget. They won't often be visible at the same time, and even then it should be only briefly while the interval is changed.

PR #2349, opened 2019-10-28
Upgrade Core JS

### LINKED ISSUE #1842: Upgrade core-js to v3.x
**Description**
After upgrading to Babel v7.x, we experienced conflicts (with sub-dependencies) while importing `core-js` v3.x forcing us to use the latest v2.x branch. We should resolve these conflicts and upgrade `core-js` to v3.x using the following imports:

```
import "core-js/stable";
import "regenerator-runtime/runtime";
```
ref: https://github.com/zloirock/core-js#babel

Currently, we are importing polyfills for es5, es6, es7 as documented in v2.x.x, but I anticipate this will create problems in IE when we implement ES2017+ features.

jasontk19:
## Description

Fixes #1842 

I ran browserstack a few times to get a sense for IE11 impact.  Unfortunately, IE11 is the flakiest when it comes to our E2E tests.  However, I manually checked the tests that I saw failing that differed from the tests that fail on develop and was not able to find anything broken after this update.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -851,6 +851,13 @@
       "requires": {
         "core-js": "^2.6.5",
         "regenerator-runtime": "^0.13.2"
+      },
+      "dependencies": {
+        "core-js": {
+          "version": "2.6.10",
+          "resolved": "https://registry.npmjs.org/core-js/-/core-js-2.6.10.tgz",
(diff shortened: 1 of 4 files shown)


jasontk19:
@ZachTRice this looks like something that is already happening in `develop`.  I'll file a new ticket for this.

PR #2375, opened 2019-11-04
Fix production build

### LINKED ISSUE #2348: Production build not able to start locally
**Describe the bug**
The app crashes after running `npm run build:prod` and starting the instance locally. Something was introduced/removed in between the latest release and the current develop branch that is causing this crash and console error. Current `master` and `48eca7aac872b88211ec997945acb62dc80212d5` work fine for reference.

1. pull down latest in develop, `npm install`, `npm run build:prod`
2. `npm start` to see app crash and console error

**Screenshots**
image

 - OS: Windows
 - Browser: Chrome
 - Version: 77.0.3865.120

jasontk19:
## Description

Fixes #2348 

* Disabled `mangle` option for `uglifyjs-webpack-plugin` (fixes prod issue).  This only increases the deployed, compressed size of `wv.js` by 0.1MB
* Enable gzip compression in `start` script to more easily assess deployed prod asset sizes in a local environment
* Update `travis.yml` to run production builds on jobs that run E2E tests so we can catch errors like this sooner


package.json
```diff
@@ -28,13 +28,13 @@
   "scripts": {
     "analyze": "cross-env ANALYZE_MODE=true NODE_ENV=production webpack",
     "browserstack": "node e2e/browserstack.js -c e2e/browserstack.conf.js -e browserstack ",
-    "build": "run-s clean getcapabilities build:config build:dev",
+    "build": "run-s clean build:config build:dev",
     "build:dev": "cross-env NODE_ENV=development webpack",
-    "build:config": "bash -c 'tasks/buildOptions.sh' && node ./tasks/config.js",
+    "build:config": "run-s getcapabilities && bash -c 'tasks/buildOptions.sh' && node ./tasks/config.js",
```
Benjaki2: I actually use `build:config` quite a bit to avoid the `getcapabilites` requests when I am editing config files or build tasks.  Maybe we could break this out into another task like `build:ci` or `build:travis`?

jasontk19: I was wondering about that use case...  Sounds good, I'll update this!

PR #2432, opened 2019-11-20
Snap layer times to previous available dates to prevent unnecessary network requests

### LINKED ISSUE #2129: Fix time snapping based on layer intervals
**Describe the bug**
Currently, we don't properly snap date/times to actual available times based on intervals for the following layer periods:
* Yearly - Incrementing month/day/minute will issue additional requests
* Monthly - Incrementing day/minute will issue additional requests

I haven't checked some of our more unusual intervals (8 day, etc) but these and others should be investigated. The `datesinDateRanges` and `prevDateInDateRange` functions should be updated to properly handle this snapping.  As of #2041 snapping should be working appropriately for sub-daily products.

**To Reproduce**
1. Load Black Marble layer
2. Set date to 2016-01-01
3. Increment to 2016-02-01 - note that tile requests are made

**Expected behavior**
We should not make requests for tiles at date/times at smaller steps than the defined interval.

**Screenshots**
Screen Shot 2019-09-13 at 2 24 29 PM

 - OS: any
 - Browser: any

edplato:
## Description

Fixes #2129  .

- [x] Add a `def.previousDate` and `def.nextDate` based on the `closestDate` determined from `prevDateInDateRange`. This allows for layers with longer ranges to not make requests if the current date is between `def.previousDate` and `def.nextDate`.
- [x] If a `closestDate` is found, it will be saved onto the layer building `options`.
- [x] Within `datesinDateRanges`, move how timezone offsets are handled as these will vary, base max dates and pushed dates on respective `dateInterval`, and minor refactor to only create certain constants if necessary for a given `def.period`.

## Further comments

I suggest future optimizations:
- [ ] Shorten date range when creating `dateArray` in `datesinDateRanges`. This can be costly CPU for slower machines with long 7000+ day interval arrays being created, for example.
- [ ] Adjust animation precache/layer fetching to tackle the problem handled in this PR (currently day animation increment will request each day even for monthly layers, which shouldn't be necessary).
- [ ] Determine use case for `def.startDate` and `def.endDate` and check if these ever conflict with the dates found in `def.dateRanges`, and if there are conflicts does any date take precedence.

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/map/layerbuilder.js
+++ b/web/js/map/layerbuilder.js
@@ -13,6 +13,7 @@ import lodashCloneDeep from 'lodash/cloneDeep';
 import lodashMerge from 'lodash/merge';
 import lodashEach from 'lodash/each';
 import { lookupFactory } from '../ol/lookupimagetile';
+import { datesinDateRanges, prevDateInDateRange } from '../modules/layers/util';
 import {
   isActive as isPaletteActive,
   getKey as getPaletteKeys,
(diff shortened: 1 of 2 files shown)


jasontk19:
Nice work!  I definitely think the further optimizations are worth capturing in new tickets.

Only thing I've noticed so far is that, on a monthly layer, switching between the 1st and 2nd day of the month sends a request (though imagery is the same).  For example, start here and increase day by 1: http://localhost:3000/?v=-28.6171875,34.6640625,61.3828125,81.703125&t=2013-04-01-T00%3A00%3A00Z&z=1&l=Coastlines,MODIS_Terra_L3_Land_Surface_Temp_Monthly_Day,MODIS_Terra_CorrectedReflectance_TrueColor(hidden)

edplato:
Good catch. I think I fixed that problem in `closestDate`.

PR #2470, opened 2019-12-02
Increase available zoom levels for arctic projection

### LINKED ISSUE #2460: Allow map to zoom in further in Arctic projection
**Describe the bug**
We now have 4 WELD surface/corrected reflectance layers in the Arctic projection that are at 30m resolution. We need to allow the Arctic map to zoom in that far. Currently it only zooms to 10km/5mi. Should probably be able to zoom as far as the Geographic map 1000m/5000ft.

**To Reproduce**
1. Go to Arctic
2. Click on Add Layers and type in "WELD"
3. Add the 4 WELD layers
4. Zoom in and see error - You cannot zoom into the full resolution of the image.

**Expected behavior**
Should probably be able to zoom as far as the Geographic map 1000m/5000ft.

**Screenshots**
Arctic 

Geographic

jasontk19:
## Description

Fixes #2460

### DIFF
--- a/config/default/common/config/wv.json/projections/arctic.json
+++ b/config/default/common/config/wv.json/projections/arctic.json
@@ -27,7 +27,7 @@
                 512.0,
                 256.0
             ],
-            "numZoomLevels": 7,
+            "numZoomLevels": 12,
             "fractionalZoom": false,
             "startCenter": [
                 0,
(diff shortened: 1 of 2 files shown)


jasontk19:
@Benjaki2 good catch!  should be good now

PR #2476, opened 2019-12-02
Create gitleaks secret check action

edplato:
## Description

Add https://github.com/marketplace/actions/trufflehog-actions-scan action to scan last 50 commits for secrets on push. This should accommodate larger pushes with a lot of commits. Fail on API key based secrets pushed into repo.

Based on action:  
https://github.com/dxa4481/truffleHog
https://github.com/max/secret-scan

## Further comments

Can tweak further to only run on certain branches, as part of a larger security check, etc.

## Checklist

@nasa-gibs/worldview


.github/workflows/secret-check.yml
```diff
@@ -0,0 +1,11 @@
+name: Secrets Check - Commit
+
+on: [push]
+
+jobs:
+  security-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
Benjaki2: I am confused by this syntax. It seems like this action would checkout `master` but it does appear to checkout the `actions-secret-check` in the logs

edplato: That is curious. I believe it will change to `develop` after merge and then eventually `master`. Will have to double check each stage to make sure.

PR #2481, opened 2019-12-03
Don't render sidebar contents when sidebar is collapsed

### LINKED ISSUE #2480: Layer palette tooltips show when sidebar is collapsed
**Describe the bug**
Layer palette tooltips show when sidebar is collapsed

**To Reproduce**
1. `?v=-83.94922083883691,31.34417884092222,-78.78468817629832,36.251586833206396&t=2019-12-02-T00%3A07%3A29Z&l=GRanD_Reservoirs,GRanD_Dams,Reference_Labels(hidden),Reference_Features(hidden),Coastlines,VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor`
2. Hover over any vector feature

**Expected behavior**
Tooltips don't show when sidebar collapsed.

**Screenshots**
Screen Shot 2019-12-03 at 9 30 13 AM

jasontk19:
## Description

Fixes #2480.

* fixes an issue where tooltips show when the sidebar was collapsed by not rendering sidebar contents at all when the sidebar is collapsed.  
* also fixes issue where tooltip shows above vector dialogs

### DIFF
--- a/e2e/features/compare/compare-test.js
+++ b/e2e/features/compare/compare-test.js
@@ -99,7 +99,7 @@ module.exports = {
       client.click(toggleButton);
       client.pause(100);
       client.expect.element(collapsedToggleButton).to.be.visible;
-      client.expect.element(toggleButton).to.not.be.visible;
+      client.waitForElementNotPresent(toggleButton, TIME_LIMIT);
       client.useCss().assert.containsText(collapsedToggleButton, 'Layers (6)');
       client.click(collapsedToggleButton);
       client.pause(100);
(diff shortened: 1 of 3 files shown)


jasontk19:
@Benjaki2 Hopefully all that is fixed by #2482.  I'm assuming the overlap is because they are from two different layers, both of which are outside the visible area and that, if both were visible you'd just see two tooltips in two different layer panels.

jasontk19:
Looks like this breaks an E2E test though, let me take a look at that...

PR #2482, opened 2019-12-03
Don't show tooltips when target is not visible in sidebar scroll container

jasontk19:
## Description

Fixes # 2469.

* Bring in `react-visibility-sensor`
* ~~Prevent webpack from transpiling `react-visibility-sensor` [more here](https://github.com/joshwnj/react-visibility-sensor/issues/148)~~
* Apply to palette legend tooltips so they don't show when the tooltip target is not within the visible area of the sidebar

### DIFF
--- a/web/js/components/sidebar/paletteLegend.js
+++ b/web/js/components/sidebar/paletteLegend.js
@@ -5,6 +5,7 @@ import { drawSidebarPaletteOnCanvas, drawTicksOnCanvas } from '../../modules/pal
 import lodashIsNumber from 'lodash/isNumber';
 import { Tooltip } from 'reactstrap';
 import { getOrbitTrackTitle } from '../../modules/layers/util';
+import VisibilitySensor from 'react-visibility-sensor';
 
 class PaletteLegend extends React.Component {
   constructor(props) {
@@ -13,7 +14,8 @@ class PaletteLegend extends React.Component {
       isRunningData: props.isRunningData,
       colorHex: props.colorHex,
       isHoveringCanvas: props.isHoveringCanvas,
-      width: this.props.width
+      width: this.props.width,
+      scrollContainerEl: null
     };
   }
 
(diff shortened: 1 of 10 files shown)


web/js/components/sidebar/paletteLegend.js
```diff
@@ -31,6 +33,11 @@ class PaletteLegend extends React.Component {
 
   componentDidMount() {
     this.updateCanvas();
+    this.setState(() => {
+      return {
+        scrollContainerEl: document.querySelector('#productsHolder .simplebar-wrapper')
```
Benjaki2: I think this needs to be relative to the active compare side to get it to work when the B state is active.

jasontk19:
@edplato maybe a dumb question but... did you stop the `npm run watch` script and restart it?  I see the above error when switching to this branch while `watch` is running but stopping it, then running `npm ci` then `npm run watch` seems to work for me.

edplato:
Changing the import to `import VisibilitySensor from 'react-visibility-sensor/visibility-sensor';` allows it to build and run.

edplato:
@jasontk19 Yeah, I exited `watch` mode, removed `node_modules` (after first fail just in case), ran `npm install`, and `npm run build` / `IGNORE_ERRORS=true npm run build`.

jasontk19:
Ok looks like it's a Windows vs Mac thing... the Travis builds for Windows are failing for that exact reason.  I'll take a look.

web/js/components/sidebar/paletteLegend.js
jasontk19: Fixed 👍

jasontk19:
There are a number of e2e tests failing due to the change that causes the layer list to not be rendered if it's not active.  I'm going through them now to update the tests and confirm nothing is actually broken by this change.

web/js/components/sidebar/paletteLegend.js
edplato: I needed to change the selector to just `#productsHolder` to make it work - with `.simplebar-wrapper` included I would still see the tooltip on hover even with sidebar collapsed.

jasontk19: It sounds like you might be seeing the issue that was fixed in #2480?

Issue #2359, opened 2019-10-29
Typing a value into the date field required a TAB or Enter to save the value


minniewong:
**Describe the bug**
When you type in a value into the date/time field, a TAB or Enter is required to save the value. If you type in a value, click on the next value with your mouse, the previous value is not saved. This occurs in the timeline and in the animation widget.

**To Reproduce**
1. Go to the date/time entry field
2. Click on OCT and change it to 07
3. Click on the Day field and see that it did not save the previous value to "07"

**Expected behavior**
Typing/manually entering in a value in the time/date fields and clicking on the next field should save the previous value. 

**Desktop (please complete the following information):**
 - OS: OSX
 - Browser: Chrome

**Additional context**
This was discussed in the comments of #2131 but was not addressed.

edplato:
Related to #1662 as additional input validation considerations arise working on a fix for this issue.

PR #2510, opened 2019-12-16
Loosen date selector input validation to allow temporary invalid values and mouse click changes

### LINKED ISSUE #1662: Allow invalid text entry into date fields
The UI currently prevents you from typing certain dates into the date panel when they're deemed "invalid".

As it turns out, you sometimes need to type in an invalid date en route to a valid one.  E.g., 
- if the currently selected date is 2019 March 11 which is also the date in real life
- you want to change the selected date to 2018 June 15
- the interface requires you to first change the year to 2018 since "June" or "15" isn't available yet in 2019.

We should probably loosen this to allow invalid dates to be entered; we can still highlight which entries are invalid and not send any requests to GIBS while there are invalid entries.  It would just allow more flexibility while entering dates as a human might want to do.

edplato:
## Description

Fixes #1662 , #2359 . Loosen date selector inputs (applies to timeline and start/end animation) to:
1) Allow temporary invalid values (e.g., currently on FEB-28, want to change to MAR-30, and type in the 30 first). Valid values of that type (e.g., year: 2020, day: 31) will be saved as a temporary value for further validation. Invalid characters (e.g., year: 'ASDA', day: 33) will not be saved for further validation; and
2) Allow mouse clicks to change date inputs.

Substantial refactors were made to retain multiple temporary invalid input values and check for valid dates using these temporary values.

- [x] #1662 Allow invalid text entry into date fields
- [x] #2359 Typing a value into the date field required a TAB or Enter to save the value
- [x] Add E2E tests
- [x] Refactor/cleanup

## Further comments

Is there any other behavior that needs to be addressed by allowing temporary invalid inputs? First thought is for animation, the inputs can be invalid, but the animation can still be played. Do we want to disable the play?

## Checklist

@nasa-gibs/worldview


jasontk19:
Looks good.  I'm wondering if we should restrict this flexibility to just typed dates?  Sometimes when I am clicking through to adjust the day value, I am working my way from right to left and want to set a later date in an earlier month but can't select it until I change the month.

Example:  Load app with `now` value of `2019-6-15` but you want to manually set the currently selected time to the 2019-7-31.  You have to change the month first before the day.

I haven't looked closely enough at the code changes yet to have a sense for how much more work that would be, and not sure if others think that's desirable but, these are my initial thoughts.

PR #2513, opened 2019-12-17
Anti-meridian crossing measurements fix

### LINKED ISSUE #2464: Creating area measurements that cross poles sometimes incorrect
**Describe the bug**
Area measurements not using great circle calculations that cross the anti-meridian are incorrect.

**To Reproduce**
1. Switch to arctic projection
2. Create an area measurement that crosses the anti-meridian

**Expected behavior**
Area measurements that cross the anti-meridian should be correct

**Screenshots**
Screen Shot 2019-12-13 at 2 15 56 PM

Any

jasontk19:
## Description

Fixes #2464.

* Fix an issue with running data throwing errors when hovering over a measurement on the map
* Remove turf.js and arc.js libraries
* Bring in geographicLib library for more accurate area/distance calculations and great circle arc rendering
  * Fixes an issue where distances and areas crossing the anti-meridian in polar projections were WAY off
  * Fixes an issue where distance and area measurements crossing the anti-meridian in geographic projection were drawn incorrectly.
* Remove "use great circle" option since all measurements now use great circle calculations.  
* Previously, distance _was_ properly being measured as lines with n-loxodrome segments.  However, if this option was unchecked, area calculations appeared to just be using a different algorithm for great circle area calculations which means that the measurement shown would change but not necessarily represent an area measurement of a polygon with n-loxodrome sides. Instead of leaving in the toggle for distance measurements only, it seemed easier to remove it altogether for now.
* Add unit tests to verify that distances ands areas crossing the anti-meridian in a variety of scenarios are equal to those which do not cross


jasontk19:
@edplato Good catch on the residual code.  That should all be removed now.

As for great circle measurements, I'm not opposed to adding back in help text in some form.  A great circle is just a straight line on a sphere.  From what I've seen, most mapping applications are doing great circle calculations under the hood, whether or not they are rendering them as such.  OpenLayers for example does great circle calculations for distance and area measurement even though they don't have built-in support for _drawing_ great circle lines.

PR #2569, opened 2020-01-14
Slow down scroll handlers, add scroll pan hoverTime calculation

### LINKED ISSUE #2554: Timeline pad scroll is too fast for reliable use
**Describe the bug**
Timeline pad scroll is too fast for reliable use. This causes other scroll wheel related issues with timing and positioning that can lead to broken timescale axis.

edplato:
## Description

Fixes #2554  .

- [x] Split wheel handlers for scroll (Y-axis) and pan (X-axis)
- [x] Refactor scroll handler timing and debounce scroll (Y-axis)
- [x] Slow down pan (X-axis)
- [x] Add `hoverTime` conditional calculation necessary for touchpad horizontal scroll gesture
- [x] Add edge case check to prevent `hoverTime` being outside of axis date range
- [x] Add `UP/DOWN` key listener to change `timeScale` like changing dates with `LEFT/RIGHT` keys

## Further comments

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/timeline/timeline-axis/timeline-axis.js
+++ b/web/js/components/timeline/timeline-axis/timeline-axis.js
@@ -26,7 +26,12 @@ class TimelineAxis extends Component {
       currentTimeRange: null,
       gridWidth: 12,
       wheelZoom: false,
-      mouseDown: false
+      mouseDown: false,
+      hitRightBound: false,
+      hitLeftBound: false,
+      updatedTimeScale: false,
+      wheelType: null,
+      clientXOnDrag: 0
     };
     // axis
     this.handleDrag = this.handleDrag.bind(this);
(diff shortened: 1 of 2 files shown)


jasontk19:
This is a huge improvement!  

There is still a little bit of weirdness when using the Mac touchpad:  Sometimes a single swipe will cause the timeline to jump from the highest to lowest zoom (Year -> Month -> Day) but other times it seems to only want to jump one step (Year -> Month).  I think the former behavior was probably intended but, I actually found the latter behavior to be easier to use since it became easier to avoid jumping through more zoom levels than I intended - requiring me to make a conscious effort to stop and swipe again to go up/down another level.

Just my $0.02

PR #2616, opened 2020-01-29
Fix formatting of the date/time that is displayed in the tooltip of a…

### LINKED ISSUE #2285: Available imagery date/time is too long
When there is no imagery available for a layer, a no entry sign appears and when you hover over the sign it shows the date/time where the imagery is available. It's currently showing too much information.

Make the information more concise - report the time as HH:MM Z or HH:MM UTC.

StephanJW:
… disabled layer. Refactored logic.

## Description

Fixes #2285.

- The timestamp when layers are in a disabled state were needlessly long. This changes their format based on the layer period. Refactored existing logic to make code more readable.

@nasa-gibs/worldview

### DIFF
--- a/web/js/containers/sidebar/layer.js
+++ b/web/js/containers/sidebar/layer.js
@@ -85,19 +85,17 @@ class Layer extends React.Component {
   formatDisabledTitleDate = (date, period) => {
     var dateString;
 
-    switch (period) {
-      case 'subdaily':
-        dateString = moment(util.parseDate(date)).format('DD MMMM YYYY HH:mm') + 'Z';
-        break;
-      default:
-        dateString = moment(util.parseDate(date)).format('DD MMMM YYYY');
+    if (period === 'subdaily') {
+      dateString = moment(util.parseDate(date)).format('DD MMMM YYYY HH:mm') + 'Z';
+    } else {
+      dateString = moment(util.parseDate(date)).format('DD MMMM YYYY');
     }
 
     return dateString;
   }
 
   getDisabledTitle = (layer) => {
-    var startDate = null; var endDate = null;
+    let startDate, endDate;
 
     if (layer.startDate) {
       startDate = this.formatDisabledTitleDate(layer.startDate, layer.period);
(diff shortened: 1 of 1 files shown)


web/js/containers/sidebar/layer.js
```diff
@@ -81,36 +82,34 @@ class Layer extends React.Component {
...
-          ' ' +
-          startDate.getFullYear();
-        endDate =
-          endDate.getDate() +
-          ' ' +
-          util.giveMonth(endDate) +
-          ' ' +
-          endDate.getFullYear();
-      }
+    var startDate = null; var endDate = null;
```
jasontk19: No need to assign newly declared variables to `null` since they will be `undefined` until assigned some other value. `undefined` is false-y so your conditionals below will still work as intended.

Also, I'd recommend using `let` or `const` over `var` (`let` in this case).

web/js/containers/sidebar/layer.js
```diff
@@ -81,36 +82,34 @@ class Layer extends React.Component {
     }
   }
 
+  formatDisabledTitleDate = (date, period) => {
+    var dateString;
+
+    switch (period) {
```
jasontk19: This is more of a personal style preference but, a `switch` statement is maybe more verbose than necessary for a conditional with only two possible outcomes.  An if statement might be simpler:

```
if (period === 'subdaily') {
  return moment(util.parseDate(date)).format('DD MMMM YYYY HH:mm') + 'Z';
}
return moment(util.parseDate(date)).format('DD MMMM YYYY');
```

web/js/containers/sidebar/layer.js
StephanJW: I had to brush up on the use of `var` vs `let` vs `const` in the realm of ES6. `let` definitely makes sense here. I'll make the change.

web/js/containers/sidebar/layer.js
StephanJW: I agree with this. I think what I had in mind using the `switch` was maybe there were additional types of formats for the time we would want to associate for various types of periods. But if it's just these two cases, then and if will definitely suffice. Thanks for reviewing - really appreciate great feedback!

Issue #1749, opened 2019-06-10
Combine date-range code in product picker and layer-info modal


Benjaki2:
Issue: Duplicate code.

It should be easy to combine them, the layer-info date-range is now broken out into it's own component.

PR #2635, opened 2020-02-04
2286 timezone in temporal coverage

### LINKED ISSUE #2286: Include timezone in the temporal coverage text
The temporal coverage section in the Information section for a subdaily layer should include the timezone e.g. Z or UTC. 

Screen Shot 2019-10-16 at 1 17 08 PM

The temporal coverage section should include the timezone e.g. Z or UTC so that a user knows what timezone the subdaily layer is being reported in.

StephanJW:
## Description

Fixes #2286 & #1749.

There is a lot of duplicated code for displaying temporal coverage date ranges. Refactored. Also fixes bug where date ranges differed from what was in the product picker vs what was shown in the info-box.

@nasa-gibs/worldview

### DIFF
--- a/web/js/util/util.js
+++ b/web/js/util/util.js
@@ -15,6 +15,7 @@ import browser from './browser';
 import { events } from './events';
 import load from './load';
 import Cache from 'cachai';
+import moment from 'moment';
 
 export default (function(self) {
   var canvas = null;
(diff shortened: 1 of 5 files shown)


web/js/util/util.js
```diff
@@ -293,6 +294,33 @@ export default (function(self) {
...
+    let dateString;
+    date = this.parseDate(date);
+
+    switch (period) {
+      case 'subdaily':
+        dateString = moment(date).format('DD MMMM YYYY HH:mm') + 'Z';
+        break;
+
+      case 'yearly':
+        if (dateType === 'END-DATE') date.setFullYear(date.getFullYear() - 1);
```
jasontk19: This looks interesting. Why do we have to subtract a year or a month from end dates for those periods?

StephanJW: I had trouble wrapping my head around it - but I carried this over from the previous code base because on those set periods, the end-date for that layer is off by a single day. For example, the BLACK MARBLE 2012-2016 layer has a end date of January 1st, 2017; however, the data coverage technically stops then on December 31st, 2016. Since it has a period of a year, the formatting would look like 2012-2017 when it should be 2012-2016; hence the subtraction. 

This is the same for months. Using the same dates above, we do NOT want January 2012 - January 2017; rather, we would like it to be January 2012 - December 2016.

jasontk19: Ah that's right, I remember the discussion on slack the other day about this.  Well, if we're already doing this and it works, that works for me 👍

Issue #2642, opened 2020-02-07
Mobile datepicker should show forward/backward buttons


jasontk19:
**What existing feature could be improved?**
Currently, it is tedious to advance one day forward or backwards on a mobile device.  You have to open the mobile date picker and scroll the date up/down, then click ok.  There is plenty of room to add simple `<` and `>` arrows next to the datepicker button on mobile to simplify this action.

ragnarokatz:
@jasontk19 Hello, I would like to take on this issue please.

ragnarokatz:
@jasontk19 Hello, I have created a PR for this issue. Please review it when you have time, thanks.

PR #2643, opened 2020-02-09
Fix issue #2642: Mobile datepicker should show forward/backward buttons

ragnarokatz:
## Description
Add forward and backward buttons to the mobile datepicker for worldview.

Fixes issue #2642.

## Screenshot
image

## Files modified:
`timeline.css`, `timeline.js`. re-used `DateChangeArrows` component.

## Further comments
I'm fairly new to front end development, as well as the reactjs library. I'm open to any feedback to my proposed design.
- I don't know if the design style I've put in (combination of `.css`, `id` and `className`) is how a front end developer usually does it. Please advise if there's a better/cleaner way.
- Right now the arrows look a little bigger than the mobile datepicker. I don't know if I should shrink the arrows or enlarge the mobile datepicker.

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/css/timeline.css
+++ b/web/css/timeline.css
@@ -641,6 +641,18 @@ button:focus {
   cursor: pointer;
 }
 
+/* Mobile date change arrows */
+.mobile-date-change-arrows-btn {
+  position: absolute;
+  left: 120px;
+  bottom: 10px;
+  background: rgba(40, 40, 40, 0.85);
+  border: thin solid #333;
+  padding: 3px;
+  border-radius: 5px;
+  cursor: pointer;
+}
+
 .datepicker-header,
 .datepicker-navbar-btn {
   font-weight: 700;
--- a/web/js/containers/timeline/timeline.js
+++ b/web/js/containers/timeline/timeline.js
@@ -905,13 +905,29 @@ class Timeline extends React.Component {
         {initialLoadComplete
           ? <ErrorBoundary>
             {isSmallScreen
-              ? <MobileDatePicker
-                date={selectedDate}
-                startDateLimit={timelineStartDateLimit}
-                endDateLimit={timelineEndDateLimit}
-                onDateChange={this.onDateChange}
-                hasSubdailyLayers={hasSubdailyLayers}
-              />
+              ? <div id="timeline-header">
+                <div id="date-selector-main">
+                  <MobileDatePicker
+                    date={selectedDate}
+                    startDateLimit={timelineStartDateLimit}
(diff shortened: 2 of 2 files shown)


jasontk19:
@ragnarokatz Nice job!  Thanks for your work on this.  As for the approach, I haven't tested it yet but it looks good to me! 

Regarding the datepicker height, this was something I tried to address recently on another branch which I am about to open a PR for.  That branch is `2545-phase-1-product-picker`.   If you want, you can combine this work with that branch and see how it will end up looking.  I suspect they will line up nicely!  

My only other feedback at this time would be that the padding to the right of the forward arrow appears greater than the padding to the left of the back arrow, which makes things look a little off.  Maybe we can tweak this a bit to equal them out?

We are about to go into testing phase for the next release so, if we can get this PR wrapped up in the next week or so we'd be happy to include it in our next release!

web/css/timeline.css
```diff
@@ -641,6 +641,23 @@ button:focus {
   cursor: pointer;
 }
 
+/* Mobile date change arrows */
+.mobile-date-change-arrows-btn {
+  position: absolute;
+  left: 120px;
+  bottom: 10px;
+  background: rgba(40, 40, 40, 0.85);
+  color: #fff;
```
edplato:
```suggestion
```

web/css/timeline.css
```diff
@@ -641,6 +641,23 @@ button:focus {
...
+  border: thin solid #333;
+  font-size: 17px;
```
edplato:
```suggestion
```

web/css/timeline.css
```diff
@@ -641,6 +641,23 @@ button:focus {
...
+  font-family: 'Open Sans', sans-serif;
```
edplato:
```suggestion
```

web/css/timeline.css
```diff
@@ -641,6 +641,23 @@ button:focus {
...
+  padding: 3px;
+  border-radius: 5px;
+  line-height: 25px;
```
edplato:
```suggestion
```

web/css/timeline.css
```diff
@@ -641,6 +641,23 @@ button:focus {
...
+  white-space: nowrap;
```
edplato:
```suggestion
```

ragnarokatz:
@jasontk19 @edplato Thank you for your feedback. I have removed extra styling code as suggested. As discussed, I will leave the un-even padding for now since it will affect the same arrows in the desktop version as well.

PR #2679, opened 2020-02-19
Add ipify metrics

edplato:
## Description

- [x] Add ipify ip capture for metrics

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/app.js
+++ b/web/js/app.js
@@ -139,7 +139,7 @@ class App extends React.Component {
     config = self.props.config;
     config.parameters = state;
 
-    const main = function() {
+    const main = async function() {
       const models = self.props.models;
 
       // Load any additional scripts as needed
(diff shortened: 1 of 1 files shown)


web/js/app.js
```diff
@@ -150,6 +150,16 @@ class App extends React.Component {
       }
       if (config.features.googleTagManager) {
         googleTagManager.init(config.features.googleTagManager.id); // Insert google tag manager
+
+        // get user ip address using https://www.ipify.org/ api
+        const response = await fetch('https://api.ipify.org?format=json');
+        const json = await response.json();
```
jasontk19: FWIW you shouldn't need to use the `await` keyword when calling this non-async function `response.json()`.  Similarly, it's not needed for accessing an object property: `json.ip`. Looks like it doesn't break anything though so, 👍

edplato: Thanks, for some reason I need it on the `response.json()` too or I keep getting `undefined`, but on `json.ip` it's definitely pointless.

jasontk19: Ah, yeah looks like I was thinking of some other library with a `response.json()` call that wasn't asynchronous.  Their docs clearly show `async response.json()`.  But now I'm wondering why 🤔

edplato: I based it off https://dev.to/johnpaulada/synchronous-fetch-with-asyncawait. Only thing I can think is the `response.json()` parsing could be a lengthy process if it's a lot of data so it needs to be a promise too?

PR #2680, opened 2020-02-20
2664 layer legend disappear fix

### LINKED ISSUE #2664: Layer color legend can disappear when zooming out
**Describe the bug**
If a layer is loaded while in an "overzoomed" state (which would cause the zot to show) zooming out to the point that the layer is no longer overzoomed causes the color legend to disappear when the zot is removed.

**To Reproduce**
1. Go to `?v=5.01815972899115,18.179856676715676,6.67160211180365,19.649827379840676&t=2015-09-29-T00%3A00%3A00Z&l=MODIS_Terra_Aerosol,MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor,Reference_Labels(hidden),Reference_Features(hidden),Coastlines`
2. Zoom out until "Aerosol Optical Depth" layer is no longer overzoomed and zot disappears
3. Observe that color palette for this layer is no longer visible.

**Expected behavior**
The color legend should not disappear.

**Screenshots**
Screen Shot 2020-02-18 at 9 12 38 AM

 - OS: OSX
 - Browser: Chrome

StephanJW:
## Description

Fixes #2664 . 

When zooming in on a layer and triggering the ZOT, the palette legend would disappear. This resolves the issue.

## Further comments

The reason for this bug appearing was because the `componentDidUpdate()` was triggering the `updateCanvas()` extremely frequently. Thus, we wanted to only redraw when a layer's settings changed (i.e. threshold, squash, color scheme, etc.) I think there is a better way to resolve this by having the ZOT property embedded with the layer object. I wasn't particular sure how to go about it and maybe it was more work than what it is worth?

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/sidebar/paletteLegend.js
+++ b/web/js/components/sidebar/paletteLegend.js
@@ -42,7 +42,7 @@ class PaletteLegend extends React.Component {
     });
   }
 
-  componentDidUpdate(prevProps, prevState) {
+  componentDidUpdate(prevProps) {
     // Only updates when layer options/settings have changed or if ZOT changes the width of the palette
     if (!lodashIsEqual(this.props.layer, prevProps.layer) || (prevProps.width !== this.props.width)) {
       this.updateCanvas();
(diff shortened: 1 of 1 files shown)


jasontk19:
This seems like a reasonable approach to me, and appears to fix the issue 👍 

Just need to re-target the branch to point at `develop`

PR #2697, opened 2020-02-24
2692 scrollbar overlap

### LINKED ISSUE #2692: Scroll bar overlaps layer list when you first add new layers
**Describe the bug**
When you first add new layers to the layer list and the scroll bar appears, the scroll bar overlaps the layer list. If you refresh the page/click on start comparison/go to the event tab and back to the layer list, the scroll bar no longer overlaps.

**To Reproduce**
1. Go to Worldview
2. Click on Add Layers
3. Add 3 new layers
4. See error - scroll bar overlaps the layer list. 

**Expected behavior**
The scroll bar should not overlap the layer list the first time you add layers. 

**Screenshots**
Screen Shot 2020-02-21 at 10 19 29 AM

 - OS: OSX
 - Browser: Chrome and Safari

jasontk19:
## Description

Fixes #2692 

* move scrollbar visible class check to a `useLayoutEffect` hook on it's own that doesn't depend on `props.scrollBarVerticalTop` to change in order to run
* debounce the call to `toggleVisibileClass` since it will be called on each render

### DIFF
--- a/web/css/tour.css
+++ b/web/css/tour.css
@@ -216,8 +216,6 @@
   padding: 0 4px 4px;
 }
 
-
-
 /* .tour-in-progress */
 .tour-in-progress .modal {
   left: auto;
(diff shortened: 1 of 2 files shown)


jasontk19:
Hmm I'll take a quick look before merging and see if there is an easy/obvious fix

jasontk19:
Ok, I think this is better.  Now there is no flickering when the scrollbar is added and it shows immediately in compare mode when switching tabs.

PR #2710, opened 2020-02-26
2695 measurement styles orange borders

jasontk19:
## Description

Styles! Please approve. kthx.

Screen Shot 2020-02-26 at 11 04 18 AM 

Screen Shot 2020-02-26 at 11 04 35 AM

### DIFF
--- a/web/css/layers.css
+++ b/web/css/layers.css
@@ -181,8 +181,8 @@
   left: 0;
 
   &.measurement-row {
+    padding: 0;
     border-right: 1px solid #eee;
-    padding: 0 0 0 8px;
   }
   &.search-row {
     border-left: 6px solid transparent;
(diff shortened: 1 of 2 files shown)


jasontk19:
@ed I think this is some kind of weird bug... there isn't actually a border there.  I see this sometimes in Chrome, but it comes and goes.  I checked every element in the stack of items and none has this border.  I will take a closer look and see if there is a workaround (negative margin on an overlying element?).  I was kind of hoping it was just some weird thing only I was running into.

edplato:
Toggling this seems to remove it for me, but I'm not sure how that trickles down and impacts anything else.
image

jasontk19:
For whatever reason, removing that border doesn't fix it in Chrome (though I did originally add that in the process of trying to hide the dark mystery border and so, have removed it).

Adding a 1px negative margin seems to do the trick.  Hopefully the combo of both changes works well in Edge/IE11 🤞

PR #2726, opened 2020-03-03
2311 previous date not displaying fix

### LINKED ISSUE #2311: The previous UTC date should be displayed when before 3:00 UTC
**Describe the bug**
The previous UTC date should be displayed when before 3:00 UTC so that imagery is visible to the user on load
**To Reproduce**
1. Go to homepage before 3:00 UTC

**Screenshots**
Screen Shot 2019-10-18 at 9 20 09 PM

**Additional context**
Looks like it was probably introduced in `v3.0.0`

StephanJW:
Fixes #2311 . The extent should now position at the right time when the page load exceed the 3 UTC threshold as well as the date-picker being correctly set.


StephanJW:
Will review this today.

web/js/containers/timeline/timeline.js
```diff
@@ -1284,7 +1284,9 @@ function mapStateToProps(state) {
     hasSubdailyLayers = hasSubDaily(layers[compare.activeString]);
     endTime = layersLastDateTime(layers[activeString], config);
   }
-  endTime = appNow;
+  /** endTime = appNow; This was originally here, wondering why the
+  above logic is there when it gets set here **/
+  endTime = state.config.todaysDate;
```
StephanJW: I attempted to make changes, but I still don't think this is behaving exactly as we want it to. An example to try would be to put in the current date using the `now` param (i.e. ?now=2020-03-12T0). The date picker gets set to the correct time which should be yesterday's date. However, advancing to the next date using the right arrow, the right arrow is still in an enabled state.  Clicking it again will disable it and change the param to be the date the page was loaded on which is being captured as `todaysDate`. It's very confusing as we have this other notion of `appNow` which is setting the `endTime` which controls the ending date of the timeline (see Timeline.js). It looks like there is some logic in there that's being completely disregarded as well (lines 1269-1275). I'm somewhat stumped as to how to resolve this effectively. Hoping for another set of eyes on this.

web/js/main.js
```diff
@@ -63,10 +63,20 @@ window.onload = () => {
       config.pageLoadTime = parameters.now
         ? util.parseDateUTC(parameters.now) || new Date()
         : new Date();
+
+      config.initialDate = config.pageLoadTime.getUTCHours() < 3
+        ? new Date((config.pageLoadTime).setUTCDate(config.pageLoadTime.getUTCDate() - 1))
```
edplato:
```suggestion
        ? new Date(config.pageLoadTime.setUTCDate(config.pageLoadTime.getUTCDate() - 1))
```

web/js/main.js
```diff
@@ -63,10 +63,20 @@ window.onload = () => {
...
+        : config.pageLoadTime;
+
+      config.todaysDate = new Date().getUTCHours() < 3
+        ? new Date((new Date()).setUTCDate(new Date().getUTCDate() - 1))
```
edplato:
```suggestion
        ? new Date(new Date().setUTCDate(new Date().getUTCDate() - 1))
```

web/js/location.js
```diff
@@ -118,11 +118,7 @@ export const mapLocationToState = (state, location) => {
 const getParameters = function(config, parameters) {
   const now = config.pageLoadTime;
   const nowMinusSevenDays = util.dateAdd(config.pageLoadTime, 'day', -7);
-  // If at the beginning of the day, wait on the previous day until GIBS
-  // catches up (about three hours)
-  const initialDate = now.getUTCHours() < 3
-    ? new Date(now).setUTCDate(now.getUTCDate() - 1)
-    : now;
+  const initialDate = config.initialDate;
```
edplato: Not necessary, but you can destructure this section to be more like:
```
const { initialDate, pageLoadTime } = config;
const now = pageLoadTime;
const nowMinusSevenDays = util.dateAdd(pageLoadTime, 'day', -7);
```

web/js/containers/timeline/timeline.js
edplato: I think you found the inherit issue here. You will need to import `testNow` from the `date` state imports above and implement my recommended fix.

edplato:
```suggestion
  // handle explicit testNow axis endTime
  endTime = testNow || endTime;
```

web/js/main.js
Benjaki2: I think the main issue is that `config.pageLoadTime` is being mutated here.

edplato: Yeah good catch, probably want to double check all those dates for that.

StephanJW:
This should be now good to go!

jasontk19:
Hmm, looks like there are still conflicts and there are some unit tests failing: https://travis-ci.org/github/nasa-gibs/worldview/builds/663631326?utm_source=github_status&utm_medium=notification

StephanJW:
I believe I have this corrected, now.

PR #2741, opened 2020-03-09
Apply AirBnB style guide rules

### LINKED ISSUE #2728: Consider stricter code linting rules for a more consistent codebase
We should consider introducing a stricter set of eslint rules for JS + React, such as those corresponding with the AirBnB style guide, in order to keep a more consistent codebase.  Alternatively, we could manually enable additional eslint rules of our choosing.

A quick test with the [eslint airbnb config](https://www.npmjs.com/package/eslint-config-airbnb) marked ~13,000 errors though, after running the autofix that number was down to 4,651.

Perhaps for now it's worth investigating which rules we can easily adhere to with the help of the autofix.  Then, we could enable these and conform to them moving forward.

jasontk19:
## Description

Fixes #2728.

[AirBnB Style Guide Reference](https://github.com/airbnb/javascript)

This brings in the ESLint rules to apply the AirBnB style guide and, addresses a great deal of the instances where our existing code did not conform to these stricter rules.  Our existing ESLint rules were combined with what's in the AirBnB rule set where possible, rather than tossing out all rules we had in place in favor of what's in the guide.

A number of rules from the AIrBnB set were manually turned off or set to `warn`.  In cases where a rule was overridden by the `warn` setting, I deemed the rule potentially useful for us but the number of manual fixes too high to attempt at this time.  After this is merged, I would suggest we pay attention to the warnings within files we are working in and, try to fix these when possible while trying to avoid introducing any new warnings with new code.  In cases where a rule was set to `off`, I deemed the rule potentially too strict for us at this time (based on the number of errors or warnings it would throw if enabled).  

A huge number of files were modified; many by the autofixer but quite a few also had to be manually edited to comply with certain rules.  Due to this, the potential for newly introduced bugs is high.  When reviewing, pay close attention to the Travis build status and consider any parts of the build/deployment process that don't normally get touched that aren't covered by CI/CD or testing.

Finally, consider combining this branch with a work in progress locally to get a feel for the overall strictness and to identify any rules you might disagree with or think go too far.


e2e/features/image-download/unsupported-test.js
```diff
@@ -6,7 +6,7 @@ const startParams = ['v=-180,-90,180,90', 't=2018-06-01', 'imageDownload='];
 const TIME_WAIT = 10000;
 
 module.exports = {
-  after: function(client) {
+  after(client) {
```
Benjaki2: @jasontk19 why does this differ from `before: (c) => {` change below? I don't know what rule this is implementing.  I would expect `after: (client) =>`

web/js/components/animation-widget/gif-post-creation.js
```diff
@@ -71,35 +70,50 @@ export class GifResults extends Component {
                 <div>
                   <b> Size: </b>
                 </div>
-                <div>{size} MB</div>
+                <div>
+                  {size}
+                  {' '}
```
Benjaki2: This change seems like it might be unwanted extra work.

web/js/components/animation-widget/loop-button.js
```diff
@@ -8,27 +8,23 @@ import { faRetweet } from '@fortawesome/free-solid-svg-icons';
...
  * @extends React.Component
  */
-class LoopButton extends React.Component {
-  render() {
-    return (
-      <a
-        title={this.props.looping ? 'Stop Loop' : 'Loop video'}
-        className={
-          this.props.looping
+const LoopButton = (props) => (
```
Benjaki2: Nice!

web/js/components/animation-widget/play-queue.js
```diff
@@ -392,9 +394,12 @@ class PlayAnimation extends React.Component {
    * @param index {string} Date string
    */
   animate(index) {
-    const { selectDate, endDate, speed, isPlaying } = this.props;
+    const {
+      selectDate, endDate, speed, isPlaying,
+    } = this.props;
     let currentDateStr = index;
-    let nextDateStr, nextDateParsed;
+    let nextDateStr; let
```
Benjaki2: `let nextDateStr, nextDateParsed;` seems better than `let nextDateStr; let nextDateParsed;`
I'd prefer either prevent same line declarations out right or allow  `let nextDateStr, nextDateParsed;`

web/js/components/compare/opacity-slider.js
```diff
@@ -18,9 +18,10 @@ class OpacitySlider extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      value: props.value
+      value: props.value,
     };
-    this.debounceOpacityUpdate = lodashDebounce(this.onSlide.bind(this), 100);
+    this.onSlide = this.onSlide.bind(this);
+    this.debounceOpacityUpdate = lodashDebounce(this.onSlide, 100);
```
Benjaki2: 👍

web/js/components/animation-widget/play-queue.js
jasontk19: Yeah in this particular instance I think the autofix wasn't able to handle the formatting properly.  It should be one var per line. This wasn't intentional.

web/js/components/animation-widget/gif-post-creation.js
jasontk19: I don't know that I love having the var, the space, and the text on three separate lines.  That said, the linter auto-formats this way if you enter new code or modify existing so, it shouldn't require any work on our part to comply.

e2e/features/image-download/unsupported-test.js
jasontk19: I'm not sure, most of this kind of formatting was handled by the autofix.  This is a change to comply with the `object-shorthand` rule for methods: https://github.com/airbnb/javascript#es6-object-shorthand

I'm not sure why it wouldn't also format the other to `before(c) { ... ` 🤷‍♂

jasontk19:
> **My initial Comments**
> 
> * I made a lot of comments on changes I like and don't like. I am not saying that we override defaults just because I think something looks weird. But if there is a consensus on some of those items it may be worth overriding them.
> * Arrow function changes probably need to be well tested since they adjust scope. I think @ZachTRice tried to make those changes once and ran into scope problems.
> * I like the next line `&&` & `||`
> * At first glance I don't really like the additional `,` and necessary bracket `(c) =>`

@Benjaki2, thanks for the feedback!

* The overwhelming majority of arrow functions introduced came from the autofix.  I had assumed it was smart enough to know when, and when not, to apply these but we can keep an eye out.  

* I don't love the required parens around single parameters in arrow functions either (Too many parens back to back decreases readability IMO).  [Supposedly it leads to slightly cleaner diff](https://github.com/airbnb/javascript#arrows--one-arg-parens) (I guess when switching form one to multi params?).  I'd definitely be up for attempting to override this so we can omit when possible.

* The trailing comma thing is uncomfortable to look at, at first.  But, the bonus is easier to read diffs and, less work when adding/removing an entry from a multiline array/object (since you only have to edit the line you are changing).  I hated it when I first saw it but after using it on another project I became a fan.
https://github.com/airbnb/javascript#commas--dangling

jasontk19:
@edplato @Benjaki2 

I created a branch with the `arrow-parens` rule set to `["error", "as-needed"]` and ran the autofix.  This setting means that [parentheses around parameters for arrow functions are only enforced when they are necessary.](https://eslint.org/docs/rules/arrow-parens.html#as-needed)

You can take a look at [the diff with this rule applied here](https://github.com/nasa-gibs/worldview/compare/2728-eslint-airbnb...eslint-airbnb-arrow-parens-as-needed) and see which you prefer.  I am very much on the fence.  Let me know if you feel strongly one way or the other.

edplato:
Re: `arrow-parens`, I kind of like the parens for that little added clarity I get.
For the single parameter, it adds a tiny bit more clarity for me:
`handleKeyPress= e => {`
**vs.**
`handleKeyPress= (e) => {`

But, when other arguments are present it seems to help a lot more:
`lodashEach(config.scripts, script => {`
**vs.**
`lodashEach(config.scripts, (script) => {`

PR #2776, opened 2020-03-19
Fix measure tooltip placement issue

### LINKED ISSUE #2515: Measurements with long line segments can draw unexpectedly
**Describe the bug**
Measurements that contain line segments which exceed 180 degrees of longitudinal change but do not cross the anti-meridian are not drawn as expected.

**To Reproduce**
1. Start distance measurement
2. Begin line somewhere in the US
3. End line somewhere in China by traveling East
4. As the line exceeds 180 degrees of length, the drawing will update to reflect the distance measurement result by jumping Westward to cross the anti-meridian.  This is because the calculations done to draw a great circle arc (instead of a straight line between start/end points) and the calculations done to measure distance are both seeking the shortest distance between the start/end points of the line.

**Expected behavior**
TBD

**Screenshots**
Screen Shot 2019-12-18 at 2 13 30 PM

jasontk19:
## Description

When drawing measurements with long segments that exceed 180 degrees, shift the entire drawing either east/west by 360 degrees to line up with the orientation of the active segment which is seeking the shortest distance between its two points

Tagging @minniewong for testing and @Benjaki2 for code review.

Fixes #2515

@nasa-gibs/worldview

### DIFF
--- a/web/js/map/measure/ui.js
+++ b/web/js/map/measure/ui.js
@@ -117,8 +117,8 @@ export function measure(map, mapUiEvents, store) {
     let tooltipCoord;
     mapUiEvents.trigger('disable-click-zoom');
     sketch = evt.feature;
-    drawChangeListener = sketch.getGeometry().on('change', (evt) => {
-      const geom = evt.target;
+    drawChangeListener = sketch.getGeometry().on('change', (e) => {
+      const geom = e.target;
       if (geom instanceof OlGeomPolygon) {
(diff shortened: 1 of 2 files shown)


jasontk19:
@minniewong I could argue either side so... I'm not sure.  There is still some weirdness when taking really long measurements when you are zoomed way out like that.  I think of these as edge cases that might not accurately represent the intended use of this tool and therefore might not warrant the additional time investment to address but, I'm open to opinions on that.

PR #2786, opened 2020-03-20
Run npm test on pre-push hook

### LINKED ISSUE #2780: Pre-push git hooks
We should implement a hook that runs on `git push` that runs `npm run test` so can catch failing unit tests and linting errors before pushing to a remote branch.

jasontk19:
## Description

Fixes #2780 

This will run `npm test` any time you do a `git push`

If you use the VSCode GUI to push, it may be easy to overlook that this is being run in the background when you push since there is not much feedback unless the check fails, in which case a dialog will pop up.  

If you do `git push` from a terminal, you will of course see the output from `npm test`.

If, at any time, you just want to push things to `remote` without waiting for the test/linting checks to run and pass (e.g.  your laptop is currently on fire and you need to get the changes pushed _quick_) simply pass the `--no-verify` flag: `git push --no-verify`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,11 +34,11 @@ RUN cd /usr/src && \
     pip --version
 RUN mkdir -p /usr/local/stow
 RUN cd /usr/local/stow && \
-    curl -O https://nodejs.org/dist/v8.11.2/node-v8.11.2-linux-x64.tar.xz && \
-    tar xf node-v8.11.2-linux-x64.tar.xz && \
-    rm -f /usr/local/stow/node/node-v8.11.2-linux-x64.tar.xz && \
-    rm -f /usr/local/stow/node-v8.11.2-linux-x64/{LICENSE,*.md} && \
-    stow -S node-v8.11.2-linux-x64
+    curl -O https://nodejs.org/download/release/v10.19.0/node-v10.19.0-linux-x64.tar.xz && \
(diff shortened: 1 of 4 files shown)


jasontk19:
@Benjaki2 Good catch... I have been using Node v10 (or now, v12) for a while without any issue.  It's probably time to update what we list in the `package.json`.  I think this is preferable to using an older version of husky.  Any concerns with that?

jasontk19:
Also, when testing, I was able to see the pre-push hook run successfully before I opened the PR so not sure what's going on there but I will try it again.

Benjaki2:
> @Benjaki2 Good catch... I have been using Node v10 (or now, v12) for a while without any issue. It's probably time to update what we list in the `package.json`. I think this is preferable to using an older version of husky. Any concerns with that?

The only thought is that we likely will want to update `Node` in the `dockerfile`, `tasks/checknodeversion.js` and maybe some of the travis tests as well.

Benjaki2:
Accidentally pressed close & comment 🤦‍♂

Benjaki2:
> Also, when testing, I was able to see the pre-push hook run successfully before I opened the PR so not sure what's going on there but I will try it again.

I probably just need delete my `node_modules` and reinstall now that my Node version is up to date... I might have forgotten to do that.

Issue #1776, opened 2019-07-02
Epic - Timeline Enhancements v1.1


edplato:
This ticket is to track future enhancements to the timeline capabilities - some idea/UI designs can be seen in in #1496 .

- [x] Add data availability bars #1772 
- [x] More relaxed date text input #1662
- [x] Timeline pan with dragger at edge of visible axis #2579 
- [x] If custom widget open and no selection made, select previous interval rather than prevent left/right arrow date change

PR #2789, opened 2020-03-23
Add Timeline Data Availability Panel

### LINKED ISSUE #1772: Timeline "data availability bar" updates
- Show layer labels similar to Earthdata Search ([example](https://search.earthdata.nasa.gov/search/project?p=!C1443528505-LAADS!C1243477370-GES_DISC&tl=1525378169!4!!&q=airs%20co2&ok=airs%20co2))
- Indicate how often and when each layer changes (e.g., place a mark on each day if it's a daily product) #461
- Indicate when layers have "intervals" of available data and gaps in between #460
- Add NRT/STD and collection information - [See best available layers](https://wiki.earthdata.nasa.gov/display/GIBS/GIBS+API+for+Developers#GIBSAPIforDevelopers-%22BestAvailable%22Layers)

edplato:
## Description

Fixes #1772 part of #1776  .

**Add Timeline Data Availability Panel**

- Add "matching coverage" line on the timeline axis for coinciding coverage of selected layers
- Panel shows individual layer coverage with full range listed and multiple time units (ex: 8-day layer) broken up into separate time lines in current and smaller time units (e.g., 8-day is shown as separate 8-day data lines in day, hour, and minute time scales; however, single day data lines would only show as separate data lines in hour and minute time scales).
- Ability to toggle layers that are hidden in the sidebar to:
  - Show up in the data availability panel
  - Be included in the "matching coverage" line on the timeline axis
- Add E2E tests, various lint fixes, logic cleanups

**PR Review Fixes:**

- [x] Break up large rendering `layer-data-items` component
- [x] Allow layers with `rangeStart` but no `rangeDates` (e.g., Blue Marble, Fires)
- [x] Hover action outline spacing issue for timescale change
- [x] Tooltip positioning to reduce overflow cutoffs
- [x] GRACE unique key
- [x] Add warning message to users if no layers available so no empty panel
- [x] Increase border radius for data lines to improve identifying where the coverage splits for adjacent data lines of the same layer
- [x] Fix missing `throttleSettings`

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/css/timeline.css
+++ b/web/css/timeline.css
@@ -795,7 +795,7 @@ button:focus {
 }
 
 #wv-toggle-data-matching-main .react-switch-case {
-  width: 30px;
+  width: 45px;
 }
 
 #wv-toggle-data-matching-main .react-switch-label {
--- a/web/css/timeline.css
+++ b/web/css/timeline.css
@@ -860,13 +860,6 @@ button:focus {
   height: 12px;
 }
 
-.data-panel-coverage-line-overlay {
-  position: absolute;
-  pointer-events: none;
-  height: 12px;
-  z-index: 5;
-}
-
 .data-panel-layer-empty {
   height: 41px;
   padding: 6px;
--- a/web/css/timeline.css
+++ b/web/css/timeline.css
@@ -787,22 +787,18 @@ button:focus {
   color: #999;
 }
 
-#wv-toggle-data-matching-main {
+.wv-toggle-data-matching-main {
   align-items: center;
   margin: 0;
   white-space: nowrap;
   flex-grow: 1;
 }
 
-#wv-toggle-data-matching-main .react-switch-case {
+.wv-toggle-data-matching-main .react-switch-case {
   width: 45px;
 }
 
-#wv-toggle-data-matching-main .react-switch-label {
-  border: thin solid rgb(97, 97, 97);
-}
-
-#wv-toggle-data-matching-main > div.react-switch-label-case.switch-col {
+.wv-toggle-data-matching-main > div.react-switch-label-case.switch-col {
   padding-left: 5px;
 }
 
--- a/web/css/timeline.css
+++ b/web/css/timeline.css
@@ -810,7 +810,8 @@ button:focus {
   position: relative;
   background-color: #212121;
   color: #fff;
-  border: thin solid #595959;
+  box-sizing: border-box;
+  border: thin solid #333;
   border-top-right-radius: 5px;
   border-top-left-radius: 5px;
 }
@@ -850,4 +851,14 @@ button:focus {
 .data-panel-coverage-line {
   position: relative;
   height: 12px;
+}
+
+.data-panel-layer-empty {
+  height: 41px;
+  padding: 6px;
+}
+
+.data-item-empty {
+  display: flex;
+  flex-direction: row;
 }
--- a/web/css/timeline.css
+++ b/web/css/timeline.css
@@ -9,7 +9,7 @@ button:focus {
 
 .timeline-inner {
   position: absolute;
-  border: 1px solid #333;
+  border: thin solid #333;
   background: rgba(40, 40, 40, 0.85);
   border-radius: 5px;
   display: flex;
@@ -42,7 +42,7 @@ button:focus {
 
   & .input-wrapper {
     background: transparent;
-    border: 1px solid transparent;
+    border: thin solid transparent;
     vertical-align: top;
     box-sizing: border-box;
     -webkit-box-sizing: border-box;
@@ -255,8 +255,8 @@ button:focus {
 }
 
 #timeline-footer {
-  border-left: 1px solid #666;
-  border-right: 1px solid #666;
+  border-left: thin solid #666;
+  border-right: thin solid #666;
 }
 
 #wv-timeline-range-selector {
@@ -539,19 +539,19 @@ button:focus {
   height: 64px;
   width: 20px;
   cursor: pointer;
-  border-left: 1px solid #1a1a1a;
+  border-left: thin solid #1a1a1a;
   border-top-right-radius: 5px;
   border-bottom-right-radius: 5px;
 }
 
-#timeline-hide:hover {
+#timeline-hide:hover,
+#timeline-data-availability-panel-handle:hover {
   background-color: #fff;
+  border-color: #000;
 }
 
-#timeline-hide:hover .wv-timeline-hide-double-chevron-left,
-#timeline-hide:hover .wv-timeline-hide-double-chevron-right,
-.wv-timeline-hide-double-chevron-left:hover,
-.wv-timeline-hide-double-chevron-right:hover {
+#timeline-hide:hover .wv-timeline-hide,
+#timeline-data-availability-panel-handle:hover .wv-timeline-data-availability-handle-chevron {
   background-position: -206px -63px;
 }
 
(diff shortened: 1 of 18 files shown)


web/js/modules/layers/util.js
```diff
@@ -117,27 +117,60 @@ export function prevDateInDateRange(def, date, dateArray) {
...
-   * @param  {object} def           A layer object
-   * @param  {object} date          A date object
+   * @param  {object} def            A layer object
+   * @param  {object} date           A date object
+   * @param  {object} startDateLimit A date object (optional) used as start date of timeline range for available data
+   * @param  {object} endDateLimit   A date object (optional) used as end date of timeline range for available data
    * @return {array}                An array of dates with normalized timezones
    */
-export function datesinDateRanges(def, date) {
+export function datesinDateRanges(def, date, startDateLimit, endDateLimit) {
```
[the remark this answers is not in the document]
jasontk19: FWIW I would consider this part optional but, nice to have.

jasontk19:
Screen Shot 2020-03-25 at 8 54 04 AM

I believe the bar in the timeline scale is intended to reflect where all layers overlap but, it appears to not do that in this case where the Blue Marble layer is included.

jasontk19:
Screen Shot 2020-03-25 at 8 57 10 AM

The hover action on the timescale menu misbehaves when the availability panel is open:  Trying to point at, and select, one of the timescale choices within the menu is not possible because it disappears when hovered over.

jasontk19:
Tooltips for sections of coverage are time

jasontk19:
Ugh sorry, the "close and comment" is way too close to the screenshot click area!

jasontk19:
Screen Shot 2020-03-25 at 9 01 47 AM

Tooltips for coverage areas can be cut off (as seen here for the Black Marble layer) or not visible at all (when hovering over the Corrected Reflectance layer in the above scenario)

jasontk19:
Screen Shot 2020-03-25 at 9 03 59 AM

It feels weird that the availability panel extends past the timeline scale on the right side.  This often gives the impression that all layers don't have coverage towards the far right end of the scale, likely due to the fact that on the left end the panel, coverage bars, and timeline scale are all in alignment.

Any thoughts on shortening the availability panel to keep it in alignment with the start/end of the timeline scale like so?  I notice that it does look kind of busy to have the measure tool and scale showing in the background to the right but I think it improves the appearance of the availability panel itself:
Screen Shot 2020-03-25 at 9 09 48 AM

jasontk19:
Screen Shot 2020-03-25 at 9 20 29 AM

"Zooming" in and out from day => month => day on the timescale with the GRACE layer added caused the following error in the console:
```
Warning: Encountered two children with the same key, `0`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

jasontk19:
Screen Shot 2020-03-25 at 9 30 20 AM

This is a pretty far out edge case but, if all layers are hidden and you open the panel, it looks a bit odd with no content.  Might be nice to have a minimum one-row height with a message like "No visible layers" or, perhaps preferably, just default to turning on the "include hidden layers" toggle when opening the panel with all layers hidden.

web/js/containers/timeline/timeline.js
```diff
@@ -99,10 +100,21 @@ class Timeline extends React.Component {
...
+      changeDate,
+      onUpdateStartDate,
+      onUpdateEndDate,
+      onUpdateStartAndEndDate,
+    } = this.props;
+
     // left/right arrows
-    const throttleSettings = { leading: true, trailing: false };
-    this.debounceDateUpdate = lodashDebounce(this.props.changeDate, 8);
+    const throttleSettings = throttleSettings;
```
jasontk19: Noticed this when resolving merge conflicts with this rebased onto my current lint changes branch... 

Looks like we're assigning a new `const` to its `undefined` self here?  Not sure if that was intentional or not.

jasontk19:
Screen Shot 2020-03-26 at 8 16 40 AM

This Fires layer (and possibly others) does not show in the availability panel but does cause a blank row to be inserted.

```
?l=Reference_Features,MODIS_Terra_Thermal_Anomalies_All,MODIS_Terra_CorrectedReflectance_TrueColor
```

Benjaki2:
Also, the toolbar looks to get cut off:
Screen Shot 2020-03-31 at 3 37 58 PM

edplato:
> Also, the toolbar looks to get cut off:
> Screen Shot 2020-03-31 at 3 37 58 PM

Yeah, getting the tooltips offset right has been quite tricky for me. I've spent a lot of time working on them and still not satisfied. it's getting to the point that I'm not sure how to proceed with time investment vs. payoff at this point tbh.

jasontk19:
Screen Shot 2020-04-03 at 8 13 41 AM
Screen Shot 2020-04-03 at 8 13 52 AM

At the "year" zoom level, Black Marble shows as a continuous bar from 2012-2017 but at the "month" zoom level it more correctly shows as two separate coverage areas in 2012 and 2016.  I know this is a weird layer but, this is a little misleading since at the "year" level the availability panel indicates there is coverage where none shows in Worldview.

Screen Shot 2020-04-03 at 8 22 35 AM

Do we need to fix this?

jasontk19:
There are a number of layer titles + subtitles that are so long they will collide with the date range text.  This doesn't look terrible (although the rare case that breaks into a second line looks pretty odd). We could probably fix later but, it may be worth noting/adding follow-on ticket.

Screen Shot 2020-04-03 at 8 30 34 AM

jasontk19:
Overall fixes look good!  I like the bar stripes to improve dragging visual feedback and think that having the panel line up on left and right with the timeline scale itself feels a lot less disorienting.  I noticed that tooltips still have some issues when a layer's availability bar extends beyond the visible left/right boundaries of the panel but I don't know that this is a big deal in that particular instance and I'm sure getting these positioned was a tedious task.

Nice work!

edplato:
I fixed the word wrapping for long titles/slim browsers. I forgot to address the Black Marble issue previously mentioned. The timeline axis matching coverage line is based on total coverage from layer `startDate` to `endDate`, so in this case it is somewhat misleading. Major parts of this feature will have to be reworked to accommodate multiple coverage lines on the timeline axis vs. the current `startDate` to `endDate`. This could be investigated and possibly addressed (possibly due to some performance constraints) in a subsequent release or Black Marble type layers can be omitted from matching coverage due to majority of the "coverage" in the `startDate` to `endDate` is actually no coverage. It almost should be two layers of Black Marble 2012 and Black Marble 2016, but I get their motivation in combining them and this feature's lack of intuitive support for that layer.

web/css/timeline.css
```diff
@@ -708,3 +735,148 @@ button:focus {
...
+}
+
+#wv-toggle-data-matching-main {
+  align-items: center;
+  margin: 0;
+  white-space: nowrap;
+  flex-grow: 1;
+}
+
+#wv-toggle-data-matching-main .react-switch-case {
```
Benjaki2, line 797: Best to avoid adding (and never hurts to replace) `#` id selectors when possible, even though they are all over the codebase.

I would also suggest adding a `.thin-border` class  in the switch component and putting the styles in switch.js component so they are reusable, this would make it it so you can just pass a border prop in the react component.
```
.switch-thin-border .react-switch-label {
  border: thin solid rgb(97, 97, 97);
  height: 17px; // account for added pixel height
  width: 32px; // account for added pixel width
}
```

edplato: I modified the switch and id styling.

For the striped bar, I'm not sure how to go about addressing this to get 100% striped bar movement on large coverage bars.

The issue is that large bars are not fully rendered. So for your above issue, it would be a massive bar rendered starting in 2000 to current time if they were fully rendered. For a fix to handle the partially rendered larger bars, an overlay is added to larger bars to simulate the revised striped bars being dragged. What you are seeing where it doesn't drag is that on the edges of large bar coverage the overlay is removed so the simulated striped bars are not being dragged any longer.

The code is https://github.com/nasa-gibs/worldview/blob/layer-data-availability/web/js/components/timeline/timeline-data/layer-data-items.js#L198-L208
Any ideas on how to address this without fully rendering the entire bars?

PR #2802, opened 2020-03-30
2670 polar zoom restore

### LINKED ISSUE #2670: Outgoing permalink for rotated polar projections does not restore zoom level properly
**Describe the bug**
Outgoing permalink for rotated polar projections does not restore zoom level properly

**To Reproduce**
1. Start here: `?v=-937368.1948149145,-1990624.0246799085,1113956.7813860718,-33354.296850244515&r=-35.6728&p=arctic&t=2019-09-16-T20%3A00%3A00Z`
2. Reload the page, view now more zoomed out
3. Reload the page again, view is now even further zoomed out

**Expected behavior**
Zoom level is restored properly 

## Before
Screen Shot 2020-02-18 at 3 01 22 PM

## After first refresh
Screen Shot 2020-02-18 at 3 01 29 PM

StephanJW:
## Description

Fixes #2670.

- Go to an arctic / antarctic project
- Select random location to view to update the extent
- Reload the page and view should be the same as before
- Do the same thing with a rotated view

Prior to these changes, the view would be correct, however, the zoom level would decrease by one each time the page was reloaded.

## Further comments

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/main.js
+++ b/web/js/main.js
@@ -60,6 +60,12 @@ window.onload = () => {
   loadingIndicator.delayed(promise, 1000);
   promise
     .done((config) => {
+      // Perform check to see if app was in the midst of a tour
+      if (parameters.tr) {
+        // Gets the extent of the first step of specified tour and overrides current view params
+        const tourParams = util.fromQueryString(config.stories[parameters.tr].steps[0].stepLink);
+        parameters.v = tourParams.v;
+      }
       config.pageLoadTime = parameters.now
         ? util.parseDateUTC(parameters.now) || new Date()
         : new Date();


web/js/map/ui.js
```diff
@@ -228,12 +228,17 @@ export function mapui(models, config, store, ui) {
     }
     self.selected = self.proj[proj.id];
     const map = self.selected;
-    const currentRotation = proj.id !== 'geographic' && proj.id !== 'webmerc' ? map.getView().getRotation() : 0;
-    store.dispatch({ type: UPDATE_MAP_UI, ui: self, rotation: currentRotation });
+
+    const isProjectionRotatable = proj.id !== 'geographic' && proj.id !== 'webmerc';
```
jasontk19: I really like the use of a variable to hold a conditional check that's used more than once here.  Makes things super readable 💯

web/js/main.js
```diff
@@ -60,6 +60,11 @@ window.onload = () => {
   loadingIndicator.delayed(promise, 1000);
   promise
     .done((config) => {
+      // Perform check to see if app was in the midst of a tour
+      if (parameters.tr) {
+        // Gets the extent of the first step of specified tour and overrides current view params
+        parameters = util.fromQueryString(config.stories[parameters.tr].steps[0].stepLink);
```
jasontk19: I get a little nervous when I see object/array access this many levels deep without any `undefined` checking.  I don't _think_ there would be a case where the first step of a tour doesn't have a `steplink` so it should be OK but... may be worth using [lodash's `get` function](https://lodash.com/docs/4.17.15#get) in the future to access nested arrary/object properties like this safely.

jasontk19:
I don't think this made it into develop properly.  Check out the graph below.  Also, looking at the [commits in develop here](https://github.com/nasa-gibs/worldview/commits/develop)

Screen Shot 2020-04-03 at 8 52 34 AM

PR #2816, opened 2020-04-08
Enable redux devtools

jasontk19:
Enable redux devtools

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1298,6 +1298,12 @@
         "@babel/types": "^7.4.4"
       }
     },
+    "@babel/helper-validator-identifier": {
+      "version": "7.9.5",
+      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.9.5.tgz",
+      "integrity": "sha512-/8arLKUFq882w4tWGj9JYzRpAlZgiWUJ+dtteNTDqrRBz9Iguck9Rn3ykuBDoUwh2TO4tSAJlrxDUOXWklJe4g==",
+      "dev": true
+    },
     "@babel/helper-wrap-function": {
       "version": "7.8.3",
       "resolved": "https://registry.npmjs.org/@babel/helper-wrap-function/-/helper-wrap-function-7.8.3.tgz",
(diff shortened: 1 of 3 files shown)


jasontk19:
Hmmmm, the extension actually still seems to crash pretty frequently on page refresh 🤔 

I'll see if I can figure out why.

Benjaki2:
There was a known bug blocking when we first tried to implement this (when first porting to redux). That's why the code was commented out. Figured it would be fixed by now though.. 🤔

jasontk19:
So, I've been working on a branch with this enabled and, though the Redux devtools extension does sometimes seem to crash if it is open when you refresh the app, it otherwise does not seem to cause any issues.  I have been able to use it effectively between page refreshes.  I'm wondering if it makes sense to have this hooked up in the app and leave it up to devs to use the plugin with that caveat in mind?

Benjaki2:
The entire app seems to crash for me when this is enabled.

Benjaki2:
I am for pushing it as is (`debug`) since you can always just disable the extension.

jasontk19:
Ok so, I dug into this some more and found a few open Github issues for the extension related to crashing.  [From what I gathered here](https://github.com/zalmoxisus/redux-devtools-extension/issues/619#issuecomment-457235409) it seems like some of the objects on our state are just too big to be serialized as many times (or as quickly) as the extension is trying to.

[The troubleshooting guide](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/Troubleshooting.md#excessive-use-of-memory-and-cpu) recommends sanitizing large objects from your actions/state to get around this issue:

> The extension is in different process and cannot access the store object directly, unlike vanilla `redux-devtools` which doesn't have this issue. In case sanitizing doesn't fit your use case, you might consider including it directly as a react component, so there will be no need to serialize the data, but it would add some complexity.

In our case, I found that removing the `map` object from our state in the `stateSanitizer` function seems to entirely alleviate the issue and I'm able to use the extension without any crashes. 
 Removing this from the state that is passed to the extension is not ideal, since someone using it who wasn't aware it was being sanitized might be confused as to why it's not showing up in the extension.

We could consider adding the `redux-devtools` as a component in the app itself (which of course would only show if you run with the debug flag) in order to use these devtools in a more performant way.  But, this sounds a little more involved and is maybe more than I want to mess with right now.

About `redux-logger`... I didn't notice any significant difference running (or not running) this alongside the devtools browser extension.  But, for practical purposes, I don't see why you'd want to have both going at the same time since you get the same info from the extension.  Personally I'm not a huge fan of the logger since you can't filter the messages in the Chrome console due to [a bug in Chrome devtools](https://developers.google.com/web/tools/chrome-devtools/console/log#text)

So, I think the suggestion to set up npm tasks to run with just the logger or just the extension might be the way to go for now to give each dev flexibility in how they want to debug redux.  At least until we can implement the component directly (if we decide that makes sense).  I think this is probably worth a quick conversation at the dev meeting before moving forward so I will hold off on any additional work for now.

jasontk19:
Closing until I can re-open with the desired approach

PR #2825, opened 2020-04-10
Stop animation playback when Product Picker is opened

### LINKED ISSUE #2818: Animation doesn't stop when you add a layer (S-06 date.animation.14)
MacBook/Firefox

Testing S-06 date.animation.14:
Start an animation and do one of the following:

   ...
  ...
   ...
    add layer

Expected result:      The animation should stop
    Pause button should reset to the play button

My link:
https://go.nasa.gov/2Rml0m6

Animation
2020-APR-02 to. 2020-APR-07
2 frames per second, looping

If I play, then click Add layers, then select Aerosol Optical Depth MAIAC, the animation continues to play, it does not stop

jasontk19:
## Description

Fixes #2818.

### DIFF
--- a/web/js/containers/sidebar/footer-content.js
+++ b/web/js/containers/sidebar/footer-content.js
@@ -16,6 +16,7 @@ import ProductPicker from '../../components/layer/product-picker/product-picker'
 import { openCustomContent } from '../../modules/modal/actions';
 import { toggleListAll } from '../../modules/natural-events/actions';
 import { DATA_GET_DATA_CLICK } from '../../modules/data/constants';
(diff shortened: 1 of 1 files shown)


jasontk19:
@Benjaki2 I like that better... will update this.

jasontk19:
Actually... that doesn't seem to work.  Should the animation reducer be able to respond to layer module actions?

jasontk19:
Actually, I don't think this will work that cleanly.  We aren't dispatching an `ADD_LAYER` action to open the product picker, we are dispatching a `OPEN_CUSTOM` action from the modal module.  Not sure I necessarily want animation to stop on any custom modal open.

PR #2841, opened 2020-04-21
prevent glitchy vector rendering onZoom #2840

### LINKED ISSUE #2840: Vector point rendering for macOS chrome 80/81 is unreliable
**Describe the bug**
The Dams vector point layer appears to render unreliably. The vector points flicker in and out some complete tiles seem go missing at certain zoom levels.

**To Reproduce**
1. Go to '?v=-129.13381015637736,-58.88861311392172,100.35338772893672,70.37722256991536&l=GRanD_Dams,Reference_Labels(hidden),Reference_Features(hidden),Coastlines,VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor'
2. zoom in and out
3. See error

 - OS: macOS
 - Browser: CHROME v80 v81

**Additional context**
This seems to be the result of a chrome update for macOS only.

Benjaki2:
## Description

Half Fixes #2840
The points no longer flicker in and out but there are a few tiles that are still not rendered properly in outer zooms

- [x] Replaced vector tile [renderMode](https://openlayers.org/en/v5.3.0/apidoc/module-ol_layer_Vector-VectorLayer.html) with `image` as opposed to `vector`
- [x] Make getCapabilities `vectorStyles` be overwritten by local styles
- [x] Adjust styles of dams layer

## Further comments

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/config/default/common/vectorstyles/GRanD_Dams.json
+++ b/config/default/common/vectorstyles/GRanD_Dams.json
@@ -1,0 +1,54 @@
+{
+  "version": 8,
+  "name": "SEDAC",
+  "sources": {
+    "GRanD_Dams": {
+      "type": "vector",
+      "tiles": [
+        "https://gibs.earthdata.nasa.gov/wmts/epsg4326/std/GRanD_Dams_v1.01_STD/default/{Time}/{TileMatrixSet}/{TileMatrix}/{TileRow}/{TileCol}.mvt"
+      ]
+    }
+  },
+  "layers": [
+    {
+      "id": "GRanD_Dams_v1.01_STD",
+      "source": "GRanD_Dams",
+      "source-layer": "GRanD_Dams_v1.01_STD",
+      "source-description": "Default",
+      "type": "circle",
+      "paint": {
+        "circle-radius":[
+          "step",
+          [
+            "zoom"
+          ],
+          1,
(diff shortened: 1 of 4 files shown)


jasontk19:
I still see quite a bit of flickering as I zoom in/out but it does seem better than how it behaves in prod right now.

Benjaki2:
Yeah, it's definitely not perfect. The main difference is that this version seems to work when you are zoomed in. The question is... is it enough of an enhancement to merit this patch? I'm going to start working on the WMS fix. Maybe I can get that in quickly

Benjaki2:
The real problem with the WMS solution is that it would be touching a lot of code.

Benjaki2:
@edplato Good call. I had tried to do `git checkout package-lock.json` since I didn't make and changes there but must have did a rm instead 🤷‍♂️. Should be good now.

jasontk19:
Ok I made the mistake of not rebuilding before deploying.  It looks a lot better than what is in production though as you said there are completely missing points at certain zoom levels.  I'd say it's worth getting the fix in for this release.

PR #2870, opened 2020-04-30
Add GA metrics for distraction free mode, data coverage panel, and permalink layers

edplato:
## Description

Fixes WV-1917

- [x] Add GA metrics for distraction free mode, data coverage panel, and permalink layers

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/containers/timeline/timeline.js
+++ b/web/js/containers/timeline/timeline.js
@@ -868,6 +868,9 @@ class Timeline extends React.Component {
   * @returns {void}
   */
   toggleDataCoveragePanel = (isOpen) => {
+    googleTagManager.pushEvent({
+      event: 'open_data_coverage_panel',
+    });
     this.setState({
       isDataCoveragePanelOpen: isOpen,
     });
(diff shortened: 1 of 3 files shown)


edplato:
> Looks good! I requested a couple very minor changes related to event names.
> 
> Dev-meeting discussion ideas sparked from code @edplato @StephanJW @jasontk19 :
> 
> * Add something like https://github.com/rangle/redux-beacon/blob/master/docs/targets/google-tag-manager.md#setup would make the google tag manager code a lot cleaner and easier to manage. I don't know if we could replace every GTM event with it, but most should work.
> * We could create an alias for `googleTagManager` in `webpack` or the `package.json` to avoid needing to add `// eslint-disable-next-line import/no-unresolved` every time it's used.. And likewise add similar alias to similarly-named dependencies.

I like the idea of integrating redux with google tag manager. Only initial extra work I can forsee is where tag manager events are fired and redux isn't currently integrated. I think this is worth making a feature ticket.

We already have `googleTagManager` as an alias in webpack. This allows the `import googleTagManager from 'googleTagManager';` vs. hunting down the relative path where that tag manager code resides: `./web/js/components/util/google-tag-manager.js`.

Benjaki2:
Ahh I didn't realize that was what was causing it. I thought it was casing related or something. It looks like https://www.npmjs.com/package/eslint-import-resolver-webpack does the trick.

We just need to add` "import/resolver": "webpack"` in the settings of the `.eslintrc`.

jasontk19:
Nice find @Benjaki2!

PR #2891, opened 2020-05-13
Daynight layer props

jasontk19:
Add `daynight` properties directly to layers so that we can do faceted filtering by day/night.

Opening this PR separate from the upcoming Product Picker work to reduce overall diff size for review on the latter.


jasontk19:
I need to update this to make all the orbit track `daynight` props use the new array format for consistency.

PR #2904, opened 2020-05-22
Phase 2 product picker

### LINKED ISSUE #2778: Product Picker Improvements - Phase 2
* Faceted filtering on many layer attributes
* ... & more!

TODOs
- [x]  Create distinct “Measurement” and “Search” component to reduce the amount of messy conditional logic
- [x] Address performance issue with many list items (implement react-infinite-scroll for list results)
- [x] Implement more discrete Redux actions/reducers for managing Product Picker state
- [x] Properly set/restore product picker state (including facet filters)
- [x] Layout changes to make room for facet list at all browser sizes
- [x] Tooltips for each facet with details
- [x] Allow collapse/expand toggle for individual facet filter groups
- [x] re-integrate “currently available” filtering toggle
- [x] “chips” for each applied filter to remove individually
- [x] “clear filters” action

jasontk19:
## Description

Fixes #2778

* Adds faceted filtering
* Breaks up product picker components into a number of smaller pieces that connect directly to Redux rather than passing lots of props from parent component down several levels 
* Created more discrete actions/reducers for product picker state (previously relied on a single overloaded "update" action)
* Removed custom responsive state breakpoints (responsive state now lines up with more commonly used breakpoints sizes like what Bootstrap is using)

Note:
* This branch was previously rebased with `optimize-date-range-building` to test these changes in unison so there are a few commits/changes from the associated PR.  I will update this branch once that one is merged to develop.
* A lot of the changes are uninteresting shifts of component logic into smaller pieces.  Since there are a lot of files touched, it may be worth directing your attention to a handful of files that constitute the majority of "new" code here which would include:
  * web/js/modules/product-picker/selectors.js
  * web/js/modules/product-picker/reducers.js
  * web/js/modules/product-picker/formatConfig.js
  * web/js/modules/product-picker/facetConfig.js
  * web/js/modules/product-picker/searchConfig.js
  * web/js/components/layer/product-picker/search-ui-provider.js
  * web/js/components/layer/product-picker/search/facets.js
  * web/js/components/layer/product-picker/search/filter-chips.js
* I am still working on adding some additional E2E tests and, figuring out why some tests are failing in Travis but everything seems to pass locally.


web/js/modules/product-picker/reducers.js
```diff
@@ -1,42 +1,186 @@
...
+        ...state,
+        collapsedFacets: {
+          ...state.collapsedFacets,
+          [field]: !collapsedFacets[field],
+        },
+      };
+    }
+
+    case SELECT_CATEGORY: {
+      googleTagManager.pushEvent({
```
Benjaki2, line 70: I like how you brought this outside of the react views. **dev meeting topic**  gtm middleware. https://github.com/rangle/redux-beacon (I keep forgetting to bring this up)

web/js/modules/product-picker/reducers.js
```diff
@@ -1,42 +1,186 @@
...
+        showMobileFacets: true,
+        category: null,
+        categoryType: 'hazards and disasters',
+        selectedMeasurementSourceIndex: 0,
+      };
+    }
+
+    case TOGGLE_FEATURED_TAB: {
+      const { config } = action;
+      const category = config.categories.featured.All;
```
Benjaki2, line 136: Maybe a getter here?

web/js/modules/product-picker/selectors.js
```diff
@@ -0,0 +1,50 @@
+import {
+  sortBy as lodashSortBy,
+  indexOf as lodashIndexOf,
+} from 'lodash';
+import { createSelector } from 'reselect';
+import buildLayerFacetProps from './formatConfig';
+
+const decodeHtml = (html) => {
```
Benjaki2, line 8: maybe a good util function?

web/js/components/layer/product-picker/search/product-facet.js
```diff
@@ -0,0 +1,98 @@
...
+
+  const renderHeaderIcons = () => (
+    <>
+      <Tooltip
+        className="facet-tooltip-content"
+        isOpen={tooltipVisible}
+        target={`${field}-tooltip-target`}
+        placement="right"
+        toggle={() => toggleTooltip(!tooltipVisible)}
+        delay={{ show: 0, hide: 1000 }}
```
edplato, line 40: I recommend reducing the `hide` time to `300`-ish ms.

jasontk19: Good catch, meant to change this up after the tooltip with a link in it was removed!

web/js/modules/product-picker/reducers.js
jasontk19: Ooo this looks interesting

web/js/modules/product-picker/selectors.js
jasontk19: I thought about it but since it's only used here (and I just broke it out of the existing function below that was moved from another file) I'd like to hold off for now.

web/js/modules/product-picker/reducers.js
jasontk19: Likes this?  Or something else:
```

const getSelectedMeasurementId = (config) => {
  const category = config.categories.featured.All;
  const selectedMeasurement = category.measurements[0];
  return config.measurements[selectedMeasurement].id;
}
  
switch ...
  ...
  case TOGGLE_FEATURED_TAB: {
    const { config } = action;
    
    return {
      ...state,
      categoryType: 'featured',
      category,
      mode: 'measurements',
      selectedMeasurement: getSelectedMeasurementId(config),
    };
  }
```

Benjaki2: I was thinking something like:

`lodashGet(config, 'categories.featured.All');`

So that if `config.categories.featured` isn't there for some reason (someone using a different config), it won't break. But i guess if you are pressing the featured tab it should be there... So probably not necessary.

jasontk19: Ah, gotcha

PR #2905, opened 2020-05-26
No previews dist build

### LINKED ISSUE #2876: Don't include preview images when feature is disabled
Currently we include a feature flag for `previewSnapshots` that turns on/off the fetching of preview images from WVS during build and whether or not those images are shown in the product picker.  However, since the images themselves are committed to the repo they get included in a `dist` build whether you want them or not which can greatly increase the file size of the deployed tgz file.

Streamline this process so that files aren't included when you don't want to use this feature.

**Describe alternatives you've considered**
* Could remove preview image directory from a `dist` build  if `previewSnapshots` is disabled
* Could move the images outside of the code repo

jasontk19:
## Description

Fixes #2876

Remove the preview images folder during dist build if that feature is disabled.

### DIFF
--- a/tasks/dist.js
+++ b/tasks/dist.js
@@ -12,7 +12,11 @@ shell.mkdir('-p', 'build/worldview');
 shell.cp('-rf', 'web/*', 'build/worldview');
 
 // Remove preview images, if that feature is disabled, for smaller dist file size
-const { features } = JSON.parse(fs.readFileSync('config/default/common/features.json'));
+const { CONFIG_ENV } = process.env;
+const featuresConfigPath = CONFIG_ENV
+  ? `config/active/${CONFIG_ENV}/features.json`
+  : 'config/default/common/features.json';
+const { features } = JSON.parse(fs.readFileSync(featuresConfigPath));
 if (!features.previewSnapshots) {
   shell.rm('-rf', 'build/worldview/images/layers/previews');
 }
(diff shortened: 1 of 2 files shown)


mcechini:
I run `IGNORE_ERRORS=true tasks/upload.js -e uat-ext mattc-uat` to do my deployment... does this change get executed when I do that?

jasontk19:
@mcechini The `tasks/upload.js` should be running the `tasks/dist.js` script which is what is modified here.  However, these changes cause it to look up the `features.json` config file where you can disable the `previewSnapshots` feature in order to acheive the desired result of not including the preview images in a dist build (or trying to fetch images for layers during the build process).  

Glancing at the options-gibs repo the other day I think I noticed the `features.json` config file is not located in the same place as the main WV repo which would probably cause some issues with this.

mcechini:
Yeah, our features file will be here:
`config/active/<env>/features.json` where `<env>` is whatever we're providing as a CLI argument (e.g. `uat-ext`)

jasontk19:
@mcechini The latest push should get things working properly by looking in the correct `config/active/<env>/` directory for the `features.json`.  

I am a little fuzzy on how\when the `config/active/release` dir from the options-gibs repo gets used but it seems like the `features.json` from the `<env>` dir should be used instead of the one found in there but let me know if you run into issues with this.

PR #2955, opened 2020-06-12
Consolidate vector styles

### LINKED ISSUE #2937: Consolidate vector styles
Right now each vector layer must have its own vector style. It would be nice if similar layers could be grouped into style JSON files

**Additional context**
This is needed for FIRMS

Benjaki2:
## Description

Fixes #2937

- [x] Allow styles to be grouped in the same document
- [x] Add logic to `layer.daynight.map` since it appears that this field isn't always an array. (#Typescript anyone).

## Further comments

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/modules/product-picker/format-config.js
+++ b/web/js/modules/product-picker/format-config.js
@@ -123,7 +123,10 @@ export default function buildLayerFacetProps(config, selectedDate) {
   return lodashMap(layers, (layer) => {
     setCoverageFacetProp(layer, selectedDate);
     setLayerPeriodFacetProps(layer);
-    if (layer.daynight && layer.daynight.length) {
+    if (layer.daynight && layer.daynight.length && layer.daynight.map) {
+      if (typeof layer.daynight === 'string') {
+        layer.daynight = [layer.daynight];
+      }
       layer.daynight = layer.daynight.map(capitalizeFirstLetter);
     }
     return layer;
--- a/web/js/modules/product-picker/format-config.js
+++ b/web/js/modules/product-picker/format-config.js
@@ -18,6 +18,7 @@ function capitalizeFirstLetter(string) {
 }
 
 function setLayerProp (layer, prop, value) {
+  if (!layer) return;
   if (prop === 'measurements' && (value && value.includes('Featured'))) {
     return;
   }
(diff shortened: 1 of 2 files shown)


web/js/modules/product-picker/format-config.js
```diff
@@ -122,7 +123,10 @@ export default function buildLayerFacetProps(config, selectedDate) {
   return lodashMap(layers, (layer) => {
     setCoverageFacetProp(layer, selectedDate);
     setLayerPeriodFacetProps(layer);
-    if (layer.daynight && layer.daynight.length) {
+    if (layer.daynight && layer.daynight.length && layer.daynight.map) {
+      if (typeof layer.daynight === 'string') {
+        layer.daynight = [layer.daynight];
+      }
```
jasontk19, line 129: Minor possible suggestion (slightly less verbose):
```suggestion
      if (layer.daynight && !Array.isArray(layer.daynight) {
        layer.daynight = [layer.daynight];
      }
```

jasontk19: I guess my question about this field is... should we allow non-array values?  I think it's fine since we can convert it to an array like this but, I guess I was intending it to always be an array in the configs.  

If you enter a non-array value for the `tracks` property in the configs, things break as well
Screen Shot 2020-06-12 at 4 06 32 PM

jasontk19: I guess what I'm getting at is, maybe what would be better is more accurate type checking on the configs themselves at build time?

Benjaki2: Hmm, I didn't mean to include `&& layer.daynight.map`. That overrides the logic I wrote below it, but,  I think you still need the two sets of 'if' statements.

Benjaki2: I totally agree (referenced `Typescript` in description area. I would love to go that route.). I just added this to prevent something potentially breaking on master in the comming release. It definitely was breaking on a non-production config I was just using. I'm not sure if there are layers on master that have this field as a string, but I imagine there might be.

Benjaki2: Should we create a ticket to add some sort of type checking, whether that be in the form of typescript or something else at build-time?

jasontk19: Since this `daynight` property will likely move into the yet-to-be-named visualization metadata service, I wouldn't worry about creating a ticket for now.  I think just doing some safety checks and casting it to an array on the client side is fine for now 👍

mcechini:
I've built my local instance of WV off of this branch and am seeing errors when I attempt to add layer in WV... We should probably get to the point where I can successfully build off of a branch with consolidated vector styles.

PR #2972, opened 2020-06-24
Add time to shared permalink if on current day

### LINKED ISSUE #2942: Add current time to URLs that are shared
**Describe the bug**
Sharing a link while on the current day, the time parameter is omitted in the URL which is used for the full/shortened shared URL. This results in shared URLs without explicit times being parsed to always use the current day.

**Additional context**

- Is it a good idea to always add the time to the URL outside of shared links as well?
- Should we have a checkbox option in the share link dialog to omit the time, so that users can have a link that DOES update daily if that's what they want (e.g., similar to Snapshots)?

edplato:
## Description

Fixes #2942  .

- [x] Add selected date time param to query string in the share copy to clipboard dialog.
- [x] Refactor share tooltips to own component to address issue with their state change causing the share component to re-render two seconds later.
- [x] Random `Date.now()` changes for mini performance/best practices

## Further comments

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/e2e/features/social/share-test.js
+++ b/e2e/features/social/share-test.js
@@ -16,7 +16,7 @@ module.exports = {
   'Clicking the social link button opens the social share dialog': (client) => {
     client.waitForElementVisible(socialCopyLinkButton, TIME_LIMIT);
     client.click(socialCopyLinkButton);
-    client.pause(1000);
+    client.waitForElementVisible(socialToolbar, TIME_LIMIT);
     client.expect.element(socialToolbar).to.be.present;
   },
   'Share link clipboard with existing time query string param in the page url will have the same serialized time': (client) => {
--- a/e2e/features/social/share-test.js
+++ b/e2e/features/social/share-test.js
@@ -1,0 +1,44 @@
+const reuseables = require('../../reuseables/skip-tour.js');
+const localSelectors = require('../../reuseables/selectors.js');
+
+const TIME_LIMIT = 10000;
+
+const {
+  socialToolbar,
+  socialCopyLinkButton,
+  socialLinkInput,
+} = localSelectors;
+
+module.exports = {
+  before(client) {
+    reuseables.loadAndSkipTour(client, TIME_LIMIT);
+  },
+  'Clicking the social link button opens the social share dialog': (client) => {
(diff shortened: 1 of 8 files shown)


web/js/containers/share.js
```diff
@@ -213,10 +210,18 @@ class ShareLinkContainer extends Component {
...
 function mapStateToProps(state) {
-  const { config } = state;
+  const { compare, config, date } = state;
+  const { active, isCompareA } = compare;
+  const { selected, selectedB } = date;
+
+  // selectedDate for shared permalink
+  const selectedDate = !active && !isCompareA
+    ? selectedB
+    : selected;
```
jasontk19, line 220: Seeing this made me realize that I didn't take the right "selected" date into consideration when doing the coverage facet.  Maybe this is minor but, I wonder if it's worth adding a Redux selector to get the selected date which then holds this logic of determining the active compare state first and then returning `state.date.selected` or `state.date.selectedB` accordingly?  If so I can implement when working [this issue](https://github.com/nasa-gibs/worldview/issues/2973).

edplato: That's a great idea. This compare A/B check pattern does repeat itself around the app in a few places.

jasontk19: Would be a good idea to use the `getSelectedDate` selector here now

e2e/features/social/share-test.js
```diff
@@ -0,0 +1,44 @@
...
+} = localSelectors;
+
+module.exports = {
+  before(client) {
+    reuseables.loadAndSkipTour(client, TIME_LIMIT);
+  },
+  'Clicking the social link button opens the social share dialog': (client) => {
+    client.waitForElementVisible(socialCopyLinkButton, TIME_LIMIT);
+    client.click(socialCopyLinkButton);
+    client.pause(1000);
```
jasontk19, line 19: Is there a particular reason you chose to `pause()` for a second here?

I wonder if it would be quicker/more reliable to use `waitForElementVisible(socialToolbar)`?

edplato: It must have been early test animation related that I did, but `waitForElementVisible` works better so I'll make that change.

PR #2985, opened 2020-07-08
Recently Used Layers

### LINKED ISSUE #2719: Consider displaying a "recently used layers" panel
Per user suggestion, it could be helpful if we provided something like a "recently used layers" or "frequently used" panel when the user is adding new layers.

jasontk19:
## Description

Fixes #2719 

- [x] Add recent layers tab to show layers recently added to map
- [x] Show tabs in product picker when using polar projections to allow toggling between recent layers and measurements
- [x] Remove ( i ) icon
- [x] Auto-select first entry in list
- [x] Lighter hover bg color for list entries

@nasa-gibs/worldview


edplato:
I have a related CSS suggestion. I see a slight jump when clicking between the individual layers within the Recent list of layers. I think this is from the top border change, and adding `border: 1px solid transparent` to `.layers-all-layer.search-row` seems to fix this (checked in Chrome).

web/js/modules/product-picker/util.js
```diff
@@ -0,0 +1,81 @@
...
+  safeLocalStorage.setItem(RECENT_LAYERS, JSON.stringify(recentLayers));
+}
+
+/**
+ *
+ * @param {*} layer - the layer being added to the map
+ */
+export function updateRecentLayers({ id: layerId, projections }) {
+  const recentLayersJson = safeLocalStorage.getItem(RECENT_LAYERS);
+  const recentLayers = JSON.parse(recentLayersJson) || lodashCloneDeep(DEFAULT_OBJ);
```
edplato, line 48:
```suggestion
  const recentLayers = (safeLocalStorage.enabled && JSON.parse(recentLayersJson)) || lodashCloneDeep(DEFAULT_OBJ);
```
This should prevent a nonbreaking JSON parse console error if local storage is disabled while adding a layer.

e2e/features/layers/recent-layers-mobile-test.js
```diff
@@ -0,0 +1,53 @@
...
+  layerPickerBackButton,
+} = require('../../reuseables/selectors.js');
+
+const TIME_LIMIT = 10000;
+
+module.exports = {
+  before: (c) => {
+    skipTour.loadAndSkipTour(c, TIME_LIMIT);
+    c.url(`${c.globals.url}?t=2020-07-04`);
+    c.resizeWindow(375, 667); // iPhone 6/7/8 dimensions
```
edplato, line 17:
```suggestion
    c.resizeWindow(375, 667); // iPhone 6/7/8 dimensions
    c.waitForElementVisible(collapsedLayerButton, TIME_LIMIT);
```
I was getting some failed local tests at this stage of the resize. May need to be added elsewhere for mobile changes.

web/js/components/layer/product-picker/browse/recent-layers.js
```diff
@@ -0,0 +1,104 @@
...
+          className="facet-tooltip-content"
+          isOpen={tooltipVisible}
+          target="recent-layer-tooltip-target"
+          placement="bottom"
+          toggle={() => toggleTooltip(!tooltipVisible)}
+          delay={{ show: 0, hide: 300 }}
+        >
+          <p>{recentLayerInfo}</p>
+        </Tooltip>
+        <FontAwesomeIcon
```
Benjaki2, line 43: Having the question mark hover tooltip as a reusable util could be helpful. This icon could also benefit from an on-hover `cursor: help` if it works for all browsers.

web/js/modules/product-picker/util.js
jasontk19: Good catch!  I will probably go with an early return if localStorage is disabled since the rest of the function becomes pointless in those cases:

```
export function updateRecentLayers({ id: layerId, projections }) {
  if (!safeLocalStorage.enabled) {
    return;
  }
...
}
```

e2e/features/layers/recent-layers-mobile-test.js
jasontk19: Thanks, I'll take a look!  Tests seem to be passing locally for me but looks like Travis is failing as of the last commit on this branch. 🤔

web/js/components/layer/product-picker/browse/recent-layers.js
```diff
@@ -0,0 +1,104 @@
...
+  const { browser, productPicker } = state;
+  const {
+    selectedLayer,
+    showMobileFacets,
+    recentLayers,
+  } = productPicker;
+
+  return {
+    browser,
+    smallView: browser.screenWidth < 1024,
```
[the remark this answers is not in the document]
jasontk19: I think we're better off trying to stick to the preset small/med/large/xl sizes where possible... I don't love that I added a new/different horizontal breakpoint for this but given the nature of the views, this size made the most sense to switch from side-by-side to top/bottom list + details.

e2e/features/layers/recent-layers-mobile-test.js
jasontk19: Ah I see, it's Firefox specific.  Nice find!

PR #2990, opened 2020-07-16
Implement React-fast-refresh

### LINKED ISSUE #2980: Investigate implementing React Fast Refresh
[React Fast Refresh is replacing `react-hot-loader`](https://github.com/facebook/react/issues/16604#issuecomment-528663101) which was previously the standard for hot reloading in a React app.  React Fast Refresh ships with React Native currently but is still be worked for React proper.  

There is an experimental webpack plugin [`react-refresh-webpack-plugin` ](https://github.com/pmmmwh/react-refresh-webpack-plugin) that may be worth trying out to see how stable it feels for us.  Since this would only be used in development, and not in production, we could probably set it up as an optional feature to enable when running locally.  That way if it ends up being too unstable devs can just run without it.

jasontk19:
## Description

Fixes #2980.

- [x] disable live reload so page doesn't reload on changes
- [x] bring in react-fast-refresh
- [x] don't transpile `node_modules` in dev mode

A few unrelated dependencies were also updated when I did an `npm audit fix`

@nasa-gibs/worldview


edplato:
I get an error/black app screen on `npm start` or `npm run watch`:
image

I see that `$RefreshSig$` is brought up causing crashes in https://github.com/facebook/react/issues/16604#issuecomment-528663101, but haven't determined the best entry point for our use based on the comment.

Have you encountered any errors like this, or any thoughts that this may be a Windows specific error?

jasontk19:
> I get an error/black app screen on `npm start` or `npm run watch`:
> image
> 
> I see that `$RefreshSig$` is brought up causing crashes in [facebook/react#16604 (comment)](https://github.com/facebook/react/issues/16604#issuecomment-528663101), but haven't determined the best entry point for our use based on the comment.
> 
> Have you encountered any errors like this, or any thoughts that this may be a Windows specific error?

So, that comment thread is probably a little misleading... those are instructions for if you wanted to implement the fast refresh logic into the build process yourself which is what the `@pmmmwh/react-refresh-webpack-plugin` library is doing for us.

I did encounter this error when I was first trying to get things set up before I got the webpack plugin configured properly.  I wonder if some part of those changes is behaving differently on a Windows environment?

jasontk19:
@edplato This is probably the most relevant info related to this error: https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/92#issuecomment-628960771

Not sure why or how this would be different on Windows though, that may take some more investigation.

jasontk19:
@edplato This is a bit of a shot in the dark but.... can you try setting `cacheDirectory: false` in the `babel-loader` plugin options and see if you get different results?  

I am curious if @StephanJW runs into the same issue on this branch.  I may have to try to get Worldview running on my Windows desktop to further troubleshoot this.

edplato:
@jasontk19 It is throwing the same error. I too am curious if @StephanJW is running into this error building in Windows. I'll continue to do some debugging on my end as well.

webpack.config.js
```diff
@@ -66,6 +67,23 @@ if (process.env.TESTING_MODE === 'true') {
...
+];
+// Inlucde any modules that need to be transpiled by babel-loader
+const transpileDependencies = [
+  'react-visibility-sensor',
+];
+if (devMode) {
+  // Don't transpile any dependencies in /node_modules except those found
+  // in transpileDependencies array
+  babelLoaderExcludes.push(
+    new RegExp(`node_modules/(?!(${transpileDependencies.join('|')})/).*`),
```
edplato, line 83:
```suggestion
    new RegExp(`node_modules(?!(/|\\\\)(${transpileDependencies.join('|')})).*`),
```
The `$RefreshSig$ is not defined` error is related to Windows vs Linux module path issues. This change makes the error go away for me, the app to build in watch mode, and the fast refresh to work based on my testing of a few changes in random React components.

jasontk19: Nice!  Good find

PR #2995, opened 2020-07-21
FIRMS Vector layers

Benjaki2:
## Description
Add vector FIRMS layers
Use `vector-proj-w-config-for-testing` to build. It has sources and styles that aren't visible in this PR because they will come from GIBS. 

`?t=2020-03-17-T21` should have all the layers in both arctic and geographic views. Right now there are some inconsistencies between the WMS and Vector styles.

Replace Raster fire layers with vector layers
- [x] Replace layers
- [x] Move vector-interaction code into more isolated component
- [x] Add capability to have breakpoints in polar views
- [x] Add breakpoints
- [x] Add subtitle to metadata tab tooltip
- [x] Use LayerID to distinguish unique tabs (instead of data IDs)
## Further comments

* You may want to look closely at the break point layers, to make sure they correspond to the correct layer selected.
* This PR won't pass Travis until the new layers get up on GIBS PROB

@minniewong To Review config and functionality
@jasontk19 To Review JS code

## Checklist

@nasa-gibs/worldview


web/js/components/map/ol-vector-interactions.js
```diff
@@ -0,0 +1,113 @@
+import {
```
jasontk19, line 1: 👍   I really like this idea of combining a specific _kind_ of map interactions into their own component.  I'm going to try to convert the measure tool to a React component like this one.

minniewong:
@Benjaki2 - I weirdly don't see the last commit "add subtitle to metadata tabs". I just pulled and built it today?

Benjaki2:
> @Benjaki2 - I weirdly don't see the last commit "add subtitle to metadata tabs". I just pulled and built it today? Maybe it's not in your `for-testing` branch

@minniewong I haven't updated it on the `for-testing` branch. That branch is just intended for building the config. If you build there, checkout the `vector-proj-w-config`, and  run `npm run watch`, It should work.

web/js/containers/map-interactions.js
```diff
@@ -120,6 +52,24 @@ export class MapInteractions extends React.Component {
           </>
 
         )}
+        <OlVectorInteractions
+          changeCursor={changeCursor}
```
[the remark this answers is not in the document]
Benjaki2: @jasontk19 I think it's better to lean on the side of connecting more. I believe there are performance benefits (and not to mention pain of prop passing) to using connect instead of prop-passing through numerous levels. In this case I don't think that performance something to consider but it does seem sloppy and feel unnecessary to be prop-checking on both levels when all these props are only relevant to the child. I considered this when isolating this component and skipped it for the sake of time. But, I totally agree with you. I will make map-interactions a folder in the containers area and connect them directly to the child (vector) component. Good call 👍

web/js/components/map/ol-vector-interactions.js
Benjaki2: sounds great. It would be good to normalize as much of that `map/` code as possible

PR #3034, opened 2020-08-10
Download measurements

### LINKED ISSUE #3005: Historical geostationary imagery not accessible
**Describe the bug**
The changes in #2893 to provide more accurate user feedback for geostationary imagery availability do not account for historical date ranges being kept for each layer.  

**To Reproduce**
View the Hurricane Dorian tour story.

**Expected behavior**
Historical ranges should be visible and available

### LINKED ISSUE #3006: Allow downloading measurements as GeoJSON and/or shapefiles

jasontk19:
## Description

Fixes #3006

Allow downloading measurements as GeoJSON ~~or Shapefiles~~.

I tested the output files in QGIS to confirm they works. To be able to view GeoJSON or Shapefile layers in polar projections, I had to assign those specific layers the correct projection.  I believe this is expected since the GeoJSON spec does not allow for setting a projection (that property was deprecated) and, the Shapefiles are being generated based on the GeoJSON.  Since working in polar projections seemed to take a lot of fiddling in QGIS regardless, this seemed acceptable.

@Benjaki2 -> code review
@minniewong  -> testing

@nasa-gibs/worldview


jasontk19:
False alarm... one last thing I wanted to check out:  Measurements don't get exported with their great circle arc transformations applied which makes large measurements look a _little_ bit different when exported.

minniewong:
When I try to load a shapefile with multiple area and distance measurements, the attribute table in QGIS only displays one of the results, though the shape appears on the map (and the dbf file shows both entries too). I *think* you want to include a unique id field. I'm also *think* it's better to have the units separately in another field.

jasontk19:
Screen Shot 2020-08-12 at 10 11 25 AM

With some changes I was able to get each feature to show as a separate entry but only when loading as GeoJSON.  I'm not sure why the Shapefile features aren't showing as separate entries but I'm assuming it has something to do with what's going on under the hood in the `shp-write` library I'm using to create them.  How important is this functionality and, what are your thoughts on offering only GeoJSON?

web/js/modules/measure/reducers.js
```diff
@@ -1,21 +1,40 @@
...
+  CHANGE_UNITS,
+  UPDATE_MEASUREMENTS,
+  TOGGLE_MEASURE_ACTIVE,
+} from './constants';
 
 const defaultState = {
   isActive: false,
   unitOfMeasure: 'km',
+  allMeasurements: {
+    'EPSG:3413': {},
```
Benjaki2, line 11: Suggestion: pull the spacial reference from the config so it could work with `EPSG:3857` as well (on non-production versions of Worldview).

PR #3058, opened 2020-08-25
Migrate EONET api, filter categories

### LINKED ISSUE #2057: Provide ability to filter events by type
**Is your feature request related to a problem? Please describe.**
Some users come to Worldview to only look at a certain type of event, they have to scroll through a long list of events to find a certain type of event. 

**Describe the solution you'd like**
We should provide the ability for users to filter the events by type. This may especially be useful for fire events.

### LINKED ISSUE #2634: Migrate to EONET V3
Currently use `v2.1` of the EONET API

Migrate to the newly released `v3.0`

**Additional context**
v2.1 is going to continue to be reachable... v3.0 gives access to a new `magnitude` field, it would be nice to visualize the magnitude.

jasontk19:
## Description

Fixes #2634
Fixes #2057 

* Migrate to v3 API
* Add a dropdown to select categories.  Instead of showing all EONET provided categories, we're only showing the ones that events were returned for. 

@nasa-gibs/worldview


minniewong (approved):
Looks really nice! 🎸

web/js/containers/sidebar/events.js
```diff
@@ -183,8 +207,17 @@ function mapStateToProps(state) {
   let visibleEvents = {};
   const events = lodashGet(requestedEvents, 'response');
   const sources = lodashGet(requestedEventSources, 'response');
+  const eventCategories = (events || []).reduce((categories, event) => {
```
Benjaki2, line 210: Maybe this could be put into `eventsRequestReducer` so this loop is only run once?

Benjaki2 (approved):
Looks really good! I like the category selector. 🔥 🌪️ 🧊

web/js/containers/sidebar/events.js
jasontk19: Looking at this more closely, it felt more like a value I was trying to derive from an existing state value, so I went with a memoized state selector instead, which will still only run once.

PR #3070, opened 2020-08-28
Fix sources facet inconsistencies

### LINKED ISSUE #3061: Facets source issue
**Describe the bug**
Because of how the measurement configs are set up, a layer may end up under multiple differently named sources which can cause unexpected results when the `sources` facet is derived from the configs:

Screen Shot 2020-08-26 at 1 51 43 PM
 
In this example only one of the OCO-2 orbit track layers is showing under the `OCO-2` source filter while the other is showing under the `OCO-2 / SpaceTrack.org` filter.  Both orbit tracks should be showing under the latter.

jasontk19:
## Description

Fixes #3061 

Switched to use the `subtitle` to derive the sources facet since the same information is in this field and this approach produces more consistent facet choices.

### DIFF
--- a/web/js/modules/product-picker/format-config.js
+++ b/web/js/modules/product-picker/format-config.js
@@ -22,10 +22,10 @@ function setLayerProp (layer, prop, value) {
   if (!layer || featuredMeasurement) {
     return;
   }
-  if (layer[prop] && !layer[prop].includes(value)) {
-    layer[prop].push(value);
-  } else if (value) {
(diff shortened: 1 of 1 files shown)


jasontk19:
Wow, good catch... there was some bad logic around where we set facet properties that was sometimes causing all the values set up to that point to be cleared out.  The result is that some layers weren't showing up under some of the facets they should have.  As a result, the counts for many facets changed quite a bit 👍

PR #3071, opened 2020-08-28
3067 fix social links

jasontk19:
## Description

Fixes # 3067


jasontk19:
@edplato I think it's because we are doing those asynchronously rather than as the direct result of a user click: https://stackoverflow.com/questions/9514698/bypass-popup-blocker-on-window-open-when-jquery-event-preventdefault-is-set

jasontk19:
Ok I think I see why we were previously trying to do `window.open` to a blank page and then set the location after .... https://stackoverflow.com/a/32908311/1132892

I'll update this.

edplato:
Interesting, that makes sense. Good catch on proposed solution.

jasontk19:
I think the real bug here was just that `status_code` is not part of the bitly response anymore... Anyway, it all seems to be working properly now.

PR #3088, opened 2020-09-08
Fix config issues

jasontk19:
## Description

* Fix an issue where orbit track name was misspelled
* Fix an issue where the Settlements config had no `id` property

@nasa-gibs/worldview


minniewong (approved):
👍

PR #3126, opened 2020-09-22
Revert history 5.0.0 update

jasontk19:
## Description

Looks like this update is causing problems:
Screen Shot 2020-09-22 at 2 24 01 PM


edplato (approved):
👍 5.0.0 breaks for me too, not seeing an obvious cause. This fixes it!

PR #3131, opened 2020-09-25
Category group config improvements

### LINKED ISSUE #3130: Category groups should be less hard-coded, more config-based
Currently there are a number of improvements that could be made to the way we define and use the category group names (e.g. `hazards and disasters`, `science disciplines`).  These group names are used as the tab names in the browsing mode of the product picker.  They are defined in the category config files but also hard-coded into the JS in several places.

* We should define a `categoryGroupOrder.json` config file to define the order, instead of doing so in JS
* We should remove any hardcoding or expectations of specific category group names that exist in the code so that existing groups could be removed, or new groups could be added to the config and simply work without any code changes being necessary.

**Additional context**
There was an Earthdata support ticket from some folks who wanted to stand up their own instance of WV who inquired about how to add new category groups that highlighted the need for this improvement.

jasontk19:
## Description

Fixes #3130

See issue description for details on improvements made.

- [x] derive category names from config
- [x] rename folders to match category names
- [x] define category group order in configs


config/default/common/config/wv.json/categories/science_disciplines/Terrestrial Hydrosphere.json
```diff
@@ -1,6 +1,6 @@
 {
     "categories": {
-        "scientific": {
+        "science disciplines": {
```
Benjaki2, line 3: We probably should set up some JSON style linter (I haven't don't this before but my vscode does some correcting 🤷 ). I [don't think you want spaces for property names](https://stackoverflow.com/a/19287394/4589331) although it does work and we probably have them throughout the codebase such as is in the line below.

I think it would be more stylistically correct to say something like
```
categories": [
  {
    id: "cat_1",
    name: "science disciplines",
    someOtherName: 'some value'
  }
]
```

Benjaki2 (approved):
Looks great! Added one comment that is more of a topic for a developer meeting than something that should be changed here.

config/default/common/config/wv.json/categories/science_disciplines/Terrestrial Hydrosphere.json
jasontk19: Hmmm yeah this is a good point.  Didn't really cross my mind since we are already violating this rule elsewhere (e.g. "hazards and disasters").  I wonder what the real consequences of having a space here might actually be for us?  I could see needing to avoid spaces if we were generating some JSON as part of an API response that would be consumed by different machines/operating systems/environments that may have different expectations of formatting.  

JSON linting would be nice, especially based on a JSON schema.  I think having a JSON schema for these kinds of config files could go a long way towards helping third parties better understand what the expected content, and meaning of these configs.  Perhaps it could also help streamline adding new layers for the uninitiated.

I will add a topic to the next tech dev meeting agenda!

PR #3134, opened 2020-09-30
Fix issues with recent layers/projections

### LINKED ISSUE #3133: Can't add layers when configured projections don't match those in local storage recent layers object
**Describe the bug**
You can run into an issue where, if the projections defined in your local storage object for recent layers doesn't match up with the projections that the app is currently running with.

**To Reproduce**
1. Build and run app with only `arctic`, `antarctic`, `geographic` projections configured
2. Add at least one layer to generate the recent layers object in local storage
3. Confirm recent layers projection keys only include the mentioned projections in step 1:
* In console run `Object.keys(JSON.parse(localStorage.getItem('recentLayers')))`
4. Clone `worldview-options-gibs` repo into `config/active` directory, 
5. Rebuild and deploy the app, which will include an additional configured projection: `webmerc`
6. Try adding a layer with this projection

Result:  Clicking the checkbox will not add the layer, and the following console error will be shown:
* `Uncaught TypeError: layers is undefined`

**Expected behavior**
Recent layers object in local storage should be amended with keys for additional projections if and when they are encountered.

jasontk19:
## Description

Fixes #3133 

### 1) Confirm the bug:
1. Clear local storage
2. Build run WV app from `develop`
3. Open app, add at least 1 layer
4. Build and run WV from `develop`, with the `worldview-options-gibs` repo sitting in the `config/active` directory
5. Open app, open product picker
6. Search for `AIRS_L2_Surface_Skin_Temperature_Day` and try to click checkbox to add this layer
7. Expect that layer won't be added
8. Go to recent layers tab
9. Try to remove layer that was added in Step 3
10.  Expect layer won't be removed

### 2) Confirm the fix:
1. Switch to this branch, confirm UI updates (no need to rebuild)
2. Repeat step 6
3. Expected that the layer is added to the map as expected.
4. Go to the recent layers tab
5. Try to remove layer that was added in Step 3
6. Expect that layer is removed properly

### DIFF
--- a/web/js/modules/product-picker/util.js
+++ b/web/js/modules/product-picker/util.js
@@ -9,6 +9,17 @@ function getDefaultObj (projections) {
   return DEFAULT_OBJ;
 }
 
+function getLayersCheckProjections (allProjections) {
+  const recentLayersJson = safeLocalStorage.getItem(RECENT_LAYERS);
+  const recentLayers = JSON.parse(recentLayersJson) || getDefaultObj(allProjections);
+  if (Object.keys(recentLayers).length !== allProjections.length) {
+    allProjections.forEach((proj) => {
+      recentLayers[proj] = recentLayers[proj] || [];
+    });
+  }
+  return recentLayers;
+}
+
 export function getRecentLayers(layerConfig, proj) {
   const byUse = (a, b) => {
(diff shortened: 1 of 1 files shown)


edplato (approved):
Works good 👍

PR #3164, opened 2020-10-16
Integrate futureLayers

### LINKED ISSUE #3162: Fully integrate futureLayers into the app
We have had the option to add the `futureLayers`/`futureTime` parameters into the layer options and now have user interest to make this feature operational. 

Allow timeline axis/data coverage/available animation end-date and related layer/sidebar UI elements/text to coincide with available `futureLayers` max time while retaining normal functionality when no `futureLayers` are available.

### LINKED ISSUE #3163: Data panel incorrect date ranges on some updates
**Describe the bug**
Data panel incorrect date ranges on some updates due to reuse of non-unique `key` in child component. Toggling hide/show of sidebar layers at certain timescale zoom levels can cause the child component to not update and result in the wrong data panel line coverage for the target layer.

edplato:
## Description

Fixes #3162 .

This feature integration allows the use of layers with future coverage set using the options `futuerLayer` and `futureTime` in the individual layer configuration files. Note, both parameters are necessary. Since the future date is dynamically created in the interface, the future date is floored to the nearest quarter hour to prevent rapid updates throughout the app.

- [x] Update permalink handling to allow future dates if a future layer is included
- [x] Integrate future layer into timeline axis (max end date), sidebar layer info, animation, and data panel
- [x] Add/update tests, add `mockFutureLayer` test parameter, and add helper util functions

Additional:
- [x] Add unique keys for data panel containers - Fixes #3163 
- [x] Fix failing test for right date picker arrow being disabled when UTC hours < 3

## Further comments

Test by using the `mockFutureLayer` url parameter like this:
`http://localhost:3000/?mockFutureLayer=VIIRS_NOAA20_CorrectedReflectance_TrueColor,5D`

## Checklist

@nasa-gibs/worldview


jasontk19:
I still need to pull and test this but, I am currently sketching out a JSON schema for our layer configs and going through all the properties and something came to mind.  I am wondering if we need both `futureLayer` and `futureTime` properties for this functionality?  Or could we just use the presence of a value for `futureTime` to also be the indicator that this is a future layer, instead of relying on `futureLayer: true` to be defined?

edplato:
That's a good point. I don't see any value in having both. I'll get rid of `futureLayer` throughout.

web/js/components/timeline/timeline-data/data-item-container.js
```diff
@@ -109,13 +120,22 @@ class DataItemContainer extends Component {
...
 
+    // layer options
     const {
-      endDate,
+      futureTime,
       id,
       startDate,
       visible,
     } = layer;
+    let {
```
jasontk19, line 130: Modifying the layer's `endDate` property here in the render function of a deeply nested component seems risky. Since we're just modifying a date slightly, this probably won't introduce any bugs but as a general practice I think it's best avoided in cases like this where other parts of the app are reading this property.  If you were, for example, debugging something elsewhere in the app and looking at the `endDate` property and you noticed it didn't match up with what's in the `wv.json`, you might have a tough time tracking down where this is being modified.

Additionally, calling `getFutureLayerEndDate` on every render seems unnecessary since it shouldn't be changing once the app loads.  It looks like you are having to call this in several other places throughout the app as well which leads me to think maybe it's best to find a place to call it once and be done.

Perhaps a solution to both is to handle all `endDate` modifications in a central place early on in the app lifecycle?  I had to do something like this to adjust `startDates` for geostationary layers and did it [here in `main.js`](https://github.com/nasa-gibs/worldview/blob/92d27c795321ae13a794bb0d9027dad05aee0ac6/web/js/main.js#L174).

jasontk19 (changes requested):
Overall functionality looks good!  Just a few code style/best practices comments.

web/js/components/timeline/timeline-data/data-item-container.js
edplato: Yeah changing the `endDate` there is dangerous. I'll get that change up.

For the `getFutureLayerEndDate` I was under the initial impression that the future end dates needed to be dynamic and keep updating if a user left their browser open, but on second thought based on what you said, adjusting the `endDate` early in the app vs repeatedly makes more sense for the use case as users can always refresh.

web/js/main.js
```diff
@@ -182,7 +184,9 @@ window.onload = () => {
         }
       });
     }
-
+    if (parameters.mockFutureLayer) {
```
jasontk19, line 187: So I noticed that if you move this call to `mockFutureTimeLayerOptions` above the call to `adjustEndDates`...

web/js/modules/layers/util.js
```diff
@@ -1288,3 +1295,59 @@ export function adjustStartDates(layers) {
...
+
+/**
+ * Build end date for future layer
+ *
+ * @method mockFutureTimeLayerOptions
+ * @param  {Array} layers array
+ * @param  {String} mockFutureLayerParameters 'targetLayerId, mockFutureTime'
+ * @returns {Array} array of layers
+ */
+export function mockFutureTimeLayerOptions(layers, mockFutureLayerParameters) {
```
jasontk19, line 1334: ... that this `mockFutureTimeLayerOptions` only has to add the `futureTime` property to the layer specified in the URL params and then `adjustEndDates` will do the rest, allowing this function to be simplified to:

```
export function mockFutureTimeLayerOptions(layers, mockFutureLayerParameters) {
  const urlParameters = mockFutureLayerParameters.split(',');
  const [targetLayerId, mockFutureTime] = urlParameters;

  if (targetLayerId && mockFutureTime) {
    layers[targetLayerId].futureTime = mockFutureTime;
  }
}
```

That way you don't have to repeat the logic of of setting the new endDate, or loop through all the layers again.

jasontk19 (approved):
Looks good! The mockFutureLayer test URL string provided tripped me up a bit since I didn't notice that the layer being adjusted was not the one shown when the page loads (since it's hidden by default and I didn't look closely enough at the layer ID).  So would just say that, for others reviewing or testing this, maybe try: 
`http://localhost:3000/?mockFutureLayer=MODIS_Terra_CorrectedReflectance_TrueColor,5D` 
... which will load the app with the the default visible layer being the one that is adjusted with a `futureTime`.

Other than that, just noticed a small simplification that could be made to the `mockFutureTimeLayerOptions` logic.

web/js/modules/date/util.js
```diff
@@ -24,13 +30,90 @@ export function tryCatchDate(str, initialState) {
...
+ * @param  {Object} config object
+ * @returns {Object} date object
+ */
+export function parsePermalinkDate(defaultStr, str, layerParameters, config) {
+  let time = tryCatchDate(str, defaultStr);
+  if (time instanceof Date) {
+    const startDate = new Date(config.startDate);
+    if (time < startDate) {
+      time = startDate;
+    } else if (time > defaultStr) {
```
Benjaki2, line 49: Looks like `time > defaultStr` is comparing a `Date` to a date `String`. Not sure if that works? Might be worth investigating.

Benjaki2 (approved):
Works great! 👍

web/js/modules/date/util.js
edplato: Good catch, thanks! There was a wrong method description and variable name that I addressed.

PR #3165, opened 2020-10-20
Remove map zoom buttons on mobile

### LINKED ISSUE #3156: Remove map zoom buttons in mobile
Perhaps a good place to start for improving the mobile UI is removing the map zoom buttons when the app is in mobile. I notice major mapping platforms interfaces like Google Map and Bing Maps do not have zoom buttons in mobile, so I expect users to be familiar with the zoom pinch interaction necessary to zoom in/out of the Worldview map.

**Additional context**
I think we can remove the rotate buttons in the mobile UI at the same time; however, I could see keeping them since the use case of these buttons may be rare enough for users when compared to other mapping applications they're familiar with.

edplato:
## Description

Fixes #3156  .

- [x] Remove map zoom buttons on mobile and handle dynamic re-adding, event listener removal, and minor `isMobile` replacement of `isSmall`

## Further comments

Note: This is a precursor to further style related work within the geosearch feature branch of UI buttons including the toolbar, zoom, rotate, measure buttons, and consistent mobile breakpoints app-wide. I wanted to get this out since it is more functional.

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/map/ui.js
+++ b/web/js/map/ui.js
@@ -220,9 +220,7 @@ export default function mapui(models, config, store, ui) {
    */
   function updateProjection(start) {
     const state = store.getState();
-    const { browser, proj } = state;
-    const isMobile = browser.lessThan.medium;
-    const hasZoomButtons = mapHasZoomButtons(proj.id);
+    const { proj } = state;
     if (self.selected) {
       // Keep track of center point on projection switch
       self.selected.previousCenter = self.selected.center;
@@ -242,11 +240,6 @@ export default function mapui(models, config, store, ui) {
     // Update the rotation buttons if polar projection to display correct value
     if (isProjectionRotatable) {
       rotation.setResetButton(currentRotation);
-    }
-
-    // if no map zoom buttons, add them
-    if (!isMobile && !hasZoomButtons) {
-      createZoomButtons(map, proj.selected);
     }
 
     // If the browser was resized, the inactive map was not notified of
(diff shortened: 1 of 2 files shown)


web/js/map/ui.js
```diff
@@ -958,10 +986,44 @@ export default function mapui(models, config, store, ui) {
...
+    const zoomOutButton = document.querySelector(`#wv-map-${proj} .wv-map-zoom-out`);
+    return zoomInButton && zoomOutButton;
+  }
+  /*
+   * @method removeZoomButtons
+   * @static
+   *
+   * @returns {void}
+   */
+  function removeZoomButtons() {
```
jasontk19, line 1010: Is there a reason we can't just use CSS media queries to hide them at mobile browser widths?

edplato: My thought process was to not add if `isMobile` and if a resize changed to `isMobile` then removing the buttons also gets rid of event listeners associated with the zoom buttons.

jasontk19: Hmmm makes sense.  I am lazy and would have done the 4 line CSS solution:
```
@media (max-width: 740px) {
  .wv-map-zoom-in, .wv-map-zoom-out {
    display: none;
  }
}
```

... but testing each solution they look/feel exactly the same from a user POV so, I think this is fine as is!

jasontk19 (approved):
👍

web/js/map/ui.js
Benjaki2: I would also lean towards what @jasontk19 stated above (to avoid more mapUI code, and css before js is a general rule of thumb for me)  or just add a mobile class to the outmost div and not worry about the add/remove controls.

Benjaki2 (approved):
Mobile looks better this way 👍

edplato:
Well, I'm obviously outnumbered on this proposed solution. I switched to use css solution. Breakpoint mobile consistency changes will be incoming on geosearch PR.

PR #3168, opened 2020-10-22
3167 encode ids when used as CSS selectors

### LINKED ISSUE #3167: Periods in layer ids break the UI (again)
**Describe the bug**
If a layer id includes a period, the UI will break when that layer is included on the map since there is code that is trying to query for HTML elements that have ids based on layer ids.

**To Reproduce**
1. Go to `http://localhost:3000/?l=Particulate_Matter_Below_2.5micrometers_2001-2010` 

**Expected behavior**
Periods in layer ids should not break the app.

e2e tests that add layers with periods in their id to the map should be added to protect against reintroducing this issue.

jasontk19:
## Description

Fixes #3167.

- [x] encode ids
- [x] add tests with layer that needs encoding 

@nasa-gibs/worldview


edplato (approved):
Looks good to me 👍

PR #3189, opened 2020-11-10
Inactive events alert

### LINKED ISSUE #3186: Permalinks with closed EONET events should show an alert
**Describe the bug**
If you load the app with a closed event ID in the permalink, the alert that is supposed to inform the user that this event is unavailable does not show.  Instead, trying to load any _other_ event will cause the alert to incorrectly show.

**To Reproduce**
1. Go to https://worldview.earthdata.nasa.gov/?e=EONET_5133
2. No alert shows
3. Cick any event in the list
4. The alert incorrectly states that the selected event is no longer active

**Expected behavior**
The alert should show only for the event included in the permalink.

jasontk19:
## Description

Fixes #3186.

- [x] make sure alert shows only when event is not available
- [x] replace old alert type with new one

## To Test:

1. Go to [http://localhost:3000/?e=EONET_5133](http://localhost:3000/?e=EONET_5133)
2. Confirm alert shows
3. Close alert
4. Click various other events, confirm they load properly 
5. Confirm alert does not show when clicking other events

### DIFF
--- a/web/js/map/natural-events/ui.js
+++ b/web/js/map/natural-events/ui.js
@@ -8,11 +8,8 @@ import util from '../../util/util';
 import { CHANGE_TAB as CHANGE_SIDEBAR_TAB } from '../../modules/sidebar/constants';
 import * as EVENT_CONSTANTS from '../../modules/natural-events/constants';
 import { activateLayersForEventCategory } from '../../modules/layers/actions';
-import {
-  deselectEvent as deselectEventAction,
-  selected as selectedAction,
-} from '../../modules/natural-events/actions';
-import { getDefaultEventDate, naturalEventsUtilGetEventById } from './util';
+import { selected as selectedAction } from '../../modules/natural-events/actions';
+import { getDefaultEventDate } from './util';
 import { selectDate } from '../../modules/date/actions';
 import { UPDATE_MAP_UI } from '../../modules/map/constants';
 import { LOCATION_POP_ACTION } from '../../redux-location-state-customs';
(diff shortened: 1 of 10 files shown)


minniewong (changes requested):
If you go to the inactive link and leave the inactive event alert box up and click on other events, the event ID updates to that event in the inactive alert box (even though the event is definitely active).

Screen Shot 2020-11-12 at 10 34 52 AM

web/js/map/natural-events/ui.js
```diff
@@ -314,7 +309,7 @@ export default function naturalEventsUI(ui, config, store, models) {
       self.markers = naturalEventMarkers.draw();
       naturalEventsTrack.update(null);
     }
-    store.dispatch(deselectEventAction);
+    // store.dispatch(deselectEventAction);
```
Benjaki2, line 312: This dispatch does seem weird and looks like it could cause some sort of never ending dispatch loop but removing it seems to also introduce an error where an event can be active in the "map" state but not in the "Global" state.

to reproduce:
1. Click on event in middle of map.
2. change to polar proj.
3. return to geo proj.
4. see error : Event is selected in list but not on map

Benjaki2 (changes requested):

web/js/map/natural-events/ui.js
jasontk19: Hmmm, yeah this seemed like either a) it wasn't doing anything since `deselectEvent` what is being called when a deselect event action is dispatched or b) was causing an infinite loop of dispatching.

With this line still commented out, I wasn't able to reproduce this specific error with the few events I tried.  Is there a specific event that is breaking this for you?

minniewong (approved):
Working great now!

web/js/map/natural-events/ui.js
Benjaki2: I'm not seeing the error anymore. 🤯

Benjaki2 (approved):
🎉

PR #3193, opened 2020-11-17
Geocode Search Feature

### LINKED ISSUE #1688: Epic - Gazetteer / Omni Search
- gazetteer for finding locations by name
- ability to position map by lat/lon coordinates
- ability to add products directly from the search box

edplato:
## Description

Fixes #1688  .

Geocode Search Feature (aka Geosearch)

- [x] This feature uses the [ArcGIS World Geocoding Service](https://developers.arcgis.com/rest/geocode/api-reference/overview-world-geocoding-service.htm) to find addresses based on input text, coordinates, and by clicking a spot on the map.

  Use the Geosearch component located with the upper right toolbar buttons to type in text and get suggestions of place names to select. Enter place names for auto suggested results or coordinates (Note: coordinates will not auto suggest once detected). Selected places will add a map marker and zoom/fly to target location (based on best zoom level of currently added layers). 
image

  Activate reverse geocoding by clicking this button. Now click on the map to instantly add a marker and get that place's location metadata.
image

  Added map marker with show as a map pin with an interactable location metadata dialog that is displayed on initial map maker creation and by clicking on the map marker. The dialog contains place details and coordinates.
image

  The coordinates dialog can be minimized by clicking the minimize button:
image

  The coordinates can be copied to the clipboard by clicking the copy button:
image

  The coordinates dialog and map marker can be removed/destroyed by clicking the remove button:
image

- [x] External [Worldview Snapshots](https://wvs.earthdata.nasa.gov/) image download feature improvements now allow map markers to be added to downloaded imagery. 
- [x] UI toolbar button resizing, map zoom/rotation button positioning, mobile breakpoint styling consistency fixes.

## Checklist

@nasa-gibs/worldview


web/js/modules/geosearch/reducers.js
```diff
@@ -0,0 +1,51 @@
...
+  reverseGeocodeResults: null,
+};
+
+export function geosearchReducer(state = geosearchState, action) {
+  switch (action.type) {
+    case UPDATE_ACTIVE_MARKER:
+      return lodashAssign({}, state, {
+        activeMarker: action.value,
+        reverseGeocodeResults: action.reverseGeocodeResults,
+      });
```
jasontk19, line 26: We shouldn't need to use `lodashAssign` for this, the spread operator in ES6 will handle this natively:

```
return { 
    ...state,
    activeMarker: action.value,
    reverseGeocodeResults: action.reverseGeocodeResults,
};
```

web/js/modules/geosearch/selectors.js
```diff
@@ -0,0 +1,200 @@
...
+  const vectorLayer = new OlLayerVector({
+    source: vectorSource,
+  });
+
+  return vectorLayer;
+};
+
+// ArcGIS World Geocoding Service API Requests and Options
+const GEOCODE_OPTIONS = {
+  urlBase: 'https://geocode.arcgis.com/arcgis/rest/services/World/GeocodeServer/',
```
jasontk19, line 134: Should this URL go into `features.json` along with our events, notifications, and snapshots API URLs?

web/js/modules/geosearch/selectors.js
```diff
@@ -0,0 +1,200 @@
+import OlPoint from 'ol/geom/Point';
```
jasontk19, line 1: A lot of the functions in this file don't strike me as true "selectors".  A selector is a function that accepts Redux state as an argument and returns data that is derived from that state.  The only function here that looks like it might fit this description would be `areCoordinatesWithinExtent`

[This article gives a great explanation of how and why to use selectors as well as how to memoize them with Reselect for better performance.](https://blog.isquaredsoftware.com/2017/12/idiomatic-redux-using-reselect-selectors/)

jasontk19 (changes requested):
Overall I think this is an awesome feature and is executed really well here. I especially like how small and well organized the individual components and new functions are. 👍 

A few possible UI changes are described here and code style changes are mentioned in line as individual comments:

- [ ] I've noticed that if you search for a location for which there are no results, you don't get any feedback in the UI.  So, a user might just keep hitting enter and wonder why it's not working. I'm not 100% sure we need something or what it would look like but it may be worth a discussion. Here's an example address where I've just typo'd the street name: `10513 Senagal dr, pensacola`

- [x] The "close" icon in the measure tooltip looks out of alignment now:
Screen Shot 2020-11-18 at 9 36 59 AM

- [x] Maybe we can use an icon other than the default "warning" for this alert?  Either an "info" or location pin might be more appropriate:
Screen Shot 2020-11-18 at 9 38 41 AM

- [x] Do we need to spell out `Latitude: #### Longitude: ####` in the tooltip?  I think that just the values themselves (e.g. `####, ####`) might reduce business within this tooltip and, would more accurately reflect what gets copied to the clipboard when the user takes that action.  The current formatting presents them as if they are totally separate values when in practice, they are more like two parts of a single value.
Screen Shot 2020-11-18 at 9 44 21 AM

- [x] Looks like the search input still sends a request for every single character change.  We should probably debounce this:
Screen Shot 2020-11-18 at 9 52 47 AM

- [x] I am seeing some weird behavior on mobile (using Chrome devtools to emulate):
1. Open geosearch, type in "america", apply search
2. Click the "Add Marker" button, and tap below on the map somewhere
3. New marker is added across the world in Iran 🤔

Benjaki2 (changes requested):
It looks really good!

I added a few code-style suggestions

edplato:
I worked out your suggestions changes and made some major refactors, so may want to take another look. Thanks for the good feedback all around.

jasontk19:
When placing a marker, there is a noticeable delay between when the user clicks and when the marker is placed.  This is causing an issue where, if you click the map to place the marker then immediately move the mouse, the marker is placed at a different location than where the user has clicked.  

In this video, I am clicking in the far bottom left corner then moving the mouse upward to the right.  Notice that the marker never gets placed in the bottom left corner where I start but instead ends up at some other location where the mouse has been moved: https://drive.google.com/file/d/1mvCfH6AGDiYCbvDDIWvDGYTA7z0ksYYy/view?usp=sharing

web/js/components/geosearch/geosearch-modal.js
```diff
@@ -0,0 +1,459 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      isTouchDevice: false,
+      showInputAlert: false,
+      showExtentAlert: false,
+      showReverseGeocodeAlert: false,
+      showNoSuggestionsAlert: false,
+    };
+    this.throttleGetSuggestions = lodashThrottle(this.getSuggestions, 400, { leading: true, trailing: true });
```
jasontk19, line 42: I think a debounce might actually make more sense for this than a throttle.  The reason being that, when using debounce no requests will be made while the user is actively typing (depending on the configured delay and the user's typing speed).  While using throttle, requests still get made every `n` milliseconds whether or not the user has finished typing.

In my test example, I typed in "Panama City, FL" and ~10 calls were made to the server during that time but the results along the way were not of any use to me (e.g. just searching for "panam")

Screen Shot 2020-12-04 at 8 43 06 AM

If this were using a debounce, and depending on my typing speed, we would likely only see ~1-2 requests being made.  

Now, these request are all really really tiny and they are coming back at a blazing fast speed so this may only really matter if there is a cap on how many requests to this endpoint we are allowed to make.  However, I think it's good API use etiquette to limit the number of requests the application makes as much as is practical, especially if there aren't any significant benefits gained by making 10x as many.

jasontk19: Oh and, since I couldn't remember exactly how throttle differed I Google'd it and found this really great article with visual examples that I think is worth sharing: https://redd.one/blog/debounce-vs-throttle

jasontk19:
This probably isn't a big deal since it's an unlikely use case but, I noticed that if you copy the coordinates from a place that you searched for and paste them back into the search input it tells you there are no suggestions (since I'm guessing it's trying to do the suggest API call rather than the reverse lookup by coordinates):
Screen Shot 2020-12-04 at 8 57 48 AM

jasontk19:
We probably don't need hover tooltips and title attr tooltips on buttons:
Screen Shot 2020-12-04 at 9 02 14 AM (2)

jasontk19:
Collapsing the search input after searching for a place name, then expanding it causes the suggest options to be shown.  We probably don't want to show the suggestions unless the user was just actively typing in the input.

jasontk19 (approved):

Benjaki2 (approved):
Looks great! 🎉 🎉 

One minor maybe bug I noticed I noticed:

1. Search and go to `Anne Arundel`
2. Switch to antarctic projection
3. Go back to geographic projection
4. see minor error: Point is still selected but name has changed to `Annapolis`

jasontk19:
A few minor things I thought of:

1. The docs for URL params should be updated with the new one for the map marker: `doc/url_parameters.md`
2. The new "marker" URL param breaks the existing pattern of having letter or acronym-based URL params, instead using a full word.  I'm not sure what the motivation was to switch params over to using a minimal number of characters (presumably a concern for overall URL length?) but, it may be worth changing this one for the purposes of consistency.

PR #3195, opened 2020-11-19
Smart handoffs

StephanJW:
## Description

This is the smart-handoffs component that overhauls data-download page. Users will now have the option to single select a layer, see the amount of granules that are pertaining to that layer based on selected date and bounding box (if applicable), and proceed to download specified granules within NASA's Earthdata Search tool.

@nasa-gibs/worldview


web/js/containers/sidebar/smart-handoff.js
```diff
@@ -0,0 +1,470 @@
...
+
+    const { selectedLayer, isSearchingForGranules } = this.state;
+    const {
+      selectedGranules, totalGranules, coordinates, currentExtent, showBoundingBox,
+    } = this.state;
+
+    // Determine if data-download 'smart-handoff' tab is activated by user
+    if (!isActive) return null;
+
+    const { boundaries } = this.state;
```
jasontk19, line 244: Lots of vars being deconstructed from `this.state` here but in three separate statements.  Would be tidier to do it all with one statement.

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -0,0 +1,470 @@
...
+
+/**
+ * ReactRedux; used for selecting the part of the data from the store
+ * that the Smarthandoff component needs. This is called every time the
+ * store state changes.
+ * @param {*} state | Encapsulates the entire Redux store state.
+ * @param {*} ownProps | Data from SmartHandoff that is used to retrieve data from the store.
+ */
+const mapStateToProps = (state, ownProps) => {
+  const { tabTypes } = ownProps;
```
jasontk19, line 424: It doesn't look like you're using this prop so you can remove any references to it here, as well as remove the `ownProps` param entirely from your `mapStateToProps` definition.

web/js/containers/sidebar/sidebar.js
```diff
@@ -240,9 +240,8 @@ class Sidebar extends React.Component {
                   <TabPane tabId="download">
                     {dataDownloadFeatureActive
                       ? (
-                        <Data
+                        <SmartHandoff
                           isActive={activeTab === 'download'}
-                          height={subComponentHeight}
                           tabTypes={tabTypes}
```
jasontk19, line 245: Can remove `tabTypes` here as well.

web/js/containers/sidebar/smart-handoff-modal.js
```diff
@@ -0,0 +1,157 @@
...
+ * @param {*} dispatch | A function of the Redux store that is triggered upon a change of state.
+ */
+const mapDispatchToProps = (dispatch) => ({
+
+});
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps,
+)(SmartHandoffModal);
```
jasontk19, line 157: Since you're not actually hooking up any state to props, or dispatching anything, you can remove the `mapStateToProps` and `mapDispatchToProps` definitions here and also remove the `connect()` call to wire this up to Redux.

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -0,0 +1,470 @@
...
+   * When fired, compare prevProps to determine if previously selected layer is still active
+   * and whether or not to update granule data base don data changes.
+   * @param {*} prevProps
+   */
+  componentDidUpdate(prevProps) {
+    const { selectedDate, activeLayers } = this.props;
+    const { currentExtent, selectedLayer } = this.state;
+
+    // Determine if existing selected layer is active still and visibility toggle is 'ON'
+    const isLayerStillActive = activeLayers.find((layer) => selectedLayer.conceptId === layer.conceptId);
```
jasontk19, line 70: Totally optional but, a fun little trick to condense these one-liners is to deconstruct the single property you are using from the param, like so:
```suggestion
    const isLayerStillActive = activeLayers.find(({ conceptId }) => selectedLayer.conceptId === conceptId);
```

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -0,0 +1,470 @@
...
+    // Determine if date changed; if so, fire update on granule count
+    const didDateChange = selectedDate !== prevProps.selectedDate;
+    if (didDateChange) this.updateGranuleCount(currentExtent);
+  }
+
+  /**
+   * Resets this component to it's default state
+   */
+  resetState() {
+    this.setState(this.baseState);
```
jasontk19, line 85: I'm not 100% sure but this might not work as intended since in the constructor you assign `this.state` to `this.baseState` I think they might just always be the same object?  I need to debug it and find out.

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -0,0 +1,470 @@
...
+
+    let urlSelectedGranules = urlTotalGranules;
+
+    // Gets the total amount of granules that the layer has
+    totalGranules = await fetch(urlTotalGranules, { timeout: 5000 })
+      .then(async(response) => {
+        const result = await response.json();
+        return result.feed.facets.children[0].children[0].children[0].count;
+      })
+      .catch((error) => 0);
```
jasontk19, line 206: It looks like you are mixing traditional promises that utiliize callbacks with the newer `async/await` approach.

The `async/await` approach would look something like:
```
const granuleResponse = await fetch(urlSelectedGranules, { timeout: 5000 });
const result = await granuleResponse.json();
totalGranules = result.feed.facets.children[0].children[0].children[0].count;
```

Same for the async code below.

jasontk19 (changes requested):
Overall looking good. I just did a quick pass at the code, still need to poke at the functionality some more.  Some other things I thought of:

- [x] Since this replaces data download, it would be nice to remove as much of that old code as possible.  I don't expect you to find every relevant line in every file but if we can remove the major files and any related Redux constants/actions/reducers that would be great.

- [ ] Tests! We definitely want to get some basic E2E tests in place for this since it totally replaces a feature that has a good bit of testing to support it.  Similarly, we will probably want to update the test suite with some steps so testers have a plan to follow while testing.

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -0,0 +1,470 @@
...
+
+    // Gets the total subset of granules that are within the defining bounding box
+    if (showBoundingBox) {
+      urlSelectedGranules += `&bounding_box=${currentExtent.southWest},${currentExtent.northEast}`;
+      selectedGranules = await fetch(urlSelectedGranules, { timeout: 5000 })
+        .then(async(response) => {
+          const result = await response.json();
+          return result.feed.facets.children[0].children[0].children[0].count;
```
jasontk19, line 214: Trying to access properties nested this deeply without some safety checks is dangerous.  If anything along the way happens to be undefined, things will break.  Since this is coming from an API response that is out of our control, it would be especially prudent to do some checks for undefined values.  The easiest way would be to just use `lodash.get()`.

Once imported, that it would look something like:
``` get(result, 'feed.facets.children[0].children[0].children[0].count') ```

File: web/js/containers/sidebar/smart-handoff.js [hunk 9]
```diff
@@ -0,0 +1,470 @@
...
+          </div>
+
+          <hr />
+
+          <div id="smart-handoff-layer-list">
+            {activeLayers.map((layer, index) => {
+              if (layer.conceptId) {
+                return (
+                  <div className="layer-item">
+                    <input
```
jasontk19, line 281: This `<input>` element needs to have a unique `key` attribute:
Screen Shot 2020-11-23 at 8 25 08 AM

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -0,0 +1,470 @@
...
+      selectedGranules = await fetch(urlSelectedGranules, { timeout: 5000 })
+        .then(async(response) => {
+          const result = await response.json();
+          return result.feed.facets.children[0].children[0].children[0].count;
+        })
+        .catch((error) => 0);
+    }
+
+    this.setState({ selectedGranules, totalGranules });
+    this.setState({ isSearchingForGranules: false });
```
jasontk19, line 220: These two `setState()` calls can be combined into one:
```suggestion
this.setState({
    selectedGranules,
    totalGranules,
    isSearchingForGranules: false,
});
```

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -0,0 +1,470 @@
...
+
+    const { boundaries } = this.state;
+    const {
+      x, y, x2, y2,
+    } = boundaries;
+
+    // Used to determine if the added smart-handoff modal should be shown
+    const { HIDE_EDS_WARNING } = safeLocalStorage.keys;
+    let showModal = safeLocalStorage.getItem(HIDE_EDS_WARNING);
+    showModal = true;
```
jasontk19, line 252: You are always setting `showModal = true` here which means if the user checks the box "do not show this modal again" it has no effect and the modal always shows.

web/js/components/util/image-crop.js
```diff
@@ -94,6 +96,7 @@ Crop.defaultProps = {
   height: 10,
   maxHeight: window.innerWidth,
   maxWidth: window.innerHeight,
+  keepSelection: false,
```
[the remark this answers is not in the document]
StephanJW: We don't always want it to be false, because in the case of smart-handoffs, we need to override and set it to true. This makes it so the bounding box doesn't disappear when a user targets something else in the UI. If I only place `keepSelection={false}`  when I instantiate the component within smart-handoffs.js, it doesn't recognize it. Thus, the reason for me adding it in. I added the default state to not change the behavior for gifs and snapshots.

web/js/containers/sidebar/smart-handoff.js
StephanJW: Cleaned up 🧹

web/css/smart-handoff.css
```diff
@@ -0,0 +1,274 @@
+/** Smart handoff side panel **/
+#smart-handoff-side-panel h1 {
```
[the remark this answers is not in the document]
StephanJW: Not a problem, I'll change them over.

web/js/containers/sidebar/smart-handoff.js
StephanJW: Appreciate the solution. Indeed this is a much safer approach.

web/js/containers/sidebar/smart-handoff.js
StephanJW: Ah yes, so originally I was leveraging the concept ID to each layer, but as I found out, concept IDs are not truly unique, multiple layer imagery can reference the same concept ID. I tried utilizing the index but the linter doesn't agree. Looks like I'll have to leverage the layer ID (i.e. AIRS_L2_Carbon_Monoxide_500hPa_Volume_Mixing_Ratio_Night).

web/js/containers/sidebar/smart-handoff.js
StephanJW: Good catch. Forgot to remove this. Had this in there when I was editing some of the CSS for the modal and wanted it visible.

web/js/containers/sidebar/smart-handoff.js
StephanJW: Fancy.

web/js/containers/sidebar/smart-handoff.js
StephanJW: Overkill, is what I did. Thanks for refining.

web/js/components/util/image-crop.js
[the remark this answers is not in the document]
jasontk19: Ah, yep.  I see what's happening now.  Looks good 👍

jasontk19:
Just remembered we have a URL param for data download to allow permalinks to a specific product download (e.g. https://worldview.earthdata.nasa.gov/?l=MODIS_Combined_Value_Added_AOD&download=MCDAODHD)

I don't know if we need to allow similar functionality in this PR specifically but, at the very least any old permalinks with that parameter should be handled gracefully.  This is probably worth a bit of a discussion at a dev meeting.

jasontk19:
I was removing layers and when I got down to only one layer in the sidebar, the app threw an error.  This can also be reproduced by just loading directly from a permalink that only includes a single layer: `http://localhost:3000/?t=2020-11-06-T00%3A00%3A00Z&l=MODIS_Terra_CorrectedReflectance_TrueColor`

Screen Shot 2020-12-17 at 9 30 06 AM

jasontk19:
A few final things:


- [ ] We still need to add Google tag manager events so we can track usage after release.  I would encourage we create a separate ticket for this and do it separately

- [ ] It feels kind of weird to be able to select a "target area" with no layer selected.  I think it might look better if we hide the checkbox when a layer to download is not selected:
Screen Shot 2020-12-21 at 9 45 28 AM

- [ ] There is a React warning being shown when you select a radio button:
Screen Shot 2020-12-21 at 9 43 39 AM

- [ ] Another React warning is being shown when a layer is added via the product picker (this may not be related to changes on this branch, need to check still):
Screen Shot 2020-12-21 at 9 47 04 AM

edplato:
Re:
> * [ ]  There is a React warning being shown when you select a radio button:
>   Screen Shot 2020-12-21 at 9 43 39 AM

Adding a var like `selectedId` appears to address this warning:
```
// Determine if download 'smart-handoff' tab is activated by user
    if (!isActive) return null;
    const selectedId = selectedLayer && selectedLayer.id;
```

Add this `selectedId` outside of the radio button and then use to check if radio button is active:
`checked={selectedId === layer.id}`


@jasontk19  I was just looking into that additional sidebar warning. I think it's from this that needs to move to `componentDidMount`:
```
    const customPalettePromise = loadCustomPalette(props.config);
    customPalettePromise.done((customs) => {
      props.loadedCustomPalettes(customs);
    });
```

edplato (commented):
Looks and functions great! I added some code changes/cleanup kind of improvements and  a dev warning fix.

In addition to my code suggestions, I also noticed when selecting a target area selection in antarctic projection that there is a common console error:
`"The bounding rectangle north value [46.72333307942501] was less than the south value [54.04793318228071]"
]` [request link](https://cmr.earthdata.nasa.gov/search/granules.json?temporal=2020-12-20T00:00:00.000Z,2020-12-20T23:59:59.999Z&collection_concept_id=C1443536017-LAADS&include_facets=v2&page_size=0&day_night_flag=Day&bounding_box=-74.47982677345105,54.04793318228071,83.38117098066836,46.72333307942501).

I am also getting an Earthdata Search UI warning of `Error retrieving collections There was a problem completing the request
` warnings in antarctic after clicking the `Continue` button and opening a page to Earthdata Search. It may be worth investigating to see if these polar issues can be addressed on our end.

Finally, I'm not sure if it has been brought up before (or if it is even possible with query string parameters) but I wanted to float the idea of opening the Earthdata Search page in the coinciding polar projection when selecting data in a polar projection from Worldview.

edplato:
I pushed up some quick image optimizations in a commit also - hope that's ok.

jasontk19:
Ok I think I've addressed all of @edplato 's comments.  Going to EDS from Worldview in a polar projection will load EDS with that projection selected.  However, there is still an issue in polar projections with regards to how we are displaying the selection box vs how EDS interprets the bounding box we pass it.  (See comparison images below). There is not an easy fix for this since the cropper plugin we are using is expecting to be a rectangle and on the EDS side, the lines drawn between coordinates follow great circle arcs.  I think the most sensible solution to this for now is to just disable area selection in polar projections, which I have done.

I made a few other minor changes, e.g. now the granule count requests are a little more optimized and it shouldn't be re-requesting total granules unless the layer or date changes.  Previously any change to bounding box would re-request selected and total granules which was unnecessary.

We still need to update the test suite. I have stubbed out a page with some things we should be testing.  It will need to be fleshed out in the format of the other suites: https://wiki.earthdata.nasa.gov/display/WOR/S-14+%5Bsh%5D+-+Smart+Handoffs

I think this PR is good to go now and if @edplato and @StephanJW I think we can merge and do any additional work in follow on tickets.

Screen Shot 2020-12-30 at 9 15 41 AM
Screen Shot 2020-12-30 at 9 15 01 AM

jasontk19 (approved):
:shipit: 

FYI this should be merged with the "merge pull request" strategy rather than the "rebase and merge" strategy since the merge conflicts have already been resolved in a previous merge commit.  Doing a rebase at this point would just require fixing the same conflicts again.

edplato (approved):
Looks good to get into `develop`! What was decided on the future of the `download` permalink query string parameter? I see it is commented out in `web/js/location.js`.

jasontk19:
FYI @edplato  I will fix this issue within the "remove jQuery" issue I'm working now since I need to rework how these requests for palettes are being made.

> @jasontk19 I was just looking into that additional sidebar warning. I think it's from this that needs to move to `componentDidMount`:
> 
> ```
>     const customPalettePromise = loadCustomPalette(props.config);
>     customPalettePromise.done((customs) => {
>       props.loadedCustomPalettes(customs);
>     });
> ```

StephanJW:
@jasontk19, @edplato - Thank you guys for taking the time to review and optimize. Looks great.

Issue #3194, opened 2020-11-18
Remove use of mouse hover/move Redux actions 


jasontk19:
**What existing feature could be improved?**
There are a couple of Redux actions that we dispatch on mouseover/mouse move type events which cause a lot of unnecessary re-rendering.  The primary culprits are `MAP/RUNNING_DATA` and `LAYER/ON_LAYER_HOVER`.  We should revise the implementation of the surrounding code to avoid dispatching events for these actions.

Some possible approaches could be to use:
* React context providers
* Native JS events
* Some other pub/sub approach or library

Ideally whichever approach we go with, the involved components can communicate directly with each other so that the rest of the app can be unaware of these updates.

PR #3201, opened 2020-12-01
Layer grouping

### LINKED ISSUE #3185: Investigate grouping of related layers (e.g. all OSM layers in one "group")
**Is your feature request related to a problem? Please describe.**
When managing sets of related layers (e.g., Coastlines, Labels, Roads), a user may want to enable only a subset of those layers at any given time but toggle them on/off, as needed.  Our current solution treats each layer separately which takes up a good bit of real estate.  So ideally we'll want to make sure the user is aware of all of the options and can adjust them as needed BUT not take up an unnecessary amount of real estate.

As an upcoming use case, we plan to split all of our OSM layers out into individual layers and some kind of grouping should help:
* Coastlines
* Borders
* Labels
* Roads

An example of our existing layout:

**Describe the solution you'd like**
It would be helpful to group these similar layers together and manage them in one place.  One idea is a sort of "accordion" expansion of related layers. 

Another would look similar to our existing "classification" palettes to toggle on/off individual layers within a "parent" layer.

Another would work like our "add related orbit track" feature:

There of course may be other options, too!  So this ticket is to investigate design options for how we could build this functionality and if it applies more generally to other layers (e.g. our multitudes of fire and SST layers).

jasontk19:
## Description

Fixes #3185 

### What Changed

* add grouping for layers
* new permalink param added to indicate grouping enabled/disabled for each compare mode: `lg` and `lg1`
* `layergroup` property for all layer configs was converted to use the corresponding measurement for a layer and to be a single string value instead of an array
* a `tasks/generateLayergroup.js` config migration script is included to help migrate configs for other repos
* layer options/remove interactions now only show on hover
* some layer related selectors were simplified and/or memoized
* #3194 - revise event system to have all events go through a single instance. Update event names to be more descriptive.

NOTE: Some of these changes break parts of data download.  Since we are intending to deprecate data download in the same release that these changes are expected to be included, this shouldn't be an issue.  But, one data download e2e test appears to be failing and, if for some reason these changes are released without smart handoffs, any issues will need to be addressed.

### How to test

1. Open Worldview
2. Add some layers
3. Confirm layers get added to appropriate group
4. Re-order groups, confirm order is reflected on the map
5. Ungroup layers
6. Re-order layers arbitrarily
7. Confirm layer order is updated and reflected on the map
8. Turn grouping back on
9. Confirm layers are re-ordered into groups in the sidebar
10. Confirm layer order is updated and reflected on the map
11. Turn grouping off
12. Confirm arbitrary ordering is restored
13. Turn grouping back on
14. Remove any single layer
15. Turn grouping off
16. Confirm layer ordering now reflects ordering exhibited by grouped layers
17. Turn on compare mode
18. Turn grouping on for "A" side
19. Switch to "B" side
20. Confirm grouping is off
21. Try various combinations of adding/removing layers and groups on both "A" and "B" side and confirm both are properly maintained when switching between "A" and "B" 

* At various steps, try copying the URL and loading from the permalink in a new tab/window to confirm loading from a permalink works


web/js/containers/sidebar/layer-list.js
```diff
@@ -86,129 +99,148 @@ class LayerList extends React.Component {
...
   activeLayers: PropTypes.array,
   available: PropTypes.func,
-  checkerBoardPattern: PropTypes.object,
+  collapsed: PropTypes.bool,
+  compareState: PropTypes.string,
   getNames: PropTypes.func,
   groupId: PropTypes.string,
-  layerGroupName: PropTypes.string,
+  isMobile: PropTypes.bool,
+  launchOpacityModal: PropTypes.func,
```
edplato, line 202: These don't appear to be being used: 
`launchOpacityModal`
`runningLayers`

I feel we had an eslint rule at some point for this, but I've run into accidentally pushing up unused propTypes recently.

web/js/containers/sidebar/layers-container.js
```diff
@@ -0,0 +1,201 @@
...
+      .flatMap(({ layers }) => layers)
+      .map((id) => activeLayersMap[id])
+      .concat(baselayers);
+    reorderOverlayGroups(newLayers, newGroups);
+  };
+
+  const renderLayerList = (group, idx) => {
+    const { groupName, layers, collapsed } = group;
+    const layersForGroup = layers.map((id) => activeLayersMap[id]);
+    return layers && ((
```
edplato, line 58: `return layers && ((` to `return layers && (` and remove closing paren.

web/js/containers/sidebar/layers-container.js
```diff
@@ -0,0 +1,201 @@
...
+  const { baselayers, overlays } = getAllActiveOverlaysBaselayers(state);
+  const overlayGroups = groupOverlays ? getActiveOverlayGroups(state) : [];
+
+  return {
+    baselayers,
+    overlays,
+    overlayGroups,
+    groupOverlays,
+    activeLayersMap: getActiveLayersMap(state),
+    layerSplit: overlays.length,
```
edplato, line 168: `layerSplit: overlays.length,`
This doesn’t appear to be used and/or logic is being repeated elsewhere (which may be necessary)

web/js/map/rotation.js
```diff
@@ -4,7 +4,7 @@ import util from '../util/util';
 import { faIconUndoSVGDomEl, faIconRedoSVGDomEl } from './fa-map-icons';
 
 export default function MapRotate(ui, models, store) {
-  this.evts = util.events();
+  this.evts = util.events;
```
edplato, line 7:
```suggestion
```
This doesn't appear to be used.

web/js/map/ui.js
```diff
@@ -579,22 +563,15 @@ export default function mapui(models, config, store, ui) {
...
-      state,
-    ).reverse();
+    const activeLayers = getAllActiveLayers(state);
     let layerGroups;
     let layerGroup;
     if (compare && compare.active) {
       layerGroups = self.selected.getLayers().getArray();
       if (layerGroups.length === 2) {
-        layerGroup = layerGroups[0].get('group') === activeLayerStr
+        layerGroup = layerGroups[0].get('group') === compare.activeString
```
edplato, line 572: `activeLayerStr` use was getting confusing - good clarity change!

web/js/modules/layers/selectors.js
```diff
@@ -202,27 +246,16 @@ export function getLayers(layers, spec = {}, state) {
...
-    }
-    if (
-      spec.renderable
-      && !isRenderable(def.id, activeLayers, spec.date, state)
-    ) {
-      return;
-    }
-    if (spec.visible && !def.visible) {
+    const notInProj = !def.projections[projId];
+    const dynamic = spec.dynamic
```
edplato, line 254: I know this wasn't really your PR specific, but I believe this `spec.dynamic` isn't being used and can be removed.

edplato:
Looks and works great. I made some real minor code suggestions. The only potential issue I came across in testing was for the `lg` permalink. The default case of `true` was being retained on refresh - not sure if this was necessary or not since other default permalink params are omitted. Here's the process to reproduce:

- default page
- turn off group and `lg=false` in permalink
- refresh
- turn on group and `lg=true` is retained

web/js/modules/layers/selectors.js
jasontk19: Nice, I'm all for cleanup if it's near my changes!

web/js/containers/sidebar/layer-list.js
jasontk19: Hmm yeah I thought we did too but maybe I'm confusing it with what we do have which is the inverse... it yells at you if you have props without corresponding PropTypes 🤔   I think it's worth adding an eslint rule if there is one.

web/js/containers/sidebar/layers-container.js
jasontk19: Yup, I think I meant to just pass it to children as `overlays.length` to reduce props in this file then forgot to remove it.  Nice catch!

web/js/map/rotation.js
jasontk19: Nice catch!  I was worried it was being referenced on instances of this module but never went back and checked.  Looks like it's not.

web/js/map/ui.js
jasontk19: Thanks! Was confusing me as well.

jasontk19:
> Looks and works great. I made some real minor code suggestions. The only potential issue I came across in testing was for the `lg` permalink. The default case of `true` was being retained on refresh - not sure if this was necessary or not since other default permalink params are omitted. Here's the process to reproduce:
> 
> * default page
> * turn off group and `lg=false` in permalink
> * refresh
> * turn on group and `lg=true` is retained

Dealing with this param got kind of tricky since I was trying to meet the following criteria:
* Arriving via permalink with the `lg` param omitted indicates an "old" permalink so `lg` becomes `false` and is inserted into the URL on load
* By default, groups are turned on but the `lg` param is not shown in the URL if no layer params are present and the checkbox hasn't been toggled.

It needs to be present if you've manually disabled groups but I'm also not sure it hurts anything that it becomes present when you load with `lg=false` then turn on groups?  I'll take a look and see if it's easily removable in these cases without breaking anything.

web/js/modules/layers/selectors.js
```diff
@@ -356,7 +387,7 @@ export function available(id, date, layers, parameters) {
  * @param {*} id - the layer id
  */
 export const memoizedAvailable = createSelector(
-  [getCurrentDate, getCurrentActiveLayers, getConfigParameters],
+  [getSelectedDate, getActiveLayers, getConfigParameters],
```
Benjaki2, line 390: 👍

Benjaki2 (approved):
💯 This is really cool! As you can see... Lots of thumbs up... I like the updates to the events and the use of react hooks.

Couple potential changes:

1. The red checkbox button seems off to me. Maybe it should be a different color so that it doesn't take away from importance of the layers button?Screen Shot 2020-12-16 at 4 15 58 PM
2. The layers should have a draggable pointer as well (where draggable)

web/js/map/ui.js
```diff
@@ -958,7 +930,7 @@ export default function mapui(models, config, store, ui) {
     };
     map.getView().on('change:resolution', () => {
       onZoomChange();
-      self.events.trigger('movestart');
+      events.trigger('movestart');
```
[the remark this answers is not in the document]
jasontk19: ah, yes... I think this got changed back when this branch got merged with Ed's recent changes... good catch!

minniewong (commented):
Looks really nice!
I noticed an issue when changing projections - I loaded 2 brightness temperature (GPM/GMI and Terra/MODIS) layers in the Geographic projection. The layer grouping indicates there are 2 layers. When i change to Arctic, the Brightness Temperature layer grouping still says that there are 2 layers, but in fact, only one of them displays in the Arctic projection.

jasontk19:
@edplato I believe the URL param is working as intended now, let me know what you think.

@minniewong Good catch, the count should now be accurately based on how many layers are available in the current projection within a group.

minniewong (approved):
Looks really good!

edplato (approved):
Looks great! 👍🏽

Issue #3209, opened 2020-12-09
Sources Facet with copyright sign does not display correctly


minniewong:
The copyright sign does not display correctly in the sources facet.
![Screen Shot 2020-12-09 at 12.59.42 PM.png](https://images.zenhubusercontent.com/5d434327dd8f953c8f021bfa/43fad381-8cc4-453b-ac41-469df5f47a82)

PR #3214, opened 2020-12-11
Fix undefined fields and html entities in field titles

### LINKED ISSUE #3210: undefined in sources facet doesn't display anything
There is an "undefined" row in the sources facet that lists 4 results, but 4 results do not display.

Screen Shot 2020-12-09 at 1 00 36 PM

jasontk19:
## Description

Fixes #3210, #3209.

### DIFF
--- a/web/css/facets.css
+++ b/web/css/facets.css
@@ -9,14 +9,16 @@
     padding: 3px;
   }
   & .filter-chip {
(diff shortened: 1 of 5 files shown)


jasontk19:
@minniewong It was the configs where the `subtitle` field isn't defined and is instead defined per-projection (e.g. Coastlines)

PR #3217, opened 2020-12-14
Remove unused ownProps

### LINKED ISSUE #3192: Remove unused references to ownProps
[Including the `ownProps` param in a `mapStateToProps` function definition will cause `mapStateToProps` to be called any time `ownProps` changes.](https://react-redux.js.org/using-react-redux/connect-mapstate#the-number-of-declared-arguments-affects-behavior)  If `ownProps` is not actually being used, this can cause unnecessary calls to `mapStateToProps`.  We should remove `ownProps` from any definitions where it is not needed. 

**Bonus**
Consider implementing memoized selectors with Reselect for any expensive operations being performed within `mapStateToProps` while doing this work for additional performance gains.

jasontk19:
## Description

Fixes #3192

- remove `ownProps` param from `mapStateToProps` where not needed
- remove properties being returned from `mapStateToProps` that are already on props

### DIFF
--- a/web/js/app.js
+++ b/web/js/app.js
@@ -207,7 +207,7 @@ class App extends React.Component {
   }
 }
 
-function mapStateToProps(state, ownProps) {
+function mapStateToProps(state) {
   return {
     state,
     isAnimationWidgetActive: state.animation.isActive,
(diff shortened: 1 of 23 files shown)


edplato (approved):
Looks good to me 👍🏽

PR #3257, opened 2021-01-04
Make sure tooltips stay open when mousing over them

jasontk19:
## Description

Fixes # 3229.


jasontk19:
@minniewong Figured out the fix... now no flickering and the layer notice tooltips stay open while you're pointing at them.

minniewong (approved):
👍  Looks great!

PR #3261, opened 2021-01-05
Remove jquery

jasontk19:
## Description

Fixes # 3161.

* Remove jQuery and jQuery-UI, use React and native JS instead
* Create Rotation and Zoom components for map interaction
* Replace jQuery AJAX calls with native promises and fetch
* Remove some unused, old code
* Remove `indicator.js` since barely used and ineffective as implemented
* Remove map icons SVG strings file, add to font awesome library and use via FA React component

We will want to make sure we test the feedback feature on UAT (I believe its part of the test suite) since we cannot test it locally.


edplato (approved):
🚀 It served this app well, but it's been a long time coming to remove it. Looks great!

Benjaki2 (approved):
Long time coming! Nice work!

PR #3280, opened 2021-01-13
fix image paths in smart handoffs modal

jasontk19:
## Description

These image paths needed to be fixed so that these images would work in deployed environments.


edplato (approved):
👍🏽

PR #3283, opened 2021-01-20
Build categoryGroupOrder if missing during config

### LINKED ISSUE #3282: Allow flexible categories naming for product picker layers
There are currently some expectations that product picker category names will remain our default names (e.g., 'hazards and disasters') that may cause issues with user custom configurations that add/remove category names or use their own unique category naming. This causes a bug where the product picker won't open when clicking the Add Layers button.

**Potential solution(s)**
There may need to be some parsing of categories from the config files during build stage that are then added to the relevant `web/js` files and/or some `initialState` handling that relies on the config object.

edplato:
## Description

Fixes #3282  .

- [x] Create `categoryGroupOrder.json` file from categories directories if missing during config build

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/tasks/buildOptions.sh
+++ b/tasks/buildOptions.sh
@@ -124,7 +124,9 @@ fi
 
 # Throw error if no categoryGroupOrder.json file present
 if [ ! -e "$BUILD_DIR/config/wv.json/categoryGroupOrder.json" ] ; then
-    die "config/wv.json/categoryGroupOrder.json does not exist"
+    echo "categoryGroupOrder.json not found.  Generating..."
+    "$PYTHON_SCRIPTS_DIR/generateCategoryGroupOrder.py" "$SRC_DIR/common/config/wv.json/categories/" \
+        "$SRC_DIR/common/config/wv.json/"
 fi
 
 # Run mergeConfig.py on all directories in /config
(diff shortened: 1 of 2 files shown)


web/js/main.js
```diff
@@ -198,6 +198,12 @@ window.onload = () => {
           }
         });
       }
+      // handle legacy config categories
+      if (!config.categoryGroupOrder) {
```
jasontk19, line 202: I think this is fine for this release but, there are lots of other config files that need to be present that we don't/wouldn't throw errors for in `main.js` if they were not present so this feels a little out of place.  I wonder if maybe a warning/error should be part of the config building process during which, if necessary, this file could actually be generated based on the categories found in the category configs.

jasontk19 (approved):

web/js/main.js
edplato: Ok, I opted to throw an error during the build process. The building of a json file using the category directories sounds like something interesting I'd like to see in a follow up ticket if you have an idea to handle that.

jasontk19: I will take a quick stab at it this afternoon and see if I can include it in this PR.  If not, this is a good solution for now.

jasontk19: Ok, I pushed a little script that will generate the `categoryGroupOrder.json` if it's not present.  Let me know what you think.

edplato: Nice, I think that will work!

jasontk19 (approved):

PR #3292, opened 2021-01-26
JSON Schema layer validation

### LINKED ISSUE #3291: JSON Schema validation for layer configs
We should validate our layer configs at build time with a JSON schema

jasontk19:
## Description

Fixes #3291 

* Create a JSON schema with which to validate our layer configs
* Set up layer validation against schema to happen during build and output errors if validation fails
* Clean up some config properties based on findings while building out the schema
* Update layer config docs
* Add `layergroup` property to new HLS layers


Benjaki2 (approved):
Nice Work!

edplato (approved):
This is a big config improvement 🚀

PR #3304, opened 2021-02-01
Fix console warning

### LINKED ISSUE #3293: Console warning when adding layers in product picker
**Describe the bug**
``` 
Warning: Cannot update a component from inside the function body of a different component.
    in Draggable (created by ConnectFunction)
    in ConnectFunction (created by PrivateDraggable)
    in PrivateDraggable (created by PublicDraggable)
    in PublicDraggable (created by LayerRow)
    in LayerRow (created by ConnectFunction)
...
```

**To Reproduce**
1. Open product picker
2. Add any layer
3. See warning in console

**Additional context**

jasontk19:
## Description

Fixes #3293

* Only happened on layer with palettes
* Update to not call `dispatch` outside of a `useEffect`

@nasa-gibs/worldview

### DIFF
--- a/web/js/containers/sidebar/layer-row.js
+++ b/web/js/containers/sidebar/layer-row.js
@@ -1,7 +1,7 @@
 /* eslint-disable react/jsx-props-no-spreading */
 /* eslint-disable react/no-danger */
 /* eslint-disable no-nested-ternary */
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Draggable } from 'react-beautiful-dnd';
 import { isEmpty as lodashIsEmpty, get as lodashGet } from 'lodash';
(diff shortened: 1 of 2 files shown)


web/js/containers/sidebar/layer-row.js
```diff
@@ -96,11 +96,13 @@ function LayerRow (props) {
...
         />
       );
-    } if (!isLoading) {
-      requestPalette(layer.id);
     }
   };
 
+  useEffect(() => {
+    if (!isLoading && layer && hasPalette) requestPalette(layer.id);
+  }, [layer]);
```
Benjaki2, line 104: Not positive, but maybe you want to use `[layer.id])`. Uncertain how that object would be compared in this case.

Benjaki2 (approved):
👍 Looks like Travis is breaking, but that is probably unrelated.

web/js/containers/sidebar/layer-row.js
jasontk19: I _think_ it's an object reference comparison and since the `layer` prop shouldn't be changing on this component I think it will have the same effect if we use `layer.id`

jasontk19: Actually, after some debugging on this it looks like this runs fewer times when we check for `layer.id` instead of `layer` so that seems like the better approach.  Also it looks like a few other places in the app are throwing this warning so I want to see if I can track those down and fix them in the same PR.

PR #3330, opened 2021-02-08
Event re-focus when switching projections

### LINKED ISSUE #3326: events.notfocus.4 - When switching projections from previously selected event, it is no longer at the top of the list
Event focus carryover - I believe previously if you had selected an event in Geographic, changed projections and changed back to geographic, it would display at the top of the Events list but it would not be highlighted.

jasontk19:
## Description

Fixes #3326

When an event is selected, if the user switches to a projection that would not show the event we store it in a `prevSelected` state property while clearing the `selected` state property.  Then when switching projections again we check to see if it would show now, and if it does we re-select it.


edplato (approved):
Looks good 👍🏽

PR #3335, opened 2021-02-09
No Smart Handoff dates for non-temporal layers

### LINKED ISSUE #3334: Smart handoff URL should not include date range for non-temporal layers
**Describe the bug**
Layers without a `dateRanges` property are not date specific and therefore shouldn't include the app's current date in the URL params when handing off to Earthdata Search since doing so would restrict the number of granules shown or prevent any from being shown entirely.

We also need to exclude the date from the granule count requests.

jasontk19:
## Description

Fixes #3334

### DIFF
--- a/web/js/components/smart-handoffs/smart-handoff-modal.js
+++ b/web/js/components/smart-handoffs/smart-handoff-modal.js
@@ -24,6 +24,7 @@ function SmartHandoffModal({
   const {
     value, title, type, version,
   } = selectedCollection;
+  const { dateRanges } = selectedLayer;
   const cmrSearchDetailURL = `https://cmr.earthdata.nasa.gov/search/concepts/${value}.html`;
 
   return (
@@ -96,7 +97,6 @@ function SmartHandoffModal({
         </div>
 
         <h1> Layer: </h1>
-
         <div className="handoff-modal-layer-title">
           {selectedLayer.title}
         </div>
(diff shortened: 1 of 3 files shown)


minniewong (approved):
Looks pretty good to me. A bonus would be to address the next level of complication where some are sort of temporal e.g. the Population Density layers are split by year 2000, 2005, 2010 etc. and you could pass through a temporal filter for 2010-01-01 to 2010-12-31, but I don't think it's totally necessary.

jasontk19:
> Looks pretty good to me. A bonus would be to address the next level of complication where some are sort of temporal e.g. the Population Density layers are split by year 2000, 2005, 2010 etc. and you could pass through a temporal filter for 2010-01-01 to 2010-12-31, but I don't think it's totally necessary.

Hmmmm... yeah we could potentially solve this one of three ways:

* Custom WV config property
* Additional vis metadata property
* GIBS date range metatdata in GetCapabilities

Checking one of those layers it looks like the granules do have a date associated with them which is July 1st of the corresponding year.

minniewong:
Don't think it has to be solved today but we can keep it in mind. Layers include: Estimated Net Migration by Decade: 1970-1980, Estimated Net Migration by Decade: 1980-1990, Estimated Net Migration by Decade: 1990-2000, Population Density 2000, 2005, 2010, 2015, 2020, Particulate Matter <2.5um, 2001 - 2010, Particulate Matter <2.5um, 2010 - 2012, Ground Level NO2, 1996 - 1998, Ground Level NO2 2010-2012.

edplato (approved):
Looks good 👍🏽

PR #3349, opened 2021-02-11
Stop animation on projection switch

### LINKED ISSUE #3347: Animation does not stop playing when you change the projection
**Describe the bug**
date.animation.14 on the timeline test [page](https://wiki.earthdata.nasa.gov/display/WOR/S-06+%5Bdate%5D%3A+Timeline) has an Expected Result of 
"The animation should stop
Pause button should reset to the play button"

this does not happen.  If I start the animation and switch to any projection, it continues to play.
Mac/Chrome Version 88.0.4324.150

jasontk19:
## Description

Fixes #3347 

Also, removes some lodash dependency where ES6 will suffice

### DIFF
--- a/web/js/containers/projection.js
+++ b/web/js/containers/projection.js
@@ -51,11 +51,11 @@ class ProjectionList extends Component {
 
   onClick(id) {
     const {
-      updateProjection, projection, config, onCloseModal,
+      updateProjection, projection, config, onCloseModal, isPlaying,
     } = this.props;
 
     if (id !== projection) {
-      updateProjection(id, config);
+      updateProjection(id, config, isPlaying);
     }
 
     googleTagManager.pushEvent({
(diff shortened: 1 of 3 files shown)


web/js/containers/projection.js
```diff
@@ -94,6 +95,7 @@ const mapDispatchToProps = (dispatch) => ({
   updateProjection: (id, config) => {
     dispatch(changeProjection(id));
     dispatch(resetProductPickerState(id));
+    dispatch(stop());
```
Benjaki2, line 98: Nit pick... But I would lean towards only firing the dispatcher only if animating.

Benjaki2 (approved):

web/js/containers/projection.js
jasontk19: Good call!

PR #3352, opened 2021-02-12
hide layers without snapshot #3343

### LINKED ISSUE #3343: Trying to Download image with HLS grid active returns an error
**Describe the bug**
Trying to Download image with HLS grid active returns an error

**To Reproduce**
1. Go to '?v=-242.16209711079125,-39.60457890548289,4.255336084090402,79.70252833303897&as=2016-09-21-T00%3A00%3A00Z&ae=2016-09-28-T00%3A00%3A00Z&l=HLS_MGRS_Granule_Grid,Reference_Labels(hidden),Reference_Features(hidden),Coastlines,VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor(hidden)&lg=false&al=true&ab=on&t=2016-09-26-T00%3A00%3A00Z'
2. Try to download image
3. See error

**Expected behavior**

**Screenshots**

 - OSX
 - Browser: Chrome

**Additional context**
I don't believe there is a WMS layer for the HLS grid. Should we force the user to disable this layer, similarly to graticules?

Benjaki2:
## Description

Fixes #3343

Ability to disable layers that don't work with WVS

- [x] Add hide layer capability 
- [x] Add snapshot to layer schema and docs

## Further comments

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/doc/config/layers.md
+++ b/doc/config/layers.md
@@ -105,7 +105,7 @@ Example:
 * **track**: Classify a layer's track direction. This will usually be either ascending or descending and will be displayed within the title of the layer in the Measurement / Sources add modal view if provided.
 * **description**: Point to a markdown file within the metadata folder to provide a layer description.
 * **wrapX**: Wrap the layer across the anti-meridian.
-* **snapshot**: Ability to be downloaded via worldview snapshots (WVS). Default is `true`.
+* **disableSnapshot**: Disable downloaded via worldview snapshots (WVS).
 * **wrapadjacentdays**: Wrap the layer across the anti-meridian but select the previous day when greater than 180 and the next day when less than -180.
 * **palette**: To display a color palette legend, a `palette` object should exist with the following properties:
   * **id**: Identifier of the palette. This should match the name of the colormap file without the extension.
--- a/doc/config/layers.md
+++ b/doc/config/layers.md
@@ -105,6 +105,7 @@ Example:
 * **track**: Classify a layer's track direction. This will usually be either ascending or descending and will be displayed within the title of the layer in the Measurement / Sources add modal view if provided.
 * **description**: Point to a markdown file within the metadata folder to provide a layer description.
 * **wrapX**: Wrap the layer across the anti-meridian.
+* **snapshot**: Ability to be downloaded via worldview snapshots (WVS). Default is `true`.
 * **wrapadjacentdays**: Wrap the layer across the anti-meridian but select the previous day when greater than 180 and the next day when less than -180.
 * **palette**: To display a color palette legend, a `palette` object should exist with the following properties:
   * **id**: Identifier of the palette. This should match the name of the colormap file without the extension.
--- a/doc/config/layers.md
+++ b/doc/config/layers.md
@@ -105,7 +105,7 @@ Example:
 * **track**: Classify a layer's track direction. This will usually be either ascending or descending and will be displayed within the title of the layer in the Measurement / Sources add modal view if provided.
 * **description**: Point to a markdown file within the metadata folder to provide a layer description.
 * **wrapX**: Wrap the layer across the anti-meridian.
-* **disableSnapshot**: Disable downloaded via worldview snapshots (WVS).
+* **disableSnapshot**: Disable worldview snapshots(WVS) for layer.
 * **wrapadjacentdays**: Wrap the layer across the anti-meridian but select the previous day when greater than 180 and the next day when less than -180.
 * **palette**: To display a color palette legend, a `palette` object should exist with the following properties:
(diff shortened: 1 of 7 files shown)


web/js/modules/image-download/util.js
```diff
@@ -287,3 +289,53 @@ export function getPercentageFromPixel(maxDimension, dimension) {
...
+ *
+ * @return {String}
+ */
+export function getNonDownloadableLayerWarning(nonDownloadableLayer) {
+  const layerStr = getNamesOfNondownloadableLayers(nonDownloadableLayer);
+  if (!layerStr) return '';
+  const multiLayers = layerStr.indexOf(',') > -1;
+  return multiLayers ? `The ${layerStr} layers cannot be used to take a snapshot. Would you `
+  + 'like to temporarily hide these layers?' : `The ${layerStr} layer cannot be used to take a snapshot. Would you `
+  + 'like to temporarily hide this layer?';
```
jasontk19, line 331:
```suggestion
  const layerPluralStr = multiLayers ? 'layers' : 'layer';
  return `The ${layerStr} ${layerPluralStr} cannot be included in a snapshot. Would you like to temporarily hide this layer?`;
```

jasontk19: Suggested a tweak to the verbiage here to "cannot be included in a snapshot" rather than "cannot be used to take a snapshot"

web/js/modules/image-download/util.js
```diff
@@ -287,3 +289,53 @@ export function getPercentageFromPixel(maxDimension, dimension) {
...
 }
+
+/**
+ * Find if there are layers that cannot be downloaded
+ * @param {Array} visibleLayers
+ *
+ * @return {Bool}
+ */
+export function hasNonDownloadableVisibleLayer(visibleLayers) {
+  const notDownloadable = !!lodashFind(visibleLayers, { snapshot: false });
```
jasontk19, line 300: Not really a big deal either way but, why use lodash when we can do this natively with ES6?  I don't think we need to phase out lodash entirely since it does have some handy methods that don't exist natively and the file size savings would only be around 25kb gzipped but, I've been trying to replace things like this when I see them since I've got a slight preference for the native approach.  For example, using the spread operator in a reducer instead of lodash assign just looks cleaner to me.

Doing a find would be just a simple one liner in ES6 as well but, I think `Array.some()` is even more appropriate here since it returns a boolean:
```
export function hasNonDownloadableVisibleLayer(visibleLayers) {
  return visibleLayers.some(({ snapshot = true }) => !snapshot);
}
```

doc/config/layers.md
```diff
@@ -105,6 +105,7 @@ Example:
 * **track**: Classify a layer's track direction. This will usually be either ascending or descending and will be displayed within the title of the layer in the Measurement / Sources add modal view if provided.
 * **description**: Point to a markdown file within the metadata folder to provide a layer description.
 * **wrapX**: Wrap the layer across the anti-meridian.
+* **snapshot**: Ability to be downloaded via worldview snapshots (WVS). Default is `true`.
```
jasontk19, line 108: I think adding a new property to configs where the default value is considered `true` is a bit awkward in JavaScript since an `undefined` value is considered `false`.  It might be preferable to to name the property something like `disableSnapshot` so the default value can be `false`.  This allows a check for matching layers to look like:
```
visibleLayers.filter(({ disableSnapshot }) => !disableSnapshot);
```
instead of the following which hurts my head a little bit to look at:
```
visibleLayers.filter(({ snapshot = true }) => !snapshot);
```

web/js/modules/image-download/util.js
```diff
@@ -287,3 +289,53 @@ export function getPercentageFromPixel(maxDimension, dimension) {
...
+  + 'like to temporarily hide this layer?';
+}
+/**
+ * Get array of layers that will be removed if nofication is accepted
+ * @param {Array} visibleLayers
+ *
+ * @return {Array}
+ */
+export function getNonDownloadableLayers(visibleLayers) {
+  return lodashFilter(visibleLayers, { snapshot: false });
```
jasontk19, line 340: Same here with ES6 v lodash.  Could be:
```
export function getNonDownloadableLayers(visibleLayers) {
  return visibleLayers.filter(({ snapshot = true }) => !snapshot);
}
```

jasontk19 (changes requested):
Functionality looks good!  Just a few minor worth considering.

edplato (approved):
Works great 👍🏽 @jasontk19 hit on some good changes related to the ES6 refactor vs `lodash` where reasonable

jasontk19 (approved):
💯

PR #3356, opened 2021-02-16
Fix location/measure tooltip interactions on tablet

### LINKED ISSUE #3336: Geosearch marker/measurement tooltips hard to close (iPad)
**Describe the bug**
When trying to interact with the "minimize" or "close" icons in the Geosearch marker's tooltip on a touch device the first touch will show the tooltip, and several more touches will eventually close the tooltip.

Similarly, the "close" icon in the measurement tooltip does not seem to respond to touch events at all and so measurements cannot be removed individually.

**Expected behavior**
On a touch device there should be no tooltip interaction for these as the expected behavior is obvious to most users. A single touch interaction should minimize/close.

 - Device: iPad AIr 2
 - OS: iOS
 - Browser: Safari

jasontk19:
## Description

Fixes #3336.

This can be tested on a device by going to GIBS UAT `/worldview-dev/touch-events/`

* Replace `react-copy-to-clipboard` with `copy-to-clipboard` so the copy action can be fired `onTouchEnd`
* Apply `onTouchEnd` handler to minimize and close buttons as well so first touch interaction does not open the tooltip but instead executes the action


web/js/components/location-search/coordinates-dialog.js
```diff
@@ -14,14 +14,23 @@ class CoordinatesDialog extends Component {
...
+    const options = window.clipboardData ? {} : { format: 'text/plain' };
+    options.onCopy = () => {
+      this.setState({
+        tooltipToggleTime: Date.now(),
+        isCopyToClipboardTooltipVisible: true,
+      });
+
+      // Prevnet keyboard overlay in iOS
+      setTimeout(() => {
+        document.getElementById('geosearch-autocomplete').blur();
```
edplato, line 30:
```suggestion
        document.getElementById('location-search-autocomplete').blur();
```
Here is the updated id name

edplato (changes requested):
Looks good 👍🏽 Tested fix also on BrowserStack iPad.

I added updated id name as a change

web/js/components/location-search/coordinates-dialog.js
jasontk19: Ah, good catch.  I guess I wasn't up to date with develop

edplato (approved):

PR #3364, opened 2021-02-18
Restore event state on sidebar collapse/expand

### LINKED ISSUE #3361: In mobile view, event list does not show selected event after selection
On mobile/simulated mobile view, if you select an event lower down on the list, the map will automatically close the events tab and zoom to the event. If you go back to the events tab, the event is not in view on the list. The list should move to the selected event.

Firefox/Chrome

jasontk19:
## Description

Fixes #3361 

- Move category selection to redux to persist it between sidebar collapse/expand
- Make sure selected event is scrolled into view when sidebar expanded after collapse


minniewong (changes requested):
I noticed that if you do the following, it doesn't show the selected event in the Event list:
1 - select an event low down on the list
2 -  change tabs to Layers or Data
3 - Collapse the Layer List
4 - Uncollapse the Layer List
5 - Click back on the event tab, the list doesn't scroll to the selected event even though it is still selected on the map.

jasontk19:
@minniewong Good catch!  It looks like there was also a problem if you switched to the layers tab then back to events.  Now it seems to work in all scenarios.  Let me know what you think.

minniewong (approved):
Working great now! Thanks!

PR #3366, opened 2021-02-18
Disabled zoom button hover styles

jasontk19:
## Description

Make disabled zoom in/out buttons look like they do in prod when hovered


Benjaki2 (approved):
👍

PR #3374, opened 2021-02-19
Handle undefined conceptids

### LINKED ISSUE #3373: Smart handoffs should not show entries for layers with invalid concept ids
**Describe the bug**
If a concept id is invalid when Worldview is built, smart handoffs looks kind of broken when those layers are selected (listing `undefined` for some values)

**Expected behavior**
We should filter out any layers that did not have valid concept ids when the build occurred to prevent showing these unfriendly values.

**Screenshots**
Screen Shot 2021-02-19 at 10 29 46 AM

jasontk19:
## Description

Fixes #3373 

This branch is based off Minnie's AMSRE layers branch since those layers were the example for which this issue was noticed.

These changes will filter out any layers from smart handoffs for which we did not receive data from CMR during the build process.   Ideally, in the future we might want to be making requests to CMR in the UI while the user is looking at layers to have the most updated possible info on whether or not the collection exists in CMR.  However, that felt like too invasive of an approach this close to the release which would risk introducing other bugs.

If concept ids get updated/added/removed in CMR after we release, a simple rebuild and deploy should address the issue.  An entire release would not be needed.

### DIFF
--- a/config/default/common/features.json
+++ b/config/default/common/features.json
@@ -9,7 +9,7 @@
         "feedback": true,
         "previewSnapshots": true,
         "vismetadata": {
-            "url": "https://uat.gibs.earthdata.nasa.gov/layer-metadata/v1.0/"
+            "url": "https://gibs.earthdata.nasa.gov/layer-metadata/v1.0/"
         },
         "googleTagManager": true,
         "notification": {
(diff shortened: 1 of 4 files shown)


minniewong (approved):
makes sense to remove them from data download. Thanks!

rboller (approved):
I fully agree with the approach and the idea that we might do this dynamically in the future.  Thanks, @jasontk19 !

jasontk19:
Great, thanks for taking a look!  I've removed Minnie's commits from this branch and want to note that this PR also includes the text change to "Area of interest" for the checkbox as well as updates the GIBS vis metadata URL to prod.

PR #3378, opened 2021-02-22
Handle splitting orbit track id into two separate layer ids for image download requests

edplato:
## Description

Changes in GC and related WVS MapServer Mapfile configurations built from it now split each Orbit Tracks layer into individual 'Lines' and 'Points' layers. To request an image from WVS for a snapshot or GIF, the Orbit Tracks layer id needs to be replaced into its two separate layer ids as follows:

`OrbitTracks_Aqua_Ascending` in WV is revised in order in its image request to WVS as:
`OrbitTracks_Aqua_Ascending_Points` and `OrbitTracks_Aqua_Ascending_Lines`

In addition, any wrap and opacity values within the query string request are repeated in order.

## Further comments

This is intended to be tested against the current WVS SIT v1.4.5.

## Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/modules/image-download/util.js
+++ b/web/js/modules/image-download/util.js
@@ -31,6 +31,48 @@ export function getLatestIntervalTime(layerDefs, dateTime) {
 }
 
 /**
+ * Process original orbit track layers to split into two separate layers and repeat
+ * wrap and opacity values for original
+ * @param {Array} layersArray
+ * @param {Array} layerWraps
+ * @param {Array} opacities
+ */
+const addOrbitTrackRevisions = function(layersArray, layerWraps, opacities) {
+  const revisedLayersArray = [...layersArray];
+  const revisedLayerWraps = [...layerWraps];
+  const revisedOpacities = [...opacities];
+  let mod = 0;
+
+  // check for OrbitTracks in layersArray
+  for (let i = 0; i < layersArray.length; i += 1) {
+    const layerId = layersArray[i];
+    if (layerId.includes('OrbitTracks')) {
+      // track index for modifications from splicing
+      const idx = i + mod;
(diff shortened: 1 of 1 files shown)


web/js/modules/image-download/util.js
```diff
@@ -43,9 +85,23 @@ export function getLatestIntervalTime(layerDefs, dateTime) {
  */
 export function getDownloadUrl(url, proj, layerDefs, lonlats, dimensions, dateTime, fileType, isWorldfile, markerCoordinates) {
   const { crs } = proj.selected;
-  const layersArray = imageUtilGetLayers(layerDefs, proj.id);
-  const layerWraps = imageUtilGetLayerWrap(layerDefs);
-  const opacities = imageUtilGetLayerOpacities(layerDefs);
+  let layersArray = imageUtilGetLayers(layerDefs, proj.id);
```
jasontk19, line 88: Since you are also checking for the presence of orbit track layers within the `addOrbitTrackRevisions` function, you could skip checking for it here and condense the changes to this function by quite a few lines to something like:
```
  const {
    layersArray,
    layerWraps,
    opacities,
  } = addOrbitTrackRevisions(
    imageUtilGetLayers(layerDefs, proj.id),
    imageUtilGetLayerWrap(layerDefs),
    imageUtilGetLayerOpacities(layerDefs),
  );
```

This would require reconciling the names of the properties being returned from `addOrbitTrackRevisions` with those used in the rest of this function.

jasontk19 (approved):
Looks good!  Tried a number of layer orders and opacity changes with requests against WVS SIT and it's all working fine. Just a minor optional suggestion for condensing the code changes some.

web/js/modules/image-download/util.js
edplato: I like that! Will do.

PR #3381, opened 2021-02-23
dont require escaping underscores mid-word

jasontk19:
## Description

* No longer require escaping underscores in markdown if they are in the middle of a word `like_this`
* Existing escaped underscores are unaffected and don't need to be changed


minniewong (approved):
Seems to work! Hooray! 🎸

PR #3383, opened 2021-02-24
new preview image with HLS imagery

jasontk19:
## Description

New preview image for README.md
preview


edplato:
I think you want to delete `web/images/preview.gif` and replace it with `web/images/preview.jpg` for the README but currently the `web/images/preview.jpg` is being used in the `og:image` meta tag, so paths would need updating.

We might want to change filenames for clarity sake since we previously had a preview.gif and preview.jpg, or even better remove the gif and just use the same preview.jpg throughout for the README.md and meta tag which helps simplify.

jasontk19:
Ahhh whoops... Good catch!

Is there a reason why we wouldn't want to use the same image for both the readme and the `og:image` meta tag? I'm leaving it as is for now and renaming this new one to `readme-preview.jpg` for clarity but curious if we need two different images for these different cases.

edplato (approved):
Looks good! I agree both preview images could be the same to limit how many we need to look at in the future.

PR #3395, opened 2021-03-01
Fix tour step with custom palette issues

### LINKED ISSUE #3392: Error thrown on changing date on tour step
**Describe the bug**
Error thrown on changing date on tour step. The error appears to happen in `web\js\map\ui.js` `updateDate()`.

**To Reproduce**
1. Go to `?l=VIIRS_SNPP_Thermal_Anomalies_375m_Day(hidden),VIIRS_SNPP_Thermal_Anomalies_375m_Night(hidden),Reference_Labels(hidden),Reference_Features(hidden),Coastlines,VIIRS_SNPP_CorrectedReflectance_TrueColor,MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor(hidden)&lg=false&tr=australia_fires_2019_2020&t=2019-09-07-T15%3A25%3A24Z`
2. Click through tour step till STEP 5
3. Click left arrow to change date
4. See error `Uncaught TypeError: Cannot read property 'forEach' of undefined`

Also seeing same error if change date is done after STEP 7 - it should be noted for debug that both steps before STEP 5 and STEP 7 had animations.

 - OS: Windows
 - Browser: Chrome

jasontk19:
## Description

Fixes #3392 

The individual requests for palettes were, for some reason, causing Redux actions to be mishandled by the map.  This prevents the request of a single palette when it is already loaded.  During each tour step change, the palettes for that step are loaded in bulk anyway.

This fixes the issue with the layer states in the map and the Redux store being out of sync, which caused both issues (changing date, map not zooming to correct location) that were noticed while investigating this issue.

### DIFF
--- a/web/js/containers/sidebar/layer-row.js
+++ b/web/js/containers/sidebar/layer-row.js
@@ -99,8 +99,11 @@ function LayerRow (props) {
     }
   };
 
+  // Request the layer palette only if it hasn't been loaded and is not currently being loaded
   useEffect(() => {
-    if (!isLoading && layer && hasPalette) requestPalette(layer.id);
+    if (!isLoading && layer && hasPalette && lodashIsEmpty(renderedPalette)) {
+      requestPalette(layer.id);
+    }
   }, [layer.id]);
 
   const getDisabledTitle = (layer) => {


edplato (approved):
This fixes #3392 tour LOCATION_POP_ACTION palette, zoom, and related issues.

Benjaki2 (approved):
nice!

PR #3399, opened 2021-03-03
Add distraction free mode query string param, add shortcuts section in about page

### LINKED ISSUE #3391: Add distraction-free mode query string option
Add distraction-free mode query string option to enable feature on page load.

edplato:
## Description

Fixes #3391  .

- [x] Add distraction free mode query string parameter `df` activated as `df=true` and `false` by default. This enhancement is accompanied by a timed out alert when distraction free mode is initially activated and/or a page is navigated to with the feature active.
- [x] Add keyboard shortcuts section to about page that is only visible on app mobile breakpoint width. Revise about page css, remove unused pages files. and move styling out of pages section.
- [x] Revise toolbar button to use separate "exit distraction free mode" button and remove reuse of info button.

## How To Test

1. Open with this URL parameter `/?df=true` to see query string loaded distraction free mode and alert dialog
2. Open about page in desktop width to see keyboard shortcuts

## PR Submission Checklist

@nasa-gibs/worldview


jasontk19 (changes requested):
Works well!  Just a few code suggestions.  

Also, could we add some conditional styling to the logo area of the sidebar when in distraction free mode?  When the rest of the sidebar is hidden this top part looks noticeably chopped off since the bottom doesn't have the border or rounded edges that the rest of the element has:
Screen Shot 2021-03-15 at 9 47 57 AM

web/css/about.css
```diff
@@ -0,0 +1,126 @@
...
+    :last-child {
+      text-align: right;
+    }
+  }
+  td {
+    padding: 7px 0;
+  }
+  code {
+    color: #f60;
+    padding: 1px 8px;
```
jasontk19, line 117: A slight tweak to the padding here makes these look more like keyboard keys IMO:
Screen Shot 2021-03-19 at 2 15 29 PM
px;
```suggestion
    padding: 1px 5px;
```

jasontk19 (approved):
I like the new rounded look to the Worldview header when the rest of the sidebar is not visible in distraction free mode.  A couple related thoughts:

* The opacity feels almost a little to weak, particularly at mobile (or "embedded") size
* If clicking this took you out of distraction free mode, we wouldn't need the "eye" button so even less UI on screen.  That might be too far from where distraction free mode is enabled to make sense though.

Screen Shot 2021-03-19 at 2 17 07 PM
Screen Shot 2021-03-19 at 2 12 29 PM

Up to you to decide if anything is worth changing here.

edplato:
I agree the opacity is a little too light - I darkened it up some. I think I'll leave the different eye/WV logo buttons for now since clicking the WV logo can be an escape to get the default page that I see myself using a lot.

Related to that, now I think I should also add the 'Distraction Free' option that's in desktop in the mobile menu as well since the `df` query string is available in mobile accessed links:
image

jasontk19:
> I think I'll leave the different eye/WV logo buttons for now since clicking the WV logo can be an escape to get the default page that I see myself using a lot.

I guess my thinking was that, only in distraction free mode would the logo just take you out of distraction free mode.  Otherwise it would still retain the existing functionality of resetting to defaults.  Resetting to defaults didn't seem like something a user would be likely to do in DF mode but swapping the functionality on this between modes could be confusing so I'm fine leaving it as it.

PR #3406, opened 2021-03-10
create PR with layer template when layers modified

jasontk19:
## Description

When merged, these changes should cause a draft PR to be automatically opened anytime someone pushes code that modifies a file in the layers directory.  The PR will use a custom template that has a checklist for layer PR specific things and some helpful testing notes on things to look for when testing layer PRs.  It will also automatically label the pr with the "layer" label which will help group the layer changes in the changelog properly when generating release notes.

To test this, I think we will have to merge it then try pushing some code to see what happens.


minniewong (approved):
Looks good to me!

PR #3434, opened 2021-03-26
Tour step location change fix

### LINKED ISSUE #3418: Map location not updating on tour step change
**Describe the bug**
Only happening in `develop`.  For some reason the map location/extents are not being loaded properly for tour steps.

**To Reproduce**
1. Open "Australia Bushfires" tour story
2. Click through steps
3. Observe incorrect map locations/extents

**Screenshots**
Screen Shot 2021-03-17 at 10 12 11 AM

jasontk19:
## Description

Fixes #3418.

I'm not entirely sure why this error was suddenly happening.  For some reason, `payload.query` was undefined in the `LOCATION_POP_ACTION` reducer in `map/ui.js`.  In this instance, `payload` is actually the `location` object which comes from the `history` package and is imported into `web/js/containers/tour.js`.  

This code hasn't changed for several versions even though it appears to work in the currently deployed PROD but does not work locally for `develop` or `main`.  Neither the `history` nor the `redux-location-state` packages have been updated recently.

Since `location.query` was essentially what was on the current Redux state, I updated this to pull directly from `state` instead which seems to work fine.  I'm not sure if this should be included in 3.9.1 or 3.10.0 since it appears to work in PROD 🤷🏻‍♂️  

## How To Test

1. Open Worldview
2. Load Australia bushfires story
3. Step through and confirm map extent location updates properly as you work through the story.  Compare against PROD if necessary to confirm.


Benjaki2 (approved):
Nice! I would say the cause is most likely a conflicting action. IMO It would be good to get this fix out since we don't know when it was introduced and when it will show up.

minniewong (approved):
🚀  Zooming to all the right places

PR #3447, opened 2021-04-06
Update dockerfile

jasontk19:
## Description

The dockerfile is used to set up the build environment in Bamboo.  This updates the Python and Node versions to match what we are testing against in our Github Actions CI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM centos:7
 
 RUN yum install -y epel-release && \
-    yum install -y https://centos7.iuscommunity.org/ius-release.rpm && \
     yum install -y \
     "@Development Tools" \
     cairo-devel \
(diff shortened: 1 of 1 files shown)


jasontk19:
Here is the build in Bamboo based on this branch to see it working: `WV-SWP1560-3`

edplato (approved):
I think it looks good. Did you find out what `yum install -y https://centos7.iuscommunity.org/ius-release.rpm && \` was doing? It looks like a dead link and doesn't seem to be causing any build issues with removing it.

jasontk19:
@edplato I'm not really sure what that RPM was supposed to be but, I had to remove it because it was failing the build since the file doesn't exist at that URL anymore.

Benjaki2 (approved):
I am still having trouble building it on my machine but the code looks good to me. There is probably something wrong with my version of docker. If it works on your computer and bamboo... I say, go for it!

jasontk19:
> I am still having trouble building it on my machine but the code looks good to me. There is probably something wrong with my version of docker. If it works on your computer and bamboo... I say, go for it!

Hmmm yeah it seems to work fine locally... What errors are you getting?

Benjaki2:
> > I am still having trouble building it on my machine but the code looks good to me. There is probably something wrong with my version of docker. If it works on your computer and bamboo... I say, go for it!
> 
> Hmmm yeah it seems to work fine locally... What errors are you getting?

Working now after restarting my terminal.

PR #3452, opened 2021-04-07
Allow future multidates in date range for active layer building without new requests

edplato:
## Description

- [x] Use start/end date capability in `dateRange` for active layer building. This uses known max `dateInterval` of a layer from traversing its  `def.dateRanges`. This will allow single requests of active layer date intervals (4-day, 8-day, etc.) beyond what's known in the GC (e.g., 8-day layer beyond known GC will only request first day of range once instead of requesting every single day on date change within 8-day span of days).

## How To Test

1. Open in UAT (PROD is newly deployed, so won't have the necessary GC lag to demonstrate fix):
`?v=-108.42460335224303,-46.812755486230635,50.903521647756975,56.57227570998278&l=MODIS_Combined_L4_LAI_4Day&lg=false&t=2021-03-14-T18%3A00%3A00Z`
2. Open network panel to see requests, and use date arrows to change date by `1 day` into the future while staying within the `4 day` span (opening the layer coverage panel is helpful here)
3. Note **NEW** date requests on each day within the `4 day` span of days

PR Fix:
1. Open `http://localhost:3000/?v=-108.42460335224303,-46.812755486230635,50.903521647756975,56.57227570998278&l=MODIS_Combined_L4_LAI_4Day&lg=false&t=2021-03-14-T18%3A00%3A00Z`
2. Open network panel to see requests, and use date arrows to change date by `1 day` into the future while staying within the `4 day` span (opening the layer coverage panel is helpful here)
3. Note **NO NEW** daterequests on each day within the `4 day` span of days

## PR Submission Checklist

@nasa-gibs/worldview

### DIFF
--- a/e2e/features/layers/layers-vector-test.js
+++ b/e2e/features/layers/layers-vector-test.js
@@ -1,7 +1,7 @@
 const skipTour = require('../../reuseables/skip-tour.js');
 
-const damsLayerQuerystring = '?v=-70.43215000968726,28.678203599725197,-59.81569241792232,31.62330063930118&l=GRanD_Dams,Reference_Labels(hidden),Reference_Features(hidden),Coastlines,VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor';
-const damsLayerWMSZoomLevelQuerystring = '?v=-166.0537832499445,-8.893604135881553,79.78417648048394,59.303969410599414&l=GRanD_Dams,Reference_Labels(hidden),Reference_Features(hidden),Coastlines,VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor';
+const damsLayerQuerystring = '?v=-70.43215000968726,28.678203599725197,-59.81569241792232,31.62330063930118&l=GRanD_Dams,Reference_Labels_15m(hidden),Reference_Features_15m(hidden),Coastlines_15m,VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor';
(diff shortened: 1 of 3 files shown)


jasontk19:
I am testing this locally by having removed the last `dateRanges` object from the compiled `wv.json` so that the latest one for this layer is for 2020:

```
{
  "startDate": "2020-01-01T00:00:00Z",
  "endDate": "2020-12-26T00:00:00Z",
  "dateInterval": "4"
}
```

This way, it should be possible to test all of the 2021 dates up until now using the new date "prediction" changes introduced here.

What I am seeing is that, it definitely works in that requests are only made every 4 days.  But, since it is calculating onward from the last known start date in `n-day` intervals, eventually the dates being requested are different from those that would actually be requested if it were built from an updated GC doc.  This is because each year for this layer resets to start at `yyyy-01-01` while the predictive method just continually adds 4 days from the last known start date.  I'm concerned that there's a possibility that this might cause the wrong `n-day` interval to show but I'm not sure yet if that's possible.  

You can see this issue illustrated in the layer coverage bar by comparing production to a local instance built off this branch.  Here is what it looks like in PROD, with an odd smaller interval right at the end of the previous year:
Screen Shot 2021-04-09 at 9 55 58 AM

And here on this branch you can see it's just smooth 4 day intervals all the way:
Screen Shot 2021-04-09 at 9 56 04 AM


If all `n-day` layers reset at the beginning of the year maybe we need to do that here as well?

jasontk19:
Oh and to be specific, locally the dates requested when crossing the year are:
2020-12-26, 
2020-12-30, 
2021-01-03, 
2021-03-07, etc

Whereas in PROD the dates are:
2020-12-26, 
2020-12-30, 
2020-12-31, (this one is technically due to the issue we are addressing here) 
2021-01-01, (here it resets to the first day of the year)
2021-03-05, etc

edplato:
Good test case to catch this issue. I'll have to revisit what kind of approaches I can work with using the existing GC. I'm seeing a lot of potential issues with this approach since layer coverage sometimes starts on a random day and then subsequent date range start dates follow a pattern (e.g., start on the first of the year), but not always. I do fear that predicting future date spans due to lacking up to the date data of the actual coverage opens the flood gates for something always being wrong.

jasontk19:
> I'm seeing a lot of potential issues with this approach since layer coverage sometimes starts on a random day and then subsequent date range start dates follow a pattern (e.g., start on the first of the year), but not always. I do fear that predicting future date spans due to lacking up to the date data of the actual coverage opens the flood gates for something always being wrong.

100% agree. The inconsistent date ranges/range patterns make this really hard if not impossible.

edplato:
Here is a list of layers in our config that:
1) have `dateRanges` with a `dateInterval` other than `1`;
2) multiple `dateRanges` ranges since the first `dateRanges` object `startDate` is an arbitrary starting date; and
3) have non Jan 01 `startDate` (i.e., not YYYY-01-01T00:00:00Z) so couldn't be "reset" for future coverage:

- AMSRU2_Snow_Water_Equivalent_5Day 
- MEaSUREs_Ice_Velocity_Greenland 🔴
- Aquarius_Soil_Moisture_Weekly 🔴
- Aquarius_Sea_Surface_Salinity_L3_7Day_Snapshot 🔴
- Aquarius_Sea_Surface_Salinity_L3_Monthly 🔴
- Aquarius_Wind_Speed_L3_7Day_Snapshot 🔴
- GRACE_Tellus_Liquid_Water_Equivalent_Thickness_Mascon_CRI 🔴
- JPL_MEaSUREs_L4_Sea_Surface_Height_Anomalies 🔴
- MODIS_Aqua_L3_EVI_16Day
- MODIS_Aqua_L3_NDVI_16Day
- OSCAR_Sea_Surface_Currents_Meridional
- OSCAR_Sea_Surface_Currents_Zonal
- TOPEX-Poseidon_JASON_Sea_Surface_Height_Anomalies_Reconstructed 🔴

 🔴 = `inactive`

Being able to rely on a `YYYY-01-01T00:00:00Z` date reset for rolling date ranges will allow better prediction of future multi dates.

PR #3479, opened 2021-04-23
3471 Fix event selection carry over on projection switch

### LINKED ISSUE #3471: events.notfocus.3: When switching from Arctic back to Geo, previously selected Geo event is no longer selected
When switching from Arctic back to Geo, previously selected Geo event is no longer selected - "King Fire should be highlighted in sidebar and visible on map". 

Works in production.

jasontk19:
## Description

Fixes #3471

- don't deselect event when it's not visible in current projection
- remove code that tried to select track points on date change for selected event

## How To Test

Follow the steps for `event.focus` in the test suite.


edplato (approved):
Looks good 👍🏽

PR #3542, opened 2021-06-04
Query CMR with concept id on layer add to validate

### LINKED ISSUE #3509: Smart handoffs should not fail on bad concept ids
**Describe the bug**
We should make sure that WV gracefully handles instances where a collection in CMR is removed that was available during the WV build.  Specifically we should check that the following actions do not cause things to look broken:

* Requests for granule counts when a layer is selected in smart handoffs or when the bounding box is moved
* Transitioning to Earthdata Search with a bad id does not cause issues

The best approach to accomplish this may be to actually modify the existing implementation so that collection data is not requested from CMR at build time but, instead is requested when a layer is present in the smart handoffs tab.  This way if the collection does not exist at that time, we can avoid showing that layer in the panel altogether.

jasontk19:
## Description

Fixes #3509 

Updates smart-handoffs to make a CMR request for each concept ID assigned to a layer to verify that it is still valid before displaying that option as an available downloadable product to the user.

## How To Test

1. Build the app as normal
2. Open in browser
3. Go to Data tab
4. Confirm "Corrected Reflectance (True Color) Suomi NPP / VIIRS" layer has one NRT data collection choice
5. Open `web/config/wv.json`
6. Search for `C1344298008-LANCEMODIS` and change this to an invalid value such as `C123456-LANCEARMSTRONG`
7. Restart webpack server
8. Confirm "Corrected Reflectance (True Color) Suomi NPP / VIIRS" layer no longer shows in data tab

### DIFF
--- a/e2e/features/smart-handoff/smart-handoff-test.js
+++ b/e2e/features/smart-handoff/smart-handoff-test.js
@@ -52,9 +52,10 @@ module.exports = {
     c.click(dataTabButton);
 
     // Ensure layer is now showing as an option for download
-    c.expect.element(cloudRadiusRadioButton).to.be.present;
-
-    c.click(cloudRadiusRadioButton);
+    c.waitForElementVisible(cloudRadiusRadioButton, TIME_LIMIT, (e) => {
+      c.click(cloudRadiusRadioButton);
+      c.pause(500);
+    });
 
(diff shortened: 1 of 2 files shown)


edplato (approved):
This handles the issue well 👍🏽

PR #3549, opened 2021-06-10
Fix component update errors

### LINKED ISSUE #3548: Fix "Cannot update a component from inside the function body of a different component" errors
**Describe the bug**
Some checkboxes trigger a React warning which is logged as a console error:
```
Cannot update a component from inside the function body of a different component.
```

This is usually due to:

1. Dispatching an action in the main body of a functional component (should be wrapped in `useEffect`)
2. Setting state from within a `render` method

**To Reproduce**
1. Turn on any orbit track layer

**Expected behavior**
No console error

**Additional context**
https://github.com/facebook/react/issues/18178#issuecomment-595846312

jasontk19:
## Description

Fixes #3548.

- don't trigger callbacks that set state in component A while setting state in component B (`checkbox.js`)
- dont dispatch action during render method/in main body of function component (`orbit-tracks.js`)

## How To Test

1. Turn on orbit track from sidebar
2. Confirm no console errors

If you've seen these errors elsewhere in the app, note them here and I will try to address in this PR.

### DIFF
--- a/web/js/components/util/checkbox.js
+++ b/web/js/components/util/checkbox.js
@@ -34,12 +34,10 @@ export default class Checkbox extends React.Component {
   }
 
   handleChange(e) {
-    this.setState((prevState) => {
-      const { onCheck } = this.props;
-      const checked = !prevState.checked;
-      if (onCheck) onCheck(checked);
-      return { checked };
-    });
+    const { checked } = this.state;
+    const { onCheck } = this.props;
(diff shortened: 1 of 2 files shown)


edplato (approved):
Looks good to me. That's an interesting use of a hook to add the PaletteLegend component to state!

jasontk19:
Hopefully a "good" interesting 😅

PR #3593, opened 2021-07-07
Don't make api calls on npm run build:config

### LINKED ISSUE #3415: npm run build:config should not make API requests
Previously, when running this script we would only compile local configs into the `wv.json` file so you could quickly test out update layer or tour config changes.  Since the vis metadata changes were introduced, this now also makes request to the GIBS vis metadata API and CMR to get additional metadata for layers.

We should rework the build scripts so that these other metadata are stored separately in the build directory when fetched so that they don't have to be fetched on every build run but instead will only be fetched when the `npm run build` script is run.

jasontk19:
## Description

Fixes #3415

Keep a temp copy of vis metadata in a local build folder just like we do with the gc data so that network requests don't need to be made when running `npm run build:config`


Benjaki2 (approved):
💯 Looks good!

PR #3594, opened 2021-07-07
Display appropriate granule count

### LINKED ISSUE #3581: Smart handoffs area of interest filtering issue
**Describe the bug**
When toggling on and off the "Set Area of Interest" checkbox on the Data tab, the async timing of requests for granules is somewhat off resulting in incorrect information showing in the "Available granules for YYYY MMM DD" section.  For example, sometimes the available granules info will show "n of x" when the checkbox is disabled when it should just show "x".

This appears to happen when you quickly toggle the checkbox and is likely an indicator that we need to cancel any pending requests when it is toggled so that we don't display stale information.

**To Reproduce**
1. Go to data tab
2. Click on "Set Area of Interest" twice in a row
3. Observe "n of x" count that should just be "x"

**Screenshots**
Screen Shot 2021-07-07 at 8 36 22 AM

jasontk19:
## Description

Fixes #3581

Only show the "x out of ..." message when the 'Set Area of Interest' checkbox is checked.

### DIFF
--- a/web/js/components/smart-handoffs/granule-count.js
+++ b/web/js/components/smart-handoffs/granule-count.js
@@ -133,7 +133,7 @@ export default function GranuleCount (props) {
         {!isLoading && (
           <>
             <span className="fade-in">
-              {granulesExist && selectedGranules >= 0 && `${selectedGranules} of `}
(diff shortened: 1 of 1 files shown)


Benjaki2 (approved):
✅

PR #3629, opened 2021-07-16
Allow layer info modals in embed mode

### LINKED ISSUE #3628: Clicking on the "i" (View description) in embed mode does not let you view the description
Clicking on the "i" in embed more to view descriptions does not allow you to view the description

edplato:
## Description

Fixes #3628  .

- [x] Allow layer info modals in embed mode

## PR Submission Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/containers/modal.js
+++ b/web/js/containers/modal.js
@@ -134,7 +134,7 @@ class ModalContainer extends Component {
 
     const isRestrictedDisplay = (isMobile && desktopOnly)
       || (!isMobile && mobileOnly)
-      || (isEmbedModeActive && size === 'lg');
+      || (isEmbedModeActive && size === 'lg' && !id.includes('LAYER_INFO_MODAL'));
     if (isRestrictedDisplay) {
       return null;
     }


web/js/containers/modal.js
```diff
@@ -134,7 +134,7 @@ class ModalContainer extends Component {
 
     const isRestrictedDisplay = (isMobile && desktopOnly)
       || (!isMobile && mobileOnly)
-      || (isEmbedModeActive && size === 'lg');
+      || (isEmbedModeActive && size === 'lg' && !id.includes('LAYER_INFO_MODAL'));
```
jasontk19, line 137: For this release this is fine but, I think there is probably a more elegant long term solution where we identify which modals to show (or not show) in embed mode by their id, define them in a more visible and consolidated location (e.g. top of the file or some other constants file) and check for them here.  Just having a blanket size check (`size === 'lg'`) seems like we might block other modals we didn't intend to.  

Also if we are blocking modals purely based on size, how certain are we that we aren't showing any clickable UI elements elsewhere that would otherwise be showing a modal?  It might be better to disable or hide the clickable element rather than block the resulting modal that would open as a result of clicking said element.

jasontk19 (approved):
👍🏻

PR #3634, opened 2021-07-20
Throw error when measurement names dont match

### LINKED ISSUE #2141: Allow layer list to ignore undefined measurements and continue building valid layers
**Describe the bug**
If a layer measurement is invalid (spelled wrong, name change, etc.) in the `config.categories` then when trying to build the layer list a single failure (where the measurement doesn't match the measurement in `config.measurements`) will break layer list functionality and throw errors.

The invalid measurement is `undefined` in `hasMeasurementSource` in `web\js\modules\layers\selectors.js`. In this case 'Dust Score' was breaking the layer list since it appeared to be looking for a match using `Dust` instead. This will need further investigation to track down the best place to add validation and/or centralize data sources since `category.measurements` are being used to access config data in `measurementConfig`.

jasontk19:
## Description

Fixes #2141

A measurement name that doesn't match up will throw a descriptive error when opening the Product Picker.

## To test

1. Change "Dust" on line 3 of `config/default/common/config/wv.json/measurements/Dust.json` to "Dust score"
2. `npm run build:config`
3. Load the app
4. Open the product picker
5. See descriptive error
6. Revert changes to configs
7. Change "Dust" on line 12 of `config/default/common/config/wv.json/categories/hazards_and_disasters/Dust Storms.json` to "Dust Score"
8. `npm run build:config`
9. Load the app
10. Open the product picker
11. See descriptive error

### DIFF
--- a/web/js/components/layer/product-picker/browse/category-cell.js
+++ b/web/js/components/layer/product-picker/browse/category-cell.js
@@ -17,6 +17,13 @@ const CategoryCell = (props) => {
     }
     : {};
 
+  const checkForSources = (measurement) => {
+    if (measurementConfig[measurement]) {
+      return hasMeasurementSource(measurementConfig[measurement]);
+    }
+    throw new Error(`No measurement config entry for "${measurement}".`);
+  };
+
   return (
     <div
       key={category.id}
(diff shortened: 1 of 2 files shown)


edplato (approved):
Looks good 👍🏽

minniewong (approved):
👍

PR #3652, opened 2021-07-28
Preload map tiles and click + hold animation

### LINKED ISSUE #3653: Pre-load map tiles for smoother user experience
Tiles for prev/next time step increment could be pre-loaded to improve user experience, especially when users click and hold the `<` or `>` arrows to "animate".

 * Pre-cache tiles outside of animation mode
   * On page load, load tiles one step before and after current date/time 
   * After a step in either direction, load tiles two steps in that direction
     * Subsequent steps will only result in loading one step worth of tiles, two steps out from current date/time
   * Pre-cache tiles when step interval changes

jasontk19:
## Description

Fixes #3653

* preload tiles on left and right of current datetime
* tweak timeline debounce/throttles slightly, especially dragger
* set OL TileLayer `preload` option to `0` to prevent loading other zoom levels which noticeably reduces overall tile requests, especially with several layers visible

### Preloading in both directions

1. Load `http://localhost:3000/?l=MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2021-04-30-T22%3A00%3A00Z`
2. Step forward once
3. Confirm tiles are already loaded
4. Load `http://localhost:3000/?l=MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2021-04-30-T22%3A00%3A00Z`
5. Step backward once
6. Confirm tiles are already preloaded
7. Continue stepping backward and confirm tiles are preloaded

### Preload on interval change

1. Load `http://localhost:3000/?l=MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2021-04-30-T22%3A00%3A00Z`
2. Change the time step to `1 Month`
3. Step forward
4. Confirm tiles preloaded
5. Step backward twice 
6. Confirm tiles preloaded

### Preloading both sides of compare mode

1. Load `http://localhost:3000/?l=GHRSST_L4_MUR_Sea_Surface_Temperature_Anomalies,MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=false&l1=MODIS_Aqua_Aerosol,Reference_Labels_15m(hidden),Reference_Features_15m(hidden),Coastlines_15m,VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg1=false&ca=true&t=2021-07-06-T15%3A33%3A45Z&t1=2021-06-13-T13%3A33%3A45Z`
2. Step forward/backward with `A` side selected
3. Confirm tiles preloaded
4. Switch to `B` side
5. Step foward/backward
6. Confirm tiles are preloaded

### Preloading "Quick Animation"

1. Load `http://localhost:3000/?v=-106.29560998856994,18.14651067122917,-80.15325921753754,44.3476082979268&z=4&i=5&ics=true&ici=5&icd=10&l=GOES-East_ABI_GeoColor,Coastlines_15m&lg=true&t=2021-08-09-T17%3A30%3A07Z`
2. Click and hold left arrow to preload tiles and animate backwards in time
3. Once preload finishes, continue holding arrow to animate 6-10 steps
4. Release mouse to stop animation
5. Click 5-10 more steps backwards
6. Confirm tiles continue to preload as you step backwards
7. Manually set time back to `2021-AUG-09 17:30`
8. Step forward in time several steps and confirm tiles continue to preload

### Preloading After Extent Change

1. Load `http://localhost:3000/?v=-106.29560998856994,18.14651067122917,-80.15325921753754,44.3476082979268&z=4&i=5&ics=true&ici=5&icd=10&l=GOES-East_ABI_GeoColor,Coastlines_15m&lg=true&t=2021-08-09-T17%3A30%3A07Z`
2. Click and hold left arrow to preload tiles and animate backwards in time
3. Once preload finishes, continue holding arrow to animate 6-10 steps
4. Release mouse to stop animation
5. Pan/zoom to a different map area
6. Click and hold either direction to animate
7. Confirm tiles preload properly before animation begins


edplato (changes requested):
I'm liking the enhancement. Here are some areas I ran into issues or could use some attention:

1. There are some changes in the timeline date changing that is causes some update issue. It looks like the core issue is around removing:
`this.debounceDateUpdate = lodashDebounce(changeDate, 8);`

    The main issues I'm encountering are when dragging the timeline dragger the map isn't smooth updating when crossing over days and waiting for a pause to show imagery. I remember this being a requested addition to the timeline dragging functionality initially.
  
    Also, the date update change is causing some issue with the `timelineAxis` updating on date change - this can be seen if in the DAY zoom level and the interval is MONTH - changing a month or two using the date change arrows doesn't update the dragger/axis. Replacing `this.throttleSelectDate` with the existing `this.debounceDateUpdate = lodashDebounce(changeDate, 8);` seems to fix this.

2. Hold the left arrow button and it hangs waiting to preload. No apparent error and it raises the question if there should be some sort of timeout if it does hang too long? http://localhost:3000/?v=3.7195557924551723,35.87784076755478,5.976944207544814,37.53677923244522&l=VIIRS_SNPP_Thermal_Anomalies_375m_Night(hidden),VIIRS_SNPP_Thermal_Anomalies_375m_Day(hidden),VIIRS_NOAA20_Thermal_Anomalies_375m_Night(hidden),VIIRS_NOAA20_Thermal_Anomalies_375m_Day(hidden),MODIS_Aqua_Thermal_Anomalies_Night(hidden),MODIS_Aqua_Thermal_Anomalies_Day(hidden),MODIS_Terra_Thermal_Anomalies_Night,MODIS_Terra_Thermal_Anomalies_Day,OMPS_Aerosol_Index(hidden),OMPS_Aerosol_Index_PyroCumuloNimbus(hidden),VIIRS_SNPP_DayNightBand_ENCC(hidden),VIIRS_SNPP_DayNightBand_At_Sensor_Radiance(hidden),IMERG_Precipitation_Rate(hidden),Reference_Labels_15m,Reference_Features_15m,VIIRS_SNPP_CorrectedReflectance_BandsM11-I2-I1(hidden),VIIRS_NOAA20_CorrectedReflectance_BandsM11-I2-I1(hidden),VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_Bands721(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_Bands721(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=false&t=2021-04-07-T00%3A00%3A00Z

3. Same link as above and switch to Antarctic after load throws this error on moving mouse onto the map:
image
image

web/js/containers/timeline/timeline.js
```diff
@@ -1084,6 +1025,19 @@ class Timeline extends React.Component {
     const containerDisplayStyle = {
       display: isDistractionFreeModeActive ? 'none' : 'block',
     };
+
+    const renderDateChangeArrows = () => (
```
edplato, line 1029: Can move outside of render.

web/js/components/animation-widget/preload-spinner.js
```diff
@@ -0,0 +1,37 @@
...
+      toggle={onClose}
+      size="sm"
+      backdrop={false}
+      wrapClassName="clickable-behind-modal"
+    >
+      <ModalHeader toggle={onClose}>{title}</ModalHeader>
+      <ModalBody>
+        {bodyMsg && (
+          <div style={{
+            minHeight: 30, marginBottom: 30, fontSize: 14,
```
edplato, line 19: I kind of like the compactness of styles within smaller reusable files like this, but want to point out if not intended for any future reuse.

web/js/components/timeline/timeline-controls/date-change-arrows.js
```diff
@@ -1,82 +1,95 @@
...
+    };
+  }
+
+  componentDidMount() {
+    document.addEventListener('keydown', this.handleKeyDown);
+    document.addEventListener('keyup', this.handleKeyUp);
+  }
+
+  componentDidUpdate (prevProps) {
+    const { tilesPreloaded, arrowDown } = this.props;
```
edplato, line 36: May want to clarify as `arrowDownDirection`.

web/js/components/animation-widget/preload-spinner.js
jasontk19: I feel the same.  We don't normally do this but it's pretty common in some other React projects to have the related styles defined within the same component file.  I think this makes a lot of sense when the styles are for something like this, where it's not a set of styles that's going to be applied elsewhere, it's very specifically for parts of this component.

web/js/components/timeline/timeline-controls/date-change-arrows.js
jasontk19: Yeah it's kind of an overloaded variable but that was intentional.  If it evaluates as `''` then no arrow is being held down, otherwise it's `left` or `right` for which direction is being held down.  I think this is more concise than having a `isArrowDown` boolean and an `arrowDownDirection` for example.

jasontk19:
> I'm liking the enhancement. Here are some areas I ran into issues or could use some attention:
> 
> 1. There are some changes in the timeline date changing that is causes some update issue. It looks like the core issue is around removing:
>    `this.debounceDateUpdate = lodashDebounce(changeDate, 8);`
>    The main issues I'm encountering are when dragging the timeline dragger the map isn't smooth updating when crossing over days and waiting for a pause to show imagery. I remember this being a requested addition to the timeline dragging functionality initially.
>    Also, the date update change is causing some issue with the `timelineAxis` updating on date change - this can be seen if in the DAY zoom level and the interval is MONTH - changing a month or two using the date change arrows doesn't update the dragger/axis. Replacing `this.throttleSelectDate` with the existing `this.debounceDateUpdate = lodashDebounce(changeDate, 8);` seems to fix this.
> 2. Hold the left arrow button and it hangs waiting to preload. No apparent error and it raises the question if there should be some sort of timeout if it does hang too long? http://localhost:3000/?v=3.7195557924551723,35.87784076755478,5.976944207544814,37.53677923244522&l=VIIRS_SNPP_Thermal_Anomalies_375m_Night(hidden),VIIRS_SNPP_Thermal_Anomalies_375m_Day(hidden),VIIRS_NOAA20_Thermal_Anomalies_375m_Night(hidden),VIIRS_NOAA20_Thermal_Anomalies_375m_Day(hidden),MODIS_Aqua_Thermal_Anomalies_Night(hidden),MODIS_Aqua_Thermal_Anomalies_Day(hidden),MODIS_Terra_Thermal_Anomalies_Night,MODIS_Terra_Thermal_Anomalies_Day,OMPS_Aerosol_Index(hidden),OMPS_Aerosol_Index_PyroCumuloNimbus(hidden),VIIRS_SNPP_DayNightBand_ENCC(hidden),VIIRS_SNPP_DayNightBand_At_Sensor_Radiance(hidden),IMERG_Precipitation_Rate(hidden),Reference_Labels_15m,Reference_Features_15m,VIIRS_SNPP_CorrectedReflectance_BandsM11-I2-I1(hidden),VIIRS_NOAA20_CorrectedReflectance_BandsM11-I2-I1(hidden),VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_Bands721(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_Bands721(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=false&t=2021-04-07-T00%3A00%3A00Z
> 3. Same link as above and switch to Antarctic after load throws this error on moving mouse onto the map:
>    image
>    image

Nice finds, I will dig into this 👍🏻

web/js/map/layerbuilder.js
```diff
@@ -387,7 +387,7 @@ export default function mapLayerBuilder(config, cache, store) {
       sourceOptions.tileClass = lookupFactory(lookup, sourceOptions);
     }
     return new OlLayerTile({
-      preload: Infinity,
+      preload: 0,
```
Benjaki2, line 390: Is it necessary to disable this?

Benjaki2 (approved):
Seems to work really well

web/js/components/animation-widget/preload-spinner.js
jasontk19: @Benjaki2 I think that StackOverflow post is primarily focused on the performance argument of CSS vs inline styles and is about a decade old at this point.  Since then several popular libraries such as [Radium](https://github.com/FormidableLabs/radium) and [JSS](https://cssinjs.org/?v=v10.7.1) have been developed to try to get around this issue while also adding other powerful features.  I am not very familiar with the intricacies of either of these but at a glance they seem useful and, there seem to be some [solid arguments for ditching CSS classes in favor of styles defined in JS](https://speakerdeck.com/vjeux/react-css-in-js)

Regarding dynamic styles in JS, I think that's an even stronger argument for keeping the style declarations within the React component files themselves. It can be tedious trying to debug layout issues for components that are both being styled by a CSS stylesheet and inline, of which we have several examples.

That said, if you think it's worth moving these few style properties into a class definition within a CSS file for consistency I'm fine with doing so.

web/js/map/layerbuilder.js
jasontk19: I should have elaborated on this change in the description to provide some more understanding for why I chose to disable this:

It's not absolutely necessary but, personally I didn't think the slightly improved user experience justified the performance cost if we are going to preload tiles like this.  That said, I think the change warrants a broader discussion among the team.  For reference, I did a quick comparison of resources transferred with `preload` set to both `0` and `Infinity` on this branch to get an idea of the cost in bandwidth between the two possible approaches.  This URL has one baselayer and two vector fire layers turned on.

Using these starting params: `/?v=-128.6060460332096,35.31341649269199,-114.574428132895,47.54658937507575&l=MODIS_Terra_Thermal_Anomalies_Night,MODIS_Terra_Thermal_Anomalies_Day,Reference_Labels_15m,Reference_Features_15m,MODIS_Terra_CorrectedReflectance_TrueColor&lg=false&t=2021-08-08-T02%3A00%3A00Z`

* Initial page load
	* `preload: 0` =  3.0MB
	* `preload: Infinity` = 6.0MB

* 5 day steps
	* `preload: 0` = 11.8MB
	* `preload: Infinity` = 20.0MB

The difference seemed to be about a factor of two.  Considering that, when stepping forward/backwards in time and not modifying the zoom level the app is requesting a lot of tiles that the user might never see, it didn't seem worth it to request almost double the number of tiles.  Especially since the only downside to setting `preload: 0` is that the user will see some blank tiles around the edges when zooming out or, slightly blurry tiles momentarily while zooming in.

But again, I think it's worth more discussion to arrive at a team consensus on this change.

web/js/components/animation-widget/preload-spinner.js
Benjaki2: Both approaches seem like viable options. Maybe we should make it a dev topic for next week. I don't mind keeping the code here as is. This seems like a bigger topic; should move to something like Radium or keep the old approach?

jasontk19:
@edplato I have been trying to figure out what's causing that error you are seeing and I think I've narrowed it down to an existing bug in `main`.  So, normally this tile preloading function is not called unless you are using the animation widget to preload tiles.  That said, you should be able to reproduce the same error on `main` by just loading a vector layer in arctic projection and pressing play: `http://localhost:3000/?v=-5386240,-5480448,5386240,5480448&p=arctic&as=2021-08-01-T00%3A00%3A00Z&ae=2021-08-11-T00%3A00%3A00Z&l=VIIRS_NOAA20_Thermal_Anomalies_375m_All,Coastlines_15m,MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&ab=on&t=2021-08-01-T04%3A00%3A00Z`

I'm assuming that this error is happening in production as well but we aren't noticing because it's not being logged to the console and the screen blocking error notices are suppressed in prod.  It appears that one of the three tile coordinates being passed to the `loadTile` function by OpenLayers while running `tileGrid.forEachTileCoord(extent, currentZ, loadTile);` is sometimes `-1`.  This seems to possibly be related to an improperly calculated extent but I'm not entirely sure yet as tracing the issue through the OpenLayers code has proved challenging so far.  

Can you check and confirm that you also see this happening on `main` under the conditions described above and, if so, would you agree that this should be fixed in another ticket?  For now I can at least wrap the `tile.load()` call in a try/catch and log the error to the console that it is not screen-blocking in develop mode and also not just invisibly swallowed in deployed prod mode like is happening currently.

jasontk19:
Just pushed a small change to deal with the preload hanging in certain cases.  This appeared to happen only when a vector layer was loaded and so, for now I am skipping preloading for vector layers (only at zoom levels where they would be rendered as vectors, not WMS).  The issue seemed to stem from some of the vector tiles not ever resolving successfully, or failing, when we try to pre-load them.  Without a `tileloadend` or `tileloaderror` event to detect for even a single tile, the preloading promise could never resolve and so the whole thing just hangs.  For now, it seemed like an issue limited in scope enough that this feature should still be useful with this workaround in place.

web/js/components/animation-widget/preload-spinner.js
jasontk19: Completely moving to inline styles at this point could be a big lift but, I think it's worth discussing whether or not we want to explore a new approach moving forward.

edplato:
> @edplato I have been trying to figure out what's causing that error you are seeing and I think I've narrowed it down to an existing bug in `main`. So, normally this tile preloading function is not called unless you are using the animation widget to preload tiles. That said, you should be able to reproduce the same error on `main` by just loading a vector layer in arctic projection and pressing play: `http://localhost:3000/?v=-5386240,-5480448,5386240,5480448&p=arctic&as=2021-08-01-T00%3A00%3A00Z&ae=2021-08-11-T00%3A00%3A00Z&l=VIIRS_NOAA20_Thermal_Anomalies_375m_All,Coastlines_15m,MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&ab=on&t=2021-08-01-T04%3A00%3A00Z`
> 
> I'm assuming that this error is happening in production as well but we aren't noticing because it's not being logged to the console and the screen blocking error notices are suppressed in prod. It appears that one of the three tile coordinates being passed to the `loadTile` function by OpenLayers while running `tileGrid.forEachTileCoord(extent, currentZ, loadTile);` is sometimes `-1`. This seems to possibly be related to an improperly calculated extent but I'm not entirely sure yet as tracing the issue through the OpenLayers code has proved challenging so far.
> 
> Can you check and confirm that you also see this happening on `main` under the conditions described above and, if so, would you agree that this should be fixed in another ticket? For now I can at least wrap the `tile.load()` call in a try/catch and log the error to the console that it is not screen-blocking in develop mode and also not just invisibly swallowed in deployed prod mode like is happening currently.

I am seeing this in `main` and PROD too, so definitely can bump off to another ticket.

edplato:
http://localhost:3000/?v=-137.71389442231077,-106.17206175298806,137.71389442231077,106.17206175298806&z=2&as=2021-06-10-T00%3A00%3A00Z&ae=2021-08-19-T00%3A00%3A00Z&l=GOES-East_ABI_GeoColor,MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&av=10&ab=on&t=2021-08-19-T00%3A00%3A00Z
Switch to arctic (see console error)

I'm getting an `undefined` source error for a geographic only layer after initially switching to arctic or antarctic. Switching back to geographic then back to arctic or antarctic doesn't show the error.

edplato:
Related to the #3688 , I think the issues includes non-vector tiles as well. This arctic preload hangs for me with default layers http://localhost:3000/?v=-9035776,-7192576,6987776,5160960&p=arctic&as=2021-01-06-T00%3A00%3A00Z&ae=2021-02-06-T00%3A00%3A00Z&ab=on&t=2021-02-06-T18%3A00%3A00Z 

I've been trying to debug this and it seems that the given extent is too large (maxExtent: [-4194304, -4194304, 4194304, 4194304], but input extent: [-4194304, -4193944, 4194304, 4194664]). This can be seen in `loadTile` where the break happens if there is no `tile.src_` and `tile.state` of `4`, which looks related to one of the three tile coordinates being passed to `loadTile` being `-1` (from outside of valid extent). It looks like the old version of the promiseTileLayer in `develop` also had the same issues with `tile`, a `-1` tile coordinate, and large extent but it resolved those hanging tiles.

This is a quick fix I came up with that seems to work and fix the #3688 hang too - something like this can be worked in here or upstream:

Use a `safeExtent` within `promiseTileLayer` at top of function...
```
  const maxExtent = map.getView().getProjection().getExtent();
  const safeExtent = olExtent.containsExtent(maxExtent, extent)
    ? extent
    : maxExtent;
```

jasontk19:
@edplato Thanks for digging into this a bit, I really appreciate having a second set of eyes on it.  The invalid extents are exactly what I was seeing as well and adding a check for `geographic` projection definitely fixes the issue.

As for the hanging issue with vector layers, fixing the invalid extents didn't seem to help that.  If I disabled the `if (layer.isVector) return Promise.resolve();` check in `promiseLayerGroup` which was my sort of intermediate hack/fix, the preloading still hangs if you are zoomed in enough to be loading vector tiles.  Did you disable this line and still see it not hang?

edplato:
After further checking, the vector didn't do an initial hang on preload that I was seeing before but still had a problem showing the tiles in animation, so there is an issue unrelated to the fix I recommended above but that should be addressed in #3688 . This PR looks good to me!

edplato (approved):
👍🏽

PR #3674, opened 2021-08-10
Match partial shortname

### LINKED ISSUE #3630: Allow for searches on parts of a shortname
It seems that searching on shortname only works when you enter the entire shortname into the layer picker, it would be good if the search worked on partial shortnames too. 

For example, the MODIS Land Surface Temperature layers start with "MOD11" or "MYD11". If you just search on those two names, nothing is returned. If you search for "MOD11_L2", "MOD11A1", "MOD11C3" you will get layers returned.

jasontk19:
## Description

Fixes #3630.

* Allows partial matching on `shortName` (but not other terms)

## How To Test

1. Search for a partial `shortName`
2. Confirm a match
3. Search for other items like "cloud"
4. Confirm results are what you expect

@nasa-gibs/worldview


minniewong (approved):
👍  Working well!

PR #3680, opened 2021-08-10
Add SEDAC vector layers to "Socioeconomic" layergroup

### LINKED ISSUE #3510: Consider assigning SEDAC vector layers to a single "Socioeconomic" layer group
Currently each layer lives in its own group, which isn't helpful when grouping since there is only 1 layer per group.

Screen Shot 2021-05-14 at 10 06 53 AM

jasontk19:
## Description

Fixes #3510

@nasa-gibs/worldview

### DIFF
--- a/config/default/common/config/wv.json/layers/sedac/GRUMP_Settlements.json
+++ b/config/default/common/config/wv.json/layers/sedac/GRUMP_Settlements.json
@@ -8,7 +8,7 @@
       "type": "vector",
       "tags": "vectors",
       "group": "overlays",
-      "layergroup": "Settlements",
+      "layergroup": "Socioeconomic",
       "product": "CIESIN_SEDAC_GRUMPv1_STLMNT01",
       "inactive": true,
       "palette": {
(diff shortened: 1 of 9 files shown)


jasontk19:
Are there other layers that we should add to this group? There are lots of other small groups like "Human Footprint" which it might makes sense to consolidate here.

minniewong (commented):
Not sure if it's weird to have vector and non-vector layers in the same group, but others that could possibly go in the socioeconomic layergroup could be (basing this on the layers that are singles):
Human Built-up and Settlement Extent, Impervious Surface, Urban Extents,

jasontk19:
> Not sure if it's weird to have vector and non-vector layers in the same group, but others that could possibly go in the socioeconomic layergroup could be (basing this on the layers that are singles):
> Human Built-up and Settlement Extent, Impervious Surface, Urban Extents,

Personally I don't think it's too weird if they're vector or not since the grouping is more based on what kind of data the layer represents rather than how it is represented.  I will add these other layers to the group!

minniewong (approved):
Looks great!

PR #3737, opened 2021-10-04
Don't show layer info for layers that aren't available in current projection

jasontk19:
## Description

Fixes [WV-1960](https://bugs.earthdata.nasa.gov/browse/WV-1960)

* Filter out layers by projection

## How To Test
1. Open app
2. Go to arctic projection
3. Open product picker
4. Browser "Cloud Fraction" layers
5. Confirm "Total Cloud Fraction L2 Day" layer info is not visible in details on right side

### DIFF
--- a/web/js/components/layer/product-picker/browse/measurement-metadata-detail.js
+++ b/web/js/components/layer/product-picker/browse/measurement-metadata-detail.js
@@ -42,10 +42,12 @@ function MeasurementMetadataDetail (props) {
     return () => (controller ? controller.abort() : null);
   }, [source]);
 
-  const renderMetadataForLayers = () => layers.map((l) => (
-    <div className="layer-description" key={l.id}>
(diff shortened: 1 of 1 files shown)


minniewong (changes requested):
Cloud Fraction, Cloud Top Height and Freeze/Thaw are all displaying some of the layer info for Aqua/AIRS which do not have any layers available in Arctic or Antarctic
Screen Shot 2021-10-04 at 3 37 12 PM
.

jasontk19:
> Just one more thing - Looks like the first 4 Brightness Temp layers re-appeared in the polar projections. As you can see, you can't actually check them. In prod, they don't even appear in the list. Screen Shot 2021-10-06 at 4 53 00 PM

I think I got a little rusty during my time off 😬 

Should be fixed now!

PR #3758, opened 2021-11-08
WV-1943 DAAC Facet

jasontk19:
## Description

- Add facet for DAAC/SIPS to layer picks

## How To Test

- Open layer picker
- Switch to search/filter mode
- Try filtering by DAAC/SIPS facet

@nasa-gibs/worldview

### DIFF
--- a/tasks/python3/getVisMetadata.py
+++ b/tasks/python3/getVisMetadata.py
@@ -16,7 +16,7 @@ input_file = args[1]
 output_file = args[2]
 
 # NOTE:  Only using these properties at this time
-use_keys = ['conceptIds']
+use_keys = ['conceptIds', 'dataCenter']
 layer_metadata = {}
 
 def get_metadata(layer_id, base_url):
@@ -24,7 +24,17 @@ def get_metadata(layer_id, base_url):
   if (response.status_code != 200):
     print('%s WARNING: No metadata config found for [%s]' % (prog, layer_id))
     return
-  layer_metadata[layer_id] = response.json()
+  metadata = response.json()
+
+  if metadata.get('conceptIds', None) is not None:
+    for collection in metadata.get("conceptIds"):
+      dataCenter = collection.get("dataCenter", None)
+      if metadata.get("dataCenter", None) is None:
+        metadata["dataCenter"] = [dataCenter]
+      elif dataCenter not in metadata["dataCenter"]:
+        metadata["dataCenter"].append(dataCenter)
+
+  layer_metadata[layer_id] = metadata
 
   # Remove any props we don't expect to use
   metadata_keys = dict(layer_metadata[layer_id]).keys()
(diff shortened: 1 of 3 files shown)


jasontk19:
Will need to wait until related GIBS PR is merged and in PROD to properly test

minniewong (commented):
The DAAC/SIPS tooltip doesn't seem to work

tasks/python3/getVisMetadata.py
```diff
@@ -48,6 +65,8 @@ def main(url):
     except Exception as e:
       print("%s:" % (e))
 
+  print(layer_metadata)
```
Benjaki2, line 68: @jasontk19 I feel like we may want to remove some of these prints. I am getting a printout of a massive object and I think this could make it harder to debug other issues.

Benjaki2 (changes requested):

tasks/python3/getVisMetadata.py
jasontk19: Whoops yeah that shouldn't still be in there, good catch

jasontk19:
> The DAAC/SIPS tooltip doesn't seem to work

How important do you think the links are for this?  We tried to get hyperlinks working in the tooltips before and went back and forth a bunch on how to get it to work satisfactorily.  I think we tried to get the tooltips to stay open longer before auto-hiding but that made the other tooltips feel sluggish when you don't need them to stay open in order to click a link.  We could probably do some more work to get some tooltips to stay open and some to auto-hide when you move the mouse away but that feels like work for a separate ticket.  If we don't absolutely need them for this I can just remove the links for now.

minniewong:
> > The DAAC/SIPS tooltip doesn't seem to work
> 
> How important do you think the links are for this? We tried to get hyperlinks working in the tooltips before and went back and forth a bunch on how to get it to work satisfactorily. I think we tried to get the tooltips to stay open longer before auto-hiding but that made the other tooltips feel sluggish when you don't need them to stay open in order to click a link. We could probably do some more work to get some tooltips to stay open and some to auto-hide when you move the mouse away but that feels like work for a separate ticket. If we don't absolutely need them for this I can just remove the links for now.

Sounds good - let's just get the tooltip to spell out what DAAC and SIPS means for now, and we can open a separate ticket for hyperlinks in the tooltips.

web/js/modules/product-picker/facet-config.js
```diff
@@ -58,7 +58,7 @@ export default [
     field: 'dataCenter',
     label: 'DAAC/SIPS',
     filterType: 'any',
-    tooltip: '<a href="Distributed Active Archive Centers / Science Investigator-led Processing System',
+    tooltip: 'Distributed Active Archive Centers / Science Investigator-led Processing System',
```
minniewong, line 61: change it to singular "Center"

minniewong (approved):
🎸

Benjaki2 (approved):
👍

PR #3759, opened 2021-11-09
WV-2054 - Orbit track measurement

jasontk19:
## Description

- Make sure orbit tracks don't get filtered out of orbit track measurement group

## How To Test

- Open layer picker
- Browse "All"
- Select "Orbital Tracks" group
- Confirm no error and all track layers show

@nasa-gibs/worldview


minniewong (approved):
👍  works great now!

PR #3776, opened 2021-12-07
WV-1959

jasontk19:
## Description

In order to allow requesting the PROD deployed version of our `wv.json` so that we can dynamically generate a docs page of layers, we need to be able to request  `config/wv.json` from other domains (e.g. `localhost` to test and `github.io` for the deployed docs)

This change allows cross-origin requests for only the `config/` directory.


Benjaki2 (approved):
Looks good to me but might be good to pass by @rboller & server-side team before releasing to make sure it doesn't introduce any security issues.

PR #3783, opened 2021-12-16
WV 1988 Layer picker should show category in mobile view

jwhurley1:
## Description

When in mobile view, the layer picker doesn't show the category that the user has selected under "Hazards and Disasters" or "Science Disciplines".  It's unclear which category the user is currently browsing once inside that category.

### Fixes

Added the category name below the parent category.  Used `mapStateToProps()` to call the `category.title` and `lastAction.type` from the Redux store.  When a category is selected or a child of a category is selected (i.e., if `category.title` contains a value), a div containing the category name will appear below the dropdown menu.  If the back button is pressed (i.e., `lastAction.type === 'PRODUCT_PICKER/TOGGLE_CATEGORY_MODE'`), then the div containing the category name is removed.

The toggling of the div is keyed off of the `selectedCategoryName` const.  This const is either blank if the back button pressed, otherwise assign to it the value of `category.title` from the Redux store. 

To create the div below the dropdown menu, the function `renderSelectedCategoryName()` is called.  If the const `selectedCategoryname` is blank, return nothing so that the space below the dropdown menu collapses.  Otherwise, the function will return the div containing the category name and create space below the dropdown menu for itself.

## How To Test

[Provide whatever information a reviewer might need to know in order to verify that the changes made in this PR are working as expected. If there are special build steps that need to be taken in order to get these changes to run (building while on a separate branch, running `npm ci`, etc) include them here.]

To test this feature, you will need to be in mobile view.  You can do this by shrinking your browser until the app switches to mobile mode, or accessing the Chrome Dev Tools and clicking the mobile device icon to simulate viewing on a mobile device.

Once in mobile mode, do the following:
- Open the Layers panel by clicking/tapping the Layers icon (has a "7" on it).
- Click/tap "+ Add Layers".
- Select any of the layers below or their categories.
- On the next page with the orange arrows, you will see the name of selected category below the "Hazards and Disasters", "Scientific Disciplines", etc. dropdown.

## PR Submission Checklist

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/layer/product-picker/browse/browse-layers.js
+++ b/web/js/components/layer/product-picker/browse/browse-layers.js
@@ -170,6 +170,11 @@ function BrowseLayers (props) {
       );
   }
 
+  function renderSelectedCategoryName() {
+    return selectedCategoryName == '' ? null
+    : <div className='selected-category'>{selectedCategoryName}</div>
+  }
+
   function renderMobileDropdown() {
     return (
       <div className="categories-dropdown-header">
@@ -195,7 +200,8 @@ function BrowseLayers (props) {
           </DropdownMenu>
         </Dropdown>
         {recentLayersHeader()}
-        {selectedCategoryName}
+
+        {renderSelectedCategoryName()}
       </div>
     );
   }
--- a/web/js/components/layer/product-picker/browse/browse-layers.js
+++ b/web/js/components/layer/product-picker/browse/browse-layers.js
@@ -255,7 +255,6 @@ function mapStateToProps(state) {
     proj,
     productPicker,
     layers,
-    lastAction,
   } = state;
   const {
     mode,
@@ -273,7 +272,6 @@ function mapStateToProps(state) {
     categoryType,
     categoryTabNames: config.categoryGroupOrder,
     measurementConfig: config.measurements,
-    lastAction,
     layerConfig: layers.layerConfig,
     listScrollTop,
     recentLayers,
--- a/web/js/components/layer/product-picker/browse/browse-layers.js
+++ b/web/js/components/layer/product-picker/browse/browse-layers.js
@@ -32,8 +32,10 @@ import safeLocalStorage from '../../../../util/local-storage';
 function BrowseLayers (props) {
   const {
     browser,
+    category,
     categoryTabNames,
     categoryType,
+    lastAction,
     mode,
     width,
     recentLayers,
@@ -253,9 +255,11 @@ function mapStateToProps(state) {
     proj,
     productPicker,
     layers,
+    lastAction,
   } = state;
   const {
     mode,
+    category,
     categoryType,
     listScrollTop,
     selectedMeasurement,
@@ -265,9 +269,11 @@ function mapStateToProps(state) {
   return {
     browser,
     mode,
+    category: category ? category.title : '',
     categoryType,
     categoryTabNames: config.categoryGroupOrder,
     measurementConfig: config.measurements,
+    lastAction,
     layerConfig: layers.layerConfig,
     listScrollTop,
     recentLayers,
--- a/web/js/components/layer/product-picker/browse/browse-layers.js
+++ b/web/js/components/layer/product-picker/browse/browse-layers.js
@@ -32,7 +32,6 @@ import safeLocalStorage from '../../../../util/local-storage';
 function BrowseLayers (props) {
   const {
     browser,
-    category,
     categoryTabNames,
     categoryType,
     mode,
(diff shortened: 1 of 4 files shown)


web/js/components/layer/product-picker/browse/browse-layers.js
```diff
@@ -66,6 +68,8 @@ function BrowseLayers (props) {
     && selectedProjection === 'geographic'
     && categoryType !== 'recent';
 
+  const selectedCategoryName = lastAction.type === 'PRODUCT_PICKER/TOGGLE_CATEGORY_MODE' ? '' : category;
```
jasontk19, line 71: This will work but, instead of having to look at the last action that was dispatched, I think we'd be better off setting the value itself in the Redux store when this action is dispatched.   So, in order to do that, you would just need to clear the value of `state.productPicker.category` when the `TOGGLE_CATEGORY_MODE` action is dispatched.  

You can do this by updating the reducer for that action to set this value along with what's already happening in there. For example right now it looks like:

```
case TOGGLE_CATEGORY_MODE: {
      return {
        ...state,
        mode: 'category',
        selectedLayer: null,
        selectedMeasurement: null,
        selectedMeasurementSourceIndex: 0,
      };
    }
```

And could be updated to:
```
case TOGGLE_CATEGORY_MODE: {
      return {
        ...state,
        mode: 'category',
        category: null
        selectedLayer: null,
        selectedMeasurement: null,
        selectedMeasurementSourceIndex: 0,
      };
    }
```

web/js/components/layer/product-picker/browse/browse-layers.js
```diff
@@ -167,6 +171,11 @@ function BrowseLayers (props) {
       );
   }
 
+  function renderSelectedCategoryName() {
```
jasontk19, line 174: This is fine but since there's not a lot of complexity in this render function, it could just be handled in one line in the `renderMobileDropdown()` function:

```
{selectedCategoryName && <div className="selected-category">{selectedCategoryName}</div>}
```

web/js/components/layer/product-picker/browse/browse-layers.js
```diff
@@ -263,9 +278,11 @@ function mapStateToProps(state) {
   return {
     browser,
     mode,
+    category: category ? category.title : '',
```
jasontk19, line 281: Instead of having to define an additional variable inside the component, you can just name the returned prop here to `selectedCategoryName`:

`selectedCategoryName: category && category.title`

... and then access it on props with that name, e.g. `const { selectedCategoryName, ..., ..., } = props;` at the top of the component.

jasontk19 (changes requested):

jasontk19:
Adding additional content here pushes the contents below outside of the container:
Screen Shot 2021-12-16 at 2 31 15 PM

web/js/components/layer/product-picker/browse/browse-layers.js
jwhurley1: Resolved in https://github.com/nasa-gibs/worldview/pull/3783/commits/50b1ef41dfb4c212ed2a1b37313a18b937f25ad6

web/js/components/layer/product-picker/browse/browse-layers.js
jwhurley1: Resolved in: https://github.com/nasa-gibs/worldview/pull/3783/commits/50b1ef41dfb4c212ed2a1b37313a18b937f25ad6

web/js/components/layer/product-picker/browse/browse-layers.js
jwhurley1: Resolved in https://github.com/nasa-gibs/worldview/pull/3783/commits/31bbe83ca76fa35171306e3253d9c8efcccb6a87

jwhurley1:
Addressed the additional content pushing outside of the container in https://github.com/nasa-gibs/worldview/pull/3783/commits/902610ac1b2ffb1ab7b340b79308204346c38605

web/js/components/layer/product-picker/browse/browse-layers.js
```diff
@@ -263,9 +269,11 @@ function mapStateToProps(state) {
   return {
     browser,
     mode,
+    selectedCategoryName: category && category.title,
     categoryType,
     categoryTabNames: config.categoryGroupOrder,
     measurementConfig: config.measurements,
+    lastAction,
```
jasontk19, line 276: Shouldn't need `lastAction` anymore

jwhurley1: Addressed in this commit: https://github.com/nasa-gibs/worldview/pull/3783/commits/524bffbe94b6f200d389a78034f31ed7e6bfcfec

jasontk19 (approved):

PR #3784, opened 2021-12-17
WV-1959: Remove format property from a layer config

jasontk19:
## Description

Fixes an issue where this layer's format showed as "png" in the GIBS docs when it should be shown as vector.


minniewong (approved):
👍

PR #3786, opened 2021-12-20
WV 2001 - Enable descriptions and sidebar settings on mobile

jwhurley1:
## Description
In mobile view, there is no way to access layer options and descriptions in the sidebar.

### Fixes
 - Added DropDown, DropDownMenu, etc. in imports from reactstrap 
 - Moved const variables inside renderControls() to global scope so that they can be used by renderControls() and renderDropdownMenu() 
 - Added toggleDropdownMenuVisible() function.  This function is used to toggle the visibility of the Dropdown Menu seen in mobile view.
 - Added renderDropdownMenu() function. This is function returns the text-based "buttons" for removing a layer, viewing layer options, and layer descriptions.
 - Updated renderControls() to include a conditional to render the mobile dropdown.  This function is used to display the controls for each layer, whether they are mobile or desktop view.  Shows the icon-based menu in desktop and the drop-down in mobile view.
 - removeLayerBtn in renderControls() will have 'hidden' class in mobile.  Before this commit, the "remove layer" was accessible in mobile.  This has been updated to be hidden in mobile view as the remove layer feature is included in the dropdown menu now.

### Reasoning
For mobile view, the solution of having an ellipses addresses the issue of overcrowding within a small real estate.  Display the three different buttons in mobile view will impede on the name of the layer.  When the ellipses is tapped, a dropdown menu is shown that appears in a separate "plane" so that it does not interfere with the layer's name.

## How To Test

Either view on a mobile device or shrink the browser window to a smaller size until the mobile CSS kicks in.  Tap the vertical ellipses icon that appears to the right of every layer in the sidebar.   A dropdown will popup with three different choices: 
 - Remove Layer - Should remove the layer from the list.  
 - Layer Options - opens a modal with various options for that layer
 - Layer Description - opens a modal with a description about the given layer

### DIFF
--- a/web/js/containers/sidebar/layer-row.js
+++ b/web/js/containers/sidebar/layer-row.js
@@ -8,7 +8,7 @@ import googleTagManager from 'googleTagManager';
 import { connect } from 'react-redux';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import {
-  UncontrolledTooltip,
+  UncontrolledTooltip, Dropdown, DropdownToggle, DropdownMenu, DropdownItem,
 } from 'reactstrap';
 import PaletteLegend from '../../components/sidebar/paletteLegend';
 import util from '../../util/util';
@@ -78,7 +78,25 @@ function LayerRow (props) {
   } = props;
 
   const encodedLayerId = util.encodeId(layer.id);
+  const { title } = names;
+  const removeLayerBtnId = `close-${compareState}${encodedLayerId}`;
+  const removeLayerBtnTitle = 'Remove layer';
+
+  const layerOptionsBtnId = `layer-options-btn-${encodedLayerId}`;
+  const layerOptionsBtnTitle = 'View options';
+
+  const layerInfoBtnId = `layer-info-btn-${encodedLayerId}`;
+  const layerInfoBtnTitle = 'View description';
   const [showButtons, toggleShowButtons] = useState(isMobile);
+  const [showDropdownBtn, setDropdownBtnVisible] = useState(false);
+  const [showDropdownMenu, setDropdownMenuVisible] = useState(false);
+
+  const toggleDropdownMenuVisible = () => {
+    if (showDropdownMenu) {
+      setDropdownBtnVisible(false);
+    }
+    setDropdownMenuVisible(!showDropdownMenu);
+  };
 
   const getPaletteLegend = () => {
     if (!lodashIsEmpty(renderedPalette)) {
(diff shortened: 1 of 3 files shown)


minniewong (commented):
That was really fast! I think the kabob menu  is a nice way to display those things. I wonder if it's better to order it Options, Description, then Remove is last, as I accidentally removed a layer as soon as I opened the menu. Also, there seems to be a weird space above the description. I think it would be better to align the text to the top.

web/js/containers/sidebar/layer-row.js
```diff
@@ -78,7 +78,25 @@ function LayerRow (props) {
   } = props;
 
   const encodedLayerId = util.encodeId(layer.id);
+  const { title } = names;
+  const removeLayerBtnId = `close-${compareState}${encodedLayerId}`;
+  const removeLayerBtnTitle = 'Remove layer';
```
jasontk19, line 83: I'd suggest we use title case for these, e.g. "Remove Layer", "View Options", "View Description"

web/js/containers/sidebar/layer-row.js
```diff
@@ -149,58 +167,68 @@ function LayerRow (props) {
...
+  const renderDropdownMenu = () => (
+    <Dropdown className="layer-group-more-options" isOpen={showDropdownMenu} toggle={toggleDropdownMenuVisible}>
+      <DropdownToggle>
+        <FontAwesomeIcon
+          className="layer-group-more"
+          icon="ellipsis-v"
+        />
+      </DropdownToggle>
+      <DropdownMenu positionFixed>
+        <DropdownItem id={removeLayerBtnId} onClick={() => onRemoveClick(layer.id)}>
```
jasontk19, line 179: Interacting with this a bit on my phone, I think we could make these dropdown menu items larger across the board for easier interaction.

jasontk19 (changes requested):
* We should make the options modal take up the full viewport when open:
Screen Shot 2021-12-21 at 9 48 43 AM

* The description modal has some extra space above the scrollable element that contains the text.  This looks more exaggerated here than it does on my device but the idea is the same.  The scrollable area should fill the modal body.
Screen Shot 2021-12-21 at 9 48 57 AM

web/js/containers/sidebar/layer-row.js
jasontk19: Also, for ordering the menu items I'd suggest the same order they show left-to-right at desktop size which would be: "View Description", "View Options", "Remove Layer"

jasontk19:
Also, you may already be aware but there are a few e2e tests that are broken on this branch now.

web/js/containers/sidebar/layer-row.js
jwhurley1: Addressed in c041aca3c

minniewong (commented):
Testing on my phone, I'm finding it hard to click on the blue pointing finger for the Fires (and other vector) layers. can we increase the click target for that? 

Screen Shot 2021-12-21 at 2 07 20 PM

minniewong:
I am seeing some error messages when testing on my computer
Screen Shot 2021-12-21 at 2 15 32 PM

minniewong (approved):
Looks really good!

jasontk19:
The description still gets a bit cut off for me.  Not this bad on my actual device but I still see the unused area at the bottom like this:
Screen Shot 2021-12-22 at 5 44 14 PM

jwhurley1:
Updated sidebar.css in https://github.com/nasa-gibs/worldview/pull/3786/commits/06f853ba6ef52869c4a051ffe9322849f5e8c1a1 to address the layer description not filling the modal in mobile view.

jasontk19 (approved):
Looks great!

PR #3787, opened 2021-12-22
WV-2080: Display QA datasets meaningfully

jasontk19:
## Description

* Show quality flag for collections that have it, to allow distinguishing QA datasets from others

Here is a screenshot of what this looks like with the QA datasets included:
Screen Shot 2021-12-22 at 3 11 47 PM


minniewong (approved):
I think this looks good!

PR #3789, opened 2022-01-03
WV-2098 "Add layers" issue in Android 11 when in portrait mode.

jwhurley1:
# Description
The original issue was noted as a bug in Android, but after reviewing more, it's an issue with screens that have a short height.  The original bug was noticed when the "Add Layer" button was click and the user tapped the "Search" field.  When this happens, the keyboard pops up and the modal that contains the list of layers disappears.

# Fixes
In `web/js/containers/modal.js` there was an expression checking if the orientation of the device changed.  If the orientation changed, the `toggleFunction()` was called and the modal closes.  This function was triggered when the keyboard popped up after tapping into the search input field.  Having the keyboard pop up caused the viewport size to switch from portrait to landscape and, thus, closing the modal.

It's possible there was a purpose for this behavior (closing on orientation changed) but none come to mind.

# Testing
- This can be tested in a browser by shrinking the window down until the mobile view engages.  
- Make sure the window is in a portrait-type ration.  
- Click the "Add Layer" button, tap the "Search" box, then resize the window so that it's ratio is landscape to imitate a keyboard opening on Android.  (Or you can open this on an Android phone and follow the steps indicated above.) 
- Before pulling down this branch, the bug will occur and the modal will close.  After pulling down this branch, following these steps will prevent the modal from closing.

### DIFF
--- a/web/js/containers/modal.js
+++ b/web/js/containers/modal.js
@@ -42,7 +42,6 @@ class ModalContainer extends Component {
       id,
       isOpen,
       customProps,
-      orientation,
       isMobile,
       screenHeight,
       screenWidth,
--- a/web/js/containers/modal.js
+++ b/web/js/containers/modal.js
@@ -55,12 +55,11 @@ class ModalContainer extends Component {
       desktopOnly,
     } = newProps;
 
-    const orientationChanged = orientation !== prevProps.orientation;
     const screenHeightChanged = screenHeight !== prevProps.screenHeight;
     const screenWidthChanged = screenWidth !== prevProps.screenWidth;
     const toggleFunction = toggleWithClose(onToggle, onClose, isOpen);
     if (isMobile && isOpen) {
-      if (desktopOnly || orientationChanged) {
+      if (desktopOnly) {
         toggleFunction();
       }
       if (customProps.mobileFullScreen && (screenHeightChanged || screenWidthChanged)) {


minniewong (approved):
Seems to work well! Let's see if @jasontk19 has any feedback

jasontk19:
This does fix the issue described (tested on a PIxel 2 in browserstack) but I am not clear on either a) why the `toggleFunction` was being called on an orientation change or b) why the `orientationChanged` value was `true` when the screen is small.  I'd like to poke around more before we merge this.

jasontk19:
I found the PR where the orientation change update was added: #3724 

For vector modals, it seems like changing the orientation causes the modal sizing to go haywire and the bottom goes off-screen.  I'm not convinced we need to close all modals on an orientation change (which I believe is what we are currently doing) but, we should do so with vector modals or otherwise find a way to prevent the sizing issue.

To see the issue:  With a mobile size viewport, [load any fires layer and zoom in far enough](http://localhost:3000/?v=-69.26065952079043,8.449645534311026,-65.96042912126285,10.305178631719832&l=VIIRS_NOAA20_Thermal_Anomalies_375m_All,Reference_Labels_15m(hidden),Reference_Features_15m(hidden),Coastlines_15m,VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2022-01-05-T15%3A37%3A00Z) that you can tap on a point to bring up a modal.

jwhurley1:
> 

Sure.  If you have some time tomorrow (1/7) that would be great!

jwhurley1:
Spoke with @jasontk19 about this.  We decided for this ticket, we should go ahead and keep these changes since this "Android" bug is a lot more visible than the vector modals.  We can address the vector modals again in another ticket.

Once I get approval from @jasontk19, I can merge this into `release-3.16.0`.

jasontk19 (approved):

PR #3793, opened 2022-01-05
WV-2118

jasontk19:
## Description

- don't break when no events returned
- don't show no events msg when API call fails

## How To Test

1. Load the events tab
2. Confirm app doesn't go to blank screen


jasontk19:
Note: You will still get the fullscreen error when running the app locally in dev mode due to the request failing.  This is fine, I think we want this so it's easier to spot failed API calls.  Users will never see this.

PR #3797, opened 2022-01-14
WV-2073: Animation Loading bar

jasontk19:
## Description
- Reduce initial buffer size to 10
- Show loading bar when animation is pre-loading

## To-Do

- [x] Change target branch to `release-3.17.0` before merge

## To Test

* Try animations of varying length
* Observe loading bar shows while tiles are pre-loading


minniewong (approved):
🚀

PR #3798, opened 2022-01-14
WV-2084 Allow override of About config

jwhurley1:
## Description
A user asked to make the About be overridden so they can change the text when spinning up their own copy.  The user asked to have each section in the About page to be overridden separately.

## Fixes
The only way to edit the About page was to edit the `about-page.js` itself.  In an effort to leave the default descriptions, separate JS files for each About section was created.  These individual JS pages hold the custom text and are named for their respective sections.

Once the custom text is filled out, they need to be activated by changing the respective toggles from `false` to `true`.

```javascript
// "Welcome" section will be overridden with the 
// description in the "welcome-custom.js" file
const overrideWelcome = true; 

// "Keyboard Shortcuts" section will use the 
// default description set in the "about-page.js"
const overrideKeyboardShortcuts = false; 
```
## How To Test
After pulling down the branch, run the following commands to get started:
```
npm ci
npm run build
npm run watch:debug:devtools  // This allows hot reloading on file save
```
1. Open `about-page.js` and any of the `[section]-custom.js` files. Example: `welcome-custom.js`
2. Make some changes to the text in `welcome-custom.js`. 
3. In `about-page.js`, change the `overrideWelcome` variable to `true`.
4. Open [http://localhost:3000](http://localhost:3000) in your browser.
5. Close the welcome modal, click on the Information button ("i" button) at the top right of the page and click "About".
6. You should _only_ see the description/text that you entered in `welcome-custom.js` file.  The original text defined in `about-page.js` should not be there.


jasontk19:
I love how thorough your PR descriptions have been, I can't tell you how happy that makes me as a reviewer.

I think this approach is ok but, I'm not sure having more javascript files to overwrite existing ones is much different from asking users who are running their own instance to just overwrite the existing about page javascript file.  I think this is on the right track though with breaking it up into replaceable sections.  I wonder if there is a way we could pull from optional markdown files which, if they existed, could override the different sections?  We already convert layer description markdown into HTML during the build process (see `/worldview/tasks/config.js`).  Either way, we will want to add documentation to explain how to do this.

jwhurley1:
> I love how thorough your PR descriptions have been, I can't tell you how happy that makes me as a reviewer.

Thanks!

> I think this approach is ok but, I'm not sure having more javascript files to overwrite existing ones is much different from asking users who are running their own instance to just overwrite the existing about page javascript file. 

I definitely agree but the user did request to have a way to keep the existing about page and override it in someway.  

> I think this is on the right track though with breaking it up into replaceable sections. I wonder if there is a way we could pull from optional markdown files which, if they existed, could override the different sections? We already convert layer description markdown into HTML during the build process (see `/worldview/tasks/config.js`). Either way, we will want to add documentation to explain how to do this.

Gotcha!  Converting markdown into HTML would much easier for the end user and think this is a good solution.  I will check this out.

jwhurley1:
# Description of https://github.com/nasa-gibs/worldview/pull/3798/commits/8ccba83a67af40248b979aa236fecde38da7adcd 
- Went with the markdown file approach.  The `about-page.js` now includes the new `about-section.js` component. 
- The `about-section.js` component accepts one prop: `section` which is a required string.  Whatever string is entered, that component looks for a matching HTML file in `[web-root]/brand/about/`.  So if the `section` value is:  `<AboutSection section="hello" />` it will pull in the `brand/about/hello.html`.  If the file does not exist, "No description available for this section." is displayed.
- The default sections are now stored as HTML files in `config/default/common/about`.  The user can now do one of the following to edit the About page:
  1. Overwrite the HTML files `config/default/common/about`.  The next time `npm run build` is executed, the files in this directory will copied to the `web/about` directory.
  2. Create .MD files in `config/default/common/about` which will become the HTML sections when `npm run build` is executed.  This is a non-destructive approach in that the default HTML files in `config/default/common/about` are not modified.  The HTML code in the .MD file will replace what's written to `[web-root]/brand/about` at build time.  If the user decides to go back to the default descriptions, they can just delete their .MD files.
- The currently available sections are:
  - `welcome` (`welcome.html` or create `welcome.md` to overwrite)
  - `keyboard` (`keyboard.html` or create `keyboard.md` to overwrite)
  - `imagery` (etc....)
  - `acknowledgements` 
  - `disclaimer` 
  - `license`
- To add more sections, add a `<AboutSection section="[section name]" />` in `about-page.js`

jasontk19:
I like this new approach.  I wonder though if we want to commit only the markdown files to the repo and have the HTML generated that gets generated on build only copied over to the build folder like we do with layer metadata?

minniewong (commented):
This works well! It would be good to update the documentation as well so others will know how to make edits. Also, should the version number still be updated with @BUILD_VERSION@?

config/default/common/brand/about/welcome.md
```diff
@@ -0,0 +1,31 @@
+<h1>Welcome to Worldview<span class="version"> Version 3.16.0 (<a href="https://github.com/nasa-gibs/worldview/releases"
```
jasontk19: As Minnie pointed out, this version shouldn't be hard-coded

web/js/components/about/about-section.js
```diff
@@ -0,0 +1,41 @@
...
+    (async () => {
+      if (!sectionName) {
+        setFn('no section name provided');
+        return;
+      }
+      try {
+        const data = await fetch(`brand/about/${sectionName}.html`);
+        const dataHTML = await data.text();
+        controller = null;
+        const setData = data.ok ? dataHTML : 'No description available for this section.';
```
jasontk19, line 20: I tried updating the "Disclaimer" section and when I did it appears to erroneously show the "No description available for this section" message:
Screen Shot 2022-01-25 at 12 06 05 PM

jasontk19 (changes requested):

config/default/common/brand/about/welcome.md
jwhurley1: I think we crossed paths when you made this comment.   Addressed in https://github.com/nasa-gibs/worldview/pull/3798/commits/34e9153063ac73d63d221078c62c00ec79365b52

web/js/components/about/about-section.js
jwhurley1: Addressed in https://github.com/nasa-gibs/worldview/pull/3798/commits/2d61241e85c5d4b24aabbca4a2b35cc16b86f0fe and https://github.com/nasa-gibs/worldview/pull/3798/commits/3e8c87c069af1cd80a805225f5338349653f7cd7

jasontk19: Related to this... if someone didn't want to have all sections, or all the _same_ sections as we use, they would still have to either include something in the markdown file with that name or modify the JavaScript right? Otherwise they would get this "not found" message.  We don't have to make this changes for this PR but in the future it might make more sense to just read in all the markdown from the config folder and insert it in order (you'd perhaps need to prefix each markdown file with sequential numbers to get the order you want).

Alternatively, maybe it doesn't need to display a "not found" message?  Or just to the console?  Up to you.

jwhurley1: I like the numbering idea.  That would be good because it keeps the control of the about sections solely on the markdown files.

And yes, I agree that it doesn't need to display a "not found" message.  I'll remove that and push up the changes.

jasontk19 (approved):
👍🏻 +1

minniewong (approved):
🎸

PR #3799, opened 2022-01-19
WV-2152: Disable layer controls during animation

jasontk19:
## Description

* Prevents layers/groups from being modified during animation playback since this can essentially break an in-progress animation.
* Prevent showing/hiding a layer that is "disabled" (no imagery at date)

## How To Test

1. [Load the app via this URL](http://localhost:3000/?v=-136.93290062204383,-67.32407163823656,-81.02150667305224,-12.521867723501241&z=4&ics=true&ici=5&icd=10&as=2022-01-17-T03%3A00%3A00Z&ae=2022-01-17-T06%3A00%3A00Z&l=Coastlines_15m,GOES-East_ABI_Band13_Clean_Infrared,MODIS_Aqua_CorrectedReflectance_TrueColor(hidden)&lg=true&al=true&av=1&ab=on&t=2022-01-17-T04%3A20%3A00Z)
2. Play the animation
3. Confirm that layers cannot be shown/hidden via eye icon
4. Confirm that layer options are not accessible
5. Confirm that layers cannot be re-ordered
6. Confirm that layer groups cannot be re-ordered

### DIFF
--- a/web/css/sidebar-panel.css
+++ b/web/css/sidebar-panel.css
@@ -121,7 +121,6 @@ a.atexpand {
 
 .sidebar-panel li.item {
   padding: 0;
-  cursor: move;
   margin: 0 auto;
 }
 
@@ -204,7 +203,7 @@ img.sidebar-panel-item-img {
   display: block;
 }
 
-img.hideReg {
+.visibility img {
   height: 14px;
   width: 20px;
 }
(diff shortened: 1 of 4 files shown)


jwhurley1:
Looks good!  Tested on Win10 Chrome, Firefox, and Edge and works as described.  👍

PR #3801, opened 2022-01-26
Progressive Web App

jwhurley1:
## Description
First pass at making Worldview a PWA.  This will make Worldview "installable" on mobile and desktop and make it appear as a native, full screen application.

This PR is focused on two files that are required to have a website pass as a PWA and make it installable: 
- `manifest.json` :  A file containing meta data about the app, icons, theme color, descriptions, and other necessary resources.
- `service-worker.js` or `sw.js`: A file that works as a proxy between the app and internet to make assets available when the app is offline.

_NOTE: Being that the Worldview app relies on the internet to download images from GIBS and OSM, not much can be done for working in "offline mode" other than caching the app's icons and local assets. This may change in future updates._

In addition to the`sw.js` file's caching being very useful, it is required for PWAs.  There is a tool that helps creating boilerplate and best practices for service workers, and that is [WorkBox](https://developers.google.com/web/tools/workbox).  Workbox has a CLI tool, which will be installed as a devDependency when running `npm ci`.  Workbox CLI can create a customized `sw.js` file by modifying the configuration settings made in `workbox-config.js`.  An `sw.js` file can be made by running `npx workbox generateSW workbox-config.js`.  

## Manifest.json
Here are some quick descriptions of some of the parameters that are less self-descriptive.

`start_url` and `scope` are just setting at what directory the service worker should start caching.  I just selected the root level here because there is very little routing, if any at all in Worldview.
```json
"start_url": ".",
"scope":".",
  ```
This controls how the installed PWA will look. `"display": "standalone"` makes the app appear as if it were not in a browser.  The`background_color` is a gray and is used for the splash screen that appears when launching the installed PWA.  The `theme_color` is currently set as the blue from the NASA logo and is the app's menu bar color.
```json
"background_color": "#212529",
"display": "standalone",
"theme_color": "#0b3d91",
```

## How To Test on Desktop
1. Clone this repo and run ` git checkout feature-pwa`.
2. Run the following command:
 `npm ci && npm run build && npm start`
3. Open [localhost:3000](http://localhost:3000) in your browser.

From here, you can either install the PWA or check to see if it passes the "Lighthouse" audit feature. To install, do the following:
- In the address bar, there should be a new icon that looks like an arrow pointing down into a box.  Click that icon to install and confirm installation.
- A new icon should appear on your desktop that will launch a PWA version of Worldview.

To test via Lighthouse Audit, open the Devtools by:
- Pressing `F12` or `CTRL/CMD + SHIFT + i`.
- Clicking the "Lighthouse" tab in the Dev Tools.
- Uncheck all boxes except "Progressive Web App" and click "Generate Audit Report"
- The results should return saying the application is installable.

## How To Test on Mobile
1. Follow "How To Test On Desktop" through step 3. 
2. Find out the desktop's IP address and make sure your mobile device is connected to the same network as the desktop.
3. On mobile device, open a browser and go to:  `[desktop's IP]:3000`
_NOTE: If your request times out, your desktop may have firewall restrictions in place that block incoming connections on port 3000. Please refer to your system administrator for assistance if this is the case._
4. In your mobile browser, access where there are more options and find "Add to Homescreen".
5. This will add Worldview to your home screen which you can then tap to open as a full screen application.


workbox-config.js
```diff
@@ -0,0 +1,29 @@
...
+  ],
+  swDest: 'web/sw.js',
+  ignoreURLParametersMatching: [
+    /^utm_/,
+    /^fbclid$/,
+  ],
+  // Define runtime caching rules.
+  runtimeCaching: [{
+    // Match any request that ends with .png, .jpg, .jpeg or .svg.
+    urlPattern: /\.(?:png|jpg|jpeg|svg)$/,
```
jasontk19, line 14: I'd like to understand the caching aspect better, I'm guessing this only applies to assets that live within the deployed app dirs somewhere and not to any requests for assets we make to external APIs (e.g. GIBS) during runtime?

jwhurley1: That's correct.  If you want to check out what cache is built, you can open Dev Tools > Application > Cache Storage (left menu) > Workbox-precache-v2.   There you will see all of the entries that are cached by the service worker.  As the app is being used and other layers are added and viewed, the are not added to the cache.

However, I will say that the cache that is created is fairly large (102 MB).  We could exclude directories if needed.

Here is some information about precaching with Workbox:  https://developers.google.com/web/tools/workbox/modules/workbox-precaching

Here is where I pulled some code for the `workbox-config.js`  https://developers.google.com/web/tools/workbox/guides/generate-service-worker/cli#adding_runtime_caching

jasontk19 (changes requested):
Ok here's a summary of some of my findings so far:

- [ ] it looks like the app will precache every image, even when I just open the app in a browser, rather than just when I install it as a PWA which we definitely don't want since there are so many images (mostly layer previews)
- [ ] related, it also preached a bunch of un-combined and un-minified javascript files which we don't want
- [ ] we need to find a way to have the workbox script run when the app is built for deployment (e.g. the `tasks/dist.js` script is run) but we probably don't want it to happen on every build locally since there isn't really a need
- [ ] the `sw.js` is picking up references to every layer metadata and html file that it gets converted to, we should modify the conversion script to remove the `*.md` files from the copied directory after the conversion happens.  This work can be done in another ticket/PR but needs to happen before this feature branch gets merged.

PR #3803, opened 2022-01-31
WV-2130: Implement Service Worker

jwhurley1:
## Description

This PR is to greatly simplify the PWA Service Worker to meet the base requirements, and implement it into the Progressive Web App feature.  In creating this simple service worker, I started by just having a blank JS file and using that as a service worker.  I ran the Lighthouse Audit for a PWA and was told the app was not installable because it was "Not available offline."  To get around this, I created a simple event listener for the `fetch` event, which has a callback that does nothing.  Just creating an event handler for the fetch event is enough to pass the Lighthouse Audit, it seems.  If we were to create a static offline page in the future, it would be handled in this callback.

## How To Test on Desktop
1. Clone this repo and run ` git checkout wv-2130`.
2. Run the following command:
 `npm ci && npm run build && npm start`
3. Open [localhost:3000](http://localhost:3000) in your browser.

From here, you can either install the PWA or check to see if it passes the "Lighthouse" audit feature. To install, do the following:
- In the address bar, there should be a new icon that looks like an arrow pointing down into a box.  Click that icon to install and confirm installation.
- A new icon should appear on your desktop that will launch a PWA version of Worldview.

To test via Lighthouse Audit, open the Devtools by:
- Pressing `F12` or `CTRL/CMD + SHIFT + i`.
- Clicking the "Lighthouse" tab in the Dev Tools.
- Uncheck all boxes except "Progressive Web App" and click "Generate Audit Report"
- The results should return saying the application is installable.

## How To Test on Mobile
1. Follow "How To Test On Desktop" through step 3. 
2. Find out the desktop's IP address and make sure your mobile device is connected to the same network as the desktop.
3. On mobile device, open a browser and go to:  `[desktop's IP]:3000`
_NOTE: If your request times out, your desktop may have firewall restrictions in place that block incoming connections on port 3000. Please refer to your system administrator for assistance if this is the case._
4. In your mobile browser, access where there are more options and find "Add to Homescreen".
5. This will add Worldview to your home screen which you can then tap to open as a full screen application.


web/service-worker.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable no-restricted-globals */
+self.addEventListener('fetch', (event) => {
+  // Ignore
```
jasontk19, line 3: So, can we not allow "installing" the app without this totally pointless event listener?  Seems weird to have to have something that does nothing.  If we are just trying to please the Lighthouse test I think we can skip it.  I guess I need to look into it more.

jasontk19: Also still unclear on if there is any point or advantage to having it cache our main, compiled `wv.js` file or not.  If there is, I suppose that could be something to do here.

jwhurley1: > So, can we not allow "installing" the app without this totally pointless event listener? Seems weird to have to have something that does nothing. If we are just trying to please the Lighthouse test I think we can skip it. I guess I need to look into it more.

When I tested locally, I removed the `<script>` tag from the index.html that's calls the service worker file, but left the manifest `<link...>` tag in there to see if we only need the manifest file.  I was not able to install on desktop nor did I get a splash screen or full screen app.  Reenabled the service worker and everything worked again.  Chrome/chromium browsers seem to expect a service worker of some sort, even if it's doing nothing.

Here are a couple of pages regarding minimal PWA requirements.  Essentially what we're doing here is a "Network-Only" approaching to caching.  In other words, we're not caching at all, but by declaring that we are making a decision on how we want to cache things.
- https://mobiforge.com/design-development/pwa-minimus-a-minimal-pwa-checklist
- https://stackoverflow.com/questions/61208907/pwa-minimal-service-worker-just-for-trigger-the-install-button
- https://developers.google.com/web/ilt/pwa/caching-files-with-service-worker#network_only

> Also still unclear on if there is any point or advantage to having it cache our main, compiled `wv.js` file or not. If there is, I suppose that could be something to do here.

Def worth a shot I think.  If someone visits Worldview and/or installs the app, the `wv.js` file won't need to be downloaded again on subsequent visits or page reloads.

web/index.html
```diff
@@ -63,6 +63,16 @@
 
   <link type="text/css" rel="Stylesheet" href="build/wv.css?v=@BUILD_NONCE@">
   <script type="text/javascript" src="build/wv.js?v=@BUILD_NONCE@"></script>
+
+  <link rel="manifest" href="/manifest.webmanifest">
+  <meta name="theme-color" content="#0b3d91"/>
+  <script>
```
jasontk19, line 69: Does this need to be in a <script> tag or can we just include this JavaScript in our `main.js` file?

web/service-worker.js
jasontk19: More thoughts on this... I think it would be preferable not to attach an event listener to `fetch` events since this will  needlessly fire for every network request.  If we have to register for _some_ event perhaps the `install` event would be better since it would happen only once?

web/index.html
jwhurley1: We could try putting it in the `main.js`, but I think we'll need to make a change to the code that registers the service worker's scope, i.e. the highest directory from which it's working from despite the location worker itself.

```javascript
navigator.serviceWorker.register('js/service-worker.js', {
  scope: '/'
});
```

Taken and modified from this page:  https://developers.google.com/web/ilt/pwa/introduction-to-service-worker#registration_and_scope.  

Also:
> If you want the service worker to control higher pages e.g. /app (without the trailing slash) you can indeed change the scope option, but you'll also need to set the `Service-Worker-Allowed` HTTP Header in your server config for the request serving the service worker script.

I don't know if any of this is really necessary if we're not planning on caching anything, but if we do decide to cache the `wv.js` or anything else in a directory above the service worker, we'll may need to keep this all in mind.

jasontk19: There is no `js` folder in the final build directory structure, would it not just live in `web/js` alongside `main.js` and the pathing would be the same?

web/service-worker.js
jasontk19: Update:  It looks like this is actually one of the install criteria: `Registers a service worker with a fetch handler`
https://web.dev/install-criteria/

web/manifest.webmanifest
```diff
@@ -0,0 +1,30 @@
+{
+  "short_name": "@NAME@",
```
jasontk19, line 2: I don't think these variable names are going to get picked up because the dist script doesn't look at this location.

web/manifest.webmanifest
```diff
@@ -1,6 +1,6 @@
 {
-  "short_name": "@NAME@",
-  "name": "@LONG_NAME@",
+  "short_name": "Worldview",
```
jasontk19, line 2: I think we probably want the variables we just need to make sure they get translated properly during build

jwhurley1: Ok, made an update to `tasks/dist.js` to include  `build/worldview/manifest.webmanifest`

web/service-worker.js
jwhurley1: Circling back to this -- So just to pass the install criteria, we should keep the service worker, as is?  At least for now?

web/manifest.webmanifest
jasontk19: Looks good, just need to put the vars back into the manifest

jwhurley1: Just pushed it up!

web/service-worker.js
jasontk19: Yep no need for changes to this

jasontk19 (approved):
I think this is good to go for now, I am still not sure why Chrome on iOS doesn't present me with an option to install but can look into that later since this is going into a feature branch for now.

jasontk19:
Oh and please be sure to squash and merge on this one due to the large number of commits.  Thanks!

PR #3805, opened 2022-01-31
WV-2143: Animation on mobile

jasontk19:
## Description

* Allows playing an animation on mobile if user arrives via a URL that has an animation configured
* Shows an alert warning the user that an animation may use a notable amount of data
* Will not show the play button on mobile if the number of frames exceeds the max (which would cause the play button to be disabled) since there is no way to change the start/end datetimes.

## How To Test

* Load any URL with an animation, use dev tools to simulate mobile size
* [Here is a test animation URL](http://localhost:3000/?v=-43.785036915393555,-8.73955528138297,-38.52479650089625,2.6441444874265896&z=4&ics=true&ici=5&icd=10&as=2022-01-22-T14%3A27%3A09Z&ae=2022-01-22-T14%3A27%3A09Z&l=GOES-East_ABI_GeoColor,Reference_Labels_15m(hidden),Reference_Features_15m(hidden),Coastlines_15m,VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&ab=on&t=2022-02-01-T14%3A27%3A09Z)
* Try on a real device by using the GIBS UAT URL provided in the original ticket comments

### DIFF
--- a/web/css/alert.css
+++ b/web/css/alert.css
@@ -1,22 +1,34 @@
 .wv-alert-container {
   position: absolute;
-  top: 60px !important;
-  right: 10px !important;
-  left: auto !important;
+  top: 60px;
+  right: 10px;
+  left: auto;
 }
+
 .wv-alert {
+  position: relative;
+  color: #fff;
+  background: #282828;
+  border: #282828;
+  border-radius: 0;
   display: flex;
   align-items: center;
   justify-content: center;
   min-height: 40px;
   cursor: pointer;
   padding: 0;
-  border: none;
   z-index: 100;
   margin-bottom: 5px;
 }
+.alert-content {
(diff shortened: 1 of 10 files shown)


rboller (changes requested):
Great to see this in action and I've added a few suggestions on the internal ticket for discussion.

minniewong (commented):
I have also added some comments on the other ticket

minniewong (approved):
Looks great! 👍

PR #3806, opened 2022-02-01
wv-2059-Update versions for VIIRS

minniewong:
## Description

Fixes WV 2059 and WV-2060 .

Update Suomi NPP/VIIRS NRT layer description dataset DOI to v2; update NOAA-20/VIIRS NRT layer descrption dataset DOIs to v2.1.


jasontk19 (approved):
+1

PR #3807, opened 2022-02-01
WV-2169: Branding version fix

jasontk19:
## Description

- Make sure the brand/about folder gets picked up by branding in dist.js
- Remove markdown files from build folders after the HTML gets generated to reduce number of files deployed significantly


jwhurley1:
Looks good!

jwhurley1 (approved):

PR #3808, opened 2022-02-03
Point to new GIBS API Docs

minniewong:
## Description

Updated URLs to point to new GIBS API docs located at nasa-gibs.github.io/gibs-api-docs/


jasontk19 (approved):
:shipit:

PR #3809, opened 2022-02-04
WV-2168: Refine animation calculation to prevent stuttering

jasontk19:
## Description

The primary issue that would still cause stuttering was the case where the user had already loaded some of the steps within the pre-buffering range of 10 frames which we make to determine an average request time on which to base the calculation for how many _additional_ frames to request before beginning playback.  

To reproduce the original issue:
NOTE: You may need to throttle your connection slightly to reproduce since we were falling back to some fixed estimated load times in cases where an average could not be calculated.
1. [load this URL](https://worldview.earthdata.nasa.gov/?as=2021-05-20-T00%3A00%3A00Z&ae=2022-01-30-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2021-09-06-T14%3A00%3A00Z)  
2. Step backwards 10 times (this loads the frames you're about to play back)
3. Start playback
4. At some point, if your connection speed is not fast enough, playback will catch up to the buffer and stuttering will occur.

### Solution

* Always make _at least_ 10 real requests that are not being pulled from the cache so we can have a reliable average fetch time and not rely on a guesstimated average time as a fallback like we did before.  For example, if 5 of the initial 10 steps were pulled from the cache, keep adding to the pre-buffer until 15 total frames are loaded.

## How To Test

In the following scenarios that "start in the middle", try stepping backwards 5-10 steps to pre-load frames to ensure the original issue is no longer reproducible.  The other scenarios are here to ensure nothing was broken while making these changes.

* Less than 10 frames
	* [Start in the middle](http://localhost:3000/?as=2022-01-01-T00%3A00%3A00Z&ae=2022-01-07-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2022-01-03-T16%3A00%3A00Z)
	* [Start before start date](http://localhost:3000/?as=2022-01-01-T00%3A00%3A00Z&ae=2022-01-07-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2021-12-29-T12%3A00%3A00Z)
* Exactly 10 frames
	* [Start in the middle](http://localhost:3000/?as=2022-01-21-T00%3A00%3A00Z&ae=2022-01-30-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2022-01-24-T14%3A00%3A00Z)
	* [Start before start date](http://localhost:3000/?as=2022-01-21-T00%3A00%3A00Z&ae=2022-01-30-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2022-01-16-T10%3A00%3A00Z)
* Exactly 11 frames
	* [Start in the middle](http://localhost:3000/?as=2022-01-20-T00%3A00%3A00Z&ae=2022-01-30-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2022-01-24-T10%3A00%3A00Z)
	* [Start before start date](http://localhost:3000/?as=2022-01-20-T00%3A00%3A00Z&ae=2022-01-30-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2022-01-17-T08%3A00%3A00Z)
* A lot more than 10 frames
	* [Start before start date](http://localhost:3000/?as=2021-05-20-T00%3A00%3A00Z&ae=2022-01-30-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2021-05-07-T10%3A00%3A00Z)
	* [Start in the middle](http://localhost:3000/?as=2021-05-20-T00%3A00%3A00Z&ae=2022-01-30-T00%3A00%3A00Z&al=true&av=10&ab=on&t=2021-09-06-T14%3A00%3A00Z)

### DIFF
--- a/web/js/components/animation-widget/play-queue.js
+++ b/web/js/components/animation-widget/play-queue.js
@@ -181,8 +181,8 @@ class PlayQueue extends React.Component {
   isPreloadSufficient() {
     const { numberOfFrames } = this.props;
     const currentBufferSize = util.objectLength(this.bufferObject);
-    if (this.canPreloadAll) {
-      return currentBufferSize === numberOfFrames;
+    if (currentBufferSize === numberOfFrames) {
+      return true;
     }
     if (currentBufferSize < this.defaultBufferSize) {
       return false;
@@ -312,8 +312,9 @@ class PlayQueue extends React.Component {
     if (!this.mounted) return;
     this.bufferObject[strDate] = strDate;
     delete this.inQueueObject[strDate];
+    const bufferLength = this.bufferArray.length;
 
-    if (!initialLoad || this.canPreloadAll) {
+    if (!initialLoad || this.canPreloadAll || bufferLength >= this.defaultBufferSize) {
       this.checkQueue();
       this.checkShouldPlay();
     }
--- a/web/js/components/animation-widget/play-queue.js
+++ b/web/js/components/animation-widget/play-queue.js
@@ -6,10 +6,25 @@ import { Progress } from 'reactstrap';
 import LoadingIndicator from './loading-indicator';
 import util from '../../util/util';
 
+// We assume anything this fast or faster is a frame that was pulled from the cache
+const MIN_REQUEST_TIME = 200; // milliseconds
 const CONCURRENT_REQUESTS = 3;
 const toString = (date) => util.toISOStringSeconds(date);
 const toDate = (dateString) => util.parseDateUTC(dateString);
 
+/**
+ * A component that handles buffering datetimes for animation frames.  Only mounted while playback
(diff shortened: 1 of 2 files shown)


jwhurley1 (approved):
Looks great!  There's a lot of code to take in, but looking it over definitely seems a lot cleaner and more optimized.  I tested with throttling and the stuttering was very much improved.  Nice work!

minniewong (commented):
All the tests work really well except for Exactly 11 frames tests. The "Determining Buffer Size" loading bar seems to hardly move and after waiting a few mins, the animation still doesn't load. I even tried to "preload" each date in the animation but it didn't seem to make a difference.

minniewong (approved):
All looks great now!

PR #3811, opened 2022-02-08
WV-2006: Toggle setting to always show datelines

jasontk19:
## Description

- Adds toggle to settings panel for option to always show datelines
- When datelines always show, dates initially show centered on the line.  If user hovers over the dateline, the dates move along the line to where the mouse is pointing.

## How To Test

1. Load app normally, don't change settings
2. Move to the left dateline, hover mouse over where dateline should be
3. Confirm dateline appears
4. Confirm dates appear where mouse points
5. Move to the right dateline, hover mouse over where dateline should be
6. Confirm dateline appears
7 Confirm dates appear where mouse points

---

1. Go to settings
2. Switch datelines to "show always"'
3. Pan map as necessary to see left dateline
4. Confirm dates show on left/right of the line
5. Hover mouse over the line
6. Confirm dates move to where mouse points
7. Pan map as necessary to see the right dateline
8. Repeat steps 4-6

---

1. With datelines showing, switch projections to arctic
2. Confirm no datelines visible
3. Switch projections back to geographic
4. Confirm datelines are showing


minniewong (approved):
Works really well! 🔥

jwhurley1 (approved):
Looks great!  Nice job, Jason!

PR #3817, opened 2022-02-14
WV-2181 

jwhurley1:
## Description

Fixes:  Right click then Take a snapshot causes error.  

When right clicking on the map, coordinates are set in the Redux store.  When single-clicked off of the right-click menu and not selecting anything,  coordinates were still stored.  In `web/js/map/ui.js` there is a function `subscribeToStore` that runs `addMarkerAndUpdatedStore()` when the `TOGGLE_DIALOG_VISIBLE` is called.  This action is called when the snapshot button is clicked is it trying to close the little tooltip next to the location marker.  Since a location marker doesn't exist yet, it was throwing an error.  The new boolean `isMarkerPlaced` in `locationSearch` was added to show declare whether a marker exists on the map or not making the markers independent from coordinates existing in the store.

### Test 1
1. Pull down changes and run `npm start` and open [localhost:3000](http://localhost:3000).
2. Right click on map to bring up context menu, but don't select anything in the menu.
3. Click on "Take a Snapshot"
4. No error should show.  The Snapshot menu should open with the grayed out screen and a non-grayed out selection box.

### Test 2
1. Right click on map to bring up context menu, and place a marker on the map.
1. Right click again and do NOT select anything, leaving the menu open.
1. Click on "Take a Snapshot"
1. No error should show.  The Snapshot menu should open with the grayed out screen and a non-grayed out selection box.
1. The location marker's tooltip should minimize/disappear and the marker should be by itself.


web/js/containers/map-interactions/map-interactions.js
```diff
@@ -30,13 +30,15 @@ class MapInteractions extends PureComponent {
...
-          <div id="wv-map" className={mapClasses} />
-        </ContextMenuTrigger>
         <RightClickMenu />
         <OlZoomButtons />
         <OlRotationButtons />
-        <OlCoordinates show={!isDistractionFreeModeActive} />
+        <ContextMenuTrigger id="context-menu-trigger">
+          <div id="wv-map" className={mapClasses} />
+        </ContextMenuTrigger>
+        {!isDistractionFreeModeActive && (
```
jasontk19, line 39: I think there may have been a bad conflict resolution here, it looks like you undid my changes that should read as follows: 
```
<OlCoordinates show={!isDistractionFreeModeActive} />
```

jasontk19 (changes requested):
Looks like the bug is fixed but I think there were some unintentional changes made during the rebase.

web/js/containers/map-interactions/map-interactions.js
jasontk19: Fixing this should fix the broken e2e tests as well.  You can confirm locally by doing:

1. Go to info menu
2. Enter distraction free mode
3. Click the eyeball icon top right to exit distraction free mode

Presently, this results in an error on this branch.  Should be fixed with the above change addressed.

minniewong (changes requested):
The fix works really well but like Jason noted I am seeing some other unexpected issue. The map coordinates box that's in the lower right corner is no longer there.

web/js/containers/map-interactions/map-interactions.js
jwhurley1: This worked.  Good eye catching this.  I was manually accepting a lot of rebase conflicts and they were all in favor of my local changes vs remote changes.  I ended up accepting all the local changes blindly...and this is what I get lol.

jwhurley1:
@minniewong 
> The fix works really well but like Jason noted I am seeing some other unexpected issue. The map coordinates box that's in the lower right corner is no longer there.

After making the change Jason K. mentioned, I see the map coordinates box.  (Hopefully this is the one you're talking about)
image

minniewong (approved):
yep! it's back and looks good!

jwhurley1:
I just found another bug, which I'm working on now.

1. Right click and "Add Marker"
2. Right click again and click off of the context menu.
3. Marker disappears.

PR #3820, opened 2022-02-16
WV-2081: Migrate to SCSS

jasontk19:
## Description

- Migrate all `*.css` files to `*.scss` files
- Update stylelint and bring in standard scss linting config
- Remove most custom rules we had defined in favor of using a standard
- Update a lot off styles to comply with linting rules
- Remove numerous unused or no longer necessary CSS processing related dependencies 
- Use SASS to import all `*.scss` files into `index.scss` which is imported into `app.js`
- Create a `variables.scss` file where shared variables and mixins will live
- Create and use a few variables and mixins for commonly shared styles

WV-2193
- Remove unused variants of Open Sans font
- Bring in Droid Sans Mono to replace Roboto Mono
- Change button font to use Open Sans Bold instead of Segoe UI font

WV-2187
- Change info icon to question mark icon in layer coverage panel

## How To Test

These changes touch every part of the app on a stylistic level though there should not be noticeable visual changes.  At some point before release, everything should be visually checked for any disturbances.


web/scss/components/image.scss
```diff
@@ -4,7 +4,7 @@
   min-height: 100px;
   position: absolute;
   display: none;
-  background: rgba(38, 43, 49, 0.45);
+  background: rgb(38 43 49 / 45%);
```
Benjaki2, line 7: is that just scss syntax for `rgba(38, 43, 49, 0.45);`?

Benjaki2 (approved):
Mixins and Variables looks like big enhancements! Everything looks good to me. We probably want to start only using variables for fonts/colors. 

🎉

minniewong (commented):
Did the dates in the events listing and layer description also get updated to droid sans mono? It doesn't seem like it, but I'm not sure. 
Screen Shot 2022-02-22 at 9 49 32 AM
Screen Shot 2022-02-22 at 9 50 14 AM

jasontk19:
> Did the dates in the events listing and layer description also get updated to droid sans mono? It doesn't seem like it, but I'm not sure. Screen Shot 2022-02-22 at 9 49 32 AM Screen Shot 2022-02-22 at 9 50 14 AM

As we discussed, I will do this work in a separate PR

PR #3823, opened 2022-02-17
WV-2093 Implement EONET's Magnitude field in Events

jwhurley1:
## Description

Utilize EONET's magnitude field for Severe Storms and Sea and Lake Ice to display wind speed (severe storms) and surface area (sea and lake ice) as a on hover tooltip on each relevant point. Also include the magnitude information in the Event listing.

**Hover on iceberg point**
image

**Hover on severe storm point**
image

**Sidebar display**
image
image

## How To Test

1. Pull down this branch and run `npm run build && npm start`
2. Open [localhost](http://localhost:3000)
3. Click on the "Events" tab on the sidebar menu and select an Iceberg event.
    - The magnitude data should appear next to each date in the sidebar.  Should say `Surface Area: [data]`
    - Hover over a track point on the map.  Magnitude data should appear below the date in the tooltip.
4. Click on the "Events" tab on the sidebar menu and select a severe storm event.
    - The magnitude data should appear next to each date in the sidebar. Should say `Wind Speed: [data]`
    - Hover over a track point on the map.  Magnitude data should appear below the date in the tooltip.

### DIFF
--- a/web/css/events.css
+++ b/web/css/events.css
@@ -94,7 +94,7 @@
 .map-item-list li.item p.magnitude {
   display: inline-block;
   margin-left: 5px;
-  padding: 0px;
+  padding: 0;
 }
 
 .map-item-list li.item p a {
--- a/web/css/events.css
+++ b/web/css/events.css
@@ -89,21 +89,35 @@
 
 .map-item-list li.item p {
   color: #666;
+
+  &.magnitude {
+    display: inline-block;
+    margin-left: 5px;
+    padding: 0;
+  }
+
+  a {
+    display: block;
+    clear: both;
+    margin: 4px 0 6px;
+    position: relative;
+    padding-left: 15px;
+  }
 }
 
-.map-item-list li.item p.magnitude {
-  display: inline-block;
-  margin-left: 5px;
-  padding: 0;
-}
+// .map-item-list li.item p.magnitude {
+//   display: inline-block;
+//   margin-left: 5px;
+//   padding: 0;
+// }
 
-.map-item-list li.item p a {
-  display: block;
-  clear: both;
-  margin: 4px 0 6px;
-  position: relative;
-  padding-left: 15px;
-}
+// .map-item-list li.item p a {
+//   display: block;
+//   clear: both;
+//   margin: 4px 0 6px;
+//   position: relative;
+//   padding-left: 15px;
+// }
 
 .wv-eventslist li.item ul li.dates {
   margin: 4px 0;
--- a/web/css/events.css
+++ b/web/css/events.css
@@ -91,6 +91,12 @@
   color: #666;
 }
 
+.map-item-list li.item p.magnitude {
+  display: inline-block;
+  margin-left: 5px;
+  padding: 0px;
+}
+
 .map-item-list li.item p a {
   display: block;
   clear: both;
--- a/web/css/events.css
+++ b/web/css/events.css
@@ -105,20 +105,6 @@
   }
 }
 
-// .map-item-list li.item p.magnitude {
-//   display: inline-block;
-//   margin-left: 5px;
-//   padding: 0;
-// }
-
-// .map-item-list li.item p a {
-//   display: block;
-//   clear: both;
-//   margin: 4px 0 6px;
-//   position: relative;
-//   padding-left: 15px;
-// }
-
 .wv-eventslist li.item ul li.dates {
   margin: 4px 0;
   a {
--- a/web/js/components/sidebar/event.js
+++ b/web/js/components/sidebar/event.js
@@ -63,6 +63,13 @@ function Event (props) {
         >
           {event.geometry.map((geometry, index) => {
             const date = util.toISOStringDate(geometry.date);
+            function magnitudeOutput() {
+              const magnitudeUnit = geometry.magnitudeUnit ? geometry.magnitudeUnit : null;
+              const magnitudeValue = geometry.magnitudeValue ? geometry.magnitudeValue : null;
+              if (!magnitudeUnit && !magnitudeValue) return;
+              const magnitudeDisplay = magnitudeUnit === 'kts' ? 'Wind Speed: ' : 'Surface Area: ';
+              return `${magnitudeDisplay} ${magnitudeValue} ${magnitudeUnit}`;
+            }
             return (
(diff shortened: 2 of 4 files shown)


minniewong:
Any way to change it to a superscript 2 for square nautical miles?

jwhurley1:
> Any way to change it to a superscript 2 for square nautical miles?

Yeah for sure. I'll put that in for the next push.
Noticing some inconsistencies with the magnitude values in the sidebar vs. the hover tooltip.  Going to figure that out...

jwhurley1:
> Now I wonder if it should be "sq nmi" or "sq NM" 🤔 . The text also needs to be updated in the events listing.

Googling around to find the accepted abbreviation tells me a few different answers.  I'm seeing nmi<sup>2</sup> and NM<sup>2</sup> mostly.  However, having "sq nmi" or "sq NM" would 1) make the code simpler and 2) eliminate issues with the superscripted 2's line-height invading the space of the text above it.

jasontk19 (changes requested):

web/js/components/sidebar/event.js
```diff
@@ -63,6 +63,13 @@ function Event (props) {
         >
           {event.geometry.map((geometry, index) => {
             const date = util.toISOStringDate(geometry.date);
+            function magnitudeOutput() {
+              const magnitudeUnit = geometry.magnitudeUnit ? geometry.magnitudeUnit : null;
+              const magnitudeValue = geometry.magnitudeValue ? geometry.magnitudeValue : null;
```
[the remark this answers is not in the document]
jwhurley1: Good call.  Updated util.js, as well, for the track points.

image

image

web/js/map/natural-events/util.js
```diff
@@ -15,8 +15,18 @@ export const getTrackPoint = function(proj, clusterPoint, isSelected, callback)
...
-  const { properties } = clusterPoint;
+  const { properties, geometry } = clusterPoint;
   const content = document.createTextNode(formatDisplayDate(properties.date));
+
+  const { magnitudeValue, magnitudeUnit } = geometry;
+  const magnitudeContent = document.createElement('div');
+  const superscriptUnit = document.createElement('sup');
+  superscriptUnit.append('2');
+  const formattedMagnitudeUnit = magnitudeUnit === 'kts' ? ' kts' : ' NM';
+  magnitudeContent.append(magnitudeValue.toLocaleString());
```
jasontk19, line 26: Be careful calling this directly on something we got from an API since it assumes that there is a value there.  If there is not, we get an error.  For example, [this event seems to cause an issue: ](http://localhost:3000/?v=-44.0328515625,-66.98218750000001,-28.6871484375,-43.4978125&e=EONET_5210,2021-01-15&efs=true&efd=2016-10-21,2022-02-18&efc=seaLakeIce&l=MODIS_Aqua_Sea_Ice(hidden),MODIS_Terra_Sea_Ice(hidden),VIIRS_SNPP_Brightness_Temp_BandI5_Night(hidden),VIIRS_SNPP_Brightness_Temp_BandI5_Day(hidden),VIIRS_NOAA20_Brightness_Temp_BandI5_Night(hidden),VIIRS_NOAA20_Brightness_Temp_BandI5_Day(hidden),MODIS_Aqua_Brightness_Temp_Band31_Night(hidden),MODIS_Aqua_Brightness_Temp_Band31_Day(hidden),MODIS_Terra_Brightness_Temp_Band31_Night(hidden),MODIS_Terra_Brightness_Temp_Band31_Day(hidden),IMERG_Precipitation_Rate(hidden),VIIRS_SNPP_DayNightBand_ENCC(hidden),VIIRS_SNPP_DayNightBand_At_Sensor_Radiance(hidden),VIIRS_SNPP_DayNightBand_AtSensor_M15(hidden),Reference_Labels_15m,Reference_Features_15m,Coastlines_15m(hidden),VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2021-01-15-T00%3A00%3A00Z)

Screen Shot 2022-02-18 at 10 34 38 AM

web/js/map/natural-events/cluster.js
```diff
@@ -93,6 +95,8 @@ export const getClusters = ({ geometry, id }, proj, selectedDate, map) => {
   const selectedCoords = geometry.find(({ date }) => date.split('T')[0] === selectedDate).coordinates;
   geometry.forEach((geom) => {
     let { coordinates } = geom;
+    const { magnitudeUnit, magnitudeValue } = geom;
```
jasontk19, line 98: This is a nitpick but, it seems somewhat redundant to destructure two props off an object just to make a new object with those same two props.  I think it's cleaner to either just pass the `geom` object to the `clusterPointToGeoJson` function and destructure the props you need in the param definition of that function signature (which you're already doing).

Or, destructure them here like you are and just pass them as separate parameters.

web/css/events.css
```diff
@@ -91,6 +91,12 @@
   color: #666;
 }
 
+.map-item-list li.item p.magnitude {
```
jasontk19, line 94: Another nitpick but, for styles that share selectors I think it's preferable to nest them to keep them logically grouped in the stylesheet and also to reduce redundant definitions both of which improve readability.  Since this selector and the one below both share `.map-item-list li.item p` you could nest the rest like so:

```
.map-item-list li.item p {
   &.magnitude {
     // these styles
   }
   a {
      // styles from the block below
   }
}
```

A lot of our older CSS does not follow this convention but I have been slowly trying to migrate as much of it as I can when working within a file.

jasontk19 (changes requested):
Looking good just one issue with undefined magnitudes and some code style suggestions!

web/js/map/natural-events/cluster.js
jwhurley1: Ah yeah.  When I was working this out, it made sense to do it this way.  Definitely makes sense to just pass the `geom` object and destructure from that. Thanks!

jasontk19 (approved):
👍🏻  :shipit:

PR #3832, opened 2022-02-23
WV-2178: Eye Icon obscures layer list text in embed mode

jwhurley1:
## Description

Pressing Play for animation in embed mode shows eye icon in layer list and obscures text.  This PR adds an additional class to the `layer-main` div, `layer-visible`, when the `isAnimating` is true.  The CSS in `layer-visible` is only activated when embed mode is active as it is a child of the `.embed-mode` class.

image

## How To Test
1. Open this URL: [localhost](http://localhost:3000/?v=-201.9310639120965,-30.11912669963873,-150.66569698303718,-10.790295040630024&z=4&ics=true&ici=5&icd=10&as=2022-01-15-T04%3A00%3A00Z&ae=2022-01-15-T05%3A20%3A00Z&em=true&l=Coastlines_15m,GOES-West_ABI_GeoColor&lg=true&al=true&ab=on&t=2022-01-15-T04%3A00%3A00Z) and click to interact with the app.
2. Press the play button on the animation menu on the bottom left.
3. The eyeball icon in the layers list should not be overlaying or obscuring the layer information like in the screenshot.

### DIFF
--- a/web/js/containers/sidebar/layer-row.js
+++ b/web/js/containers/sidebar/layer-row.js
@@ -307,6 +307,8 @@ function LayerRow (props) {
     return baseClasses;
   };
 
+  const getMainLayerClasses = () => (isAnimating ? 'layer-main layer-visible' : 'layer-main');
+
   const visibilityTitle = !isVisible && !isDisabled
     ? 'Show layer'
     : isDisabled
@@ -340,7 +342,7 @@ function LayerRow (props) {
 
       <Zot zot={zot} layer={layer.id} isMobile={isMobile} />
 
-      <div className="layer-main">
+      <div className={getMainLayerClasses()}>
         <div className="layer-info" style={{ minHeight: isVectorLayer ? '60px' : '40px' }}>
           <div className="layer-buttons">
             {showButtons && renderControls()}
--- a/web/js/containers/sidebar/layer-row.js
+++ b/web/js/containers/sidebar/layer-row.js
@@ -307,8 +307,6 @@ function LayerRow (props) {
     return baseClasses;
   };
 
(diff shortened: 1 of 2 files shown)


jasontk19:
I am not clear on what the desired behavior is or if this fixes it.  These changes appear to cause the eye icon to show and push the layer description contents over.  However, I wasn't sure that the icons were supposed to show at all in embed mode.  Here is the URL i tested with: http://localhost:3000/?v=-189.98423105176428,-38.55893760847881,-162.6125298433694,-2.3504841317899476&z=4&ics=true&ici=5&icd=10&as=2022-01-15-T04%3A00%3A00Z&ae=2022-01-15-T05%3A20%3A00Z&em=true&l=Coastlines_15m,GOES-West_ABI_GeoColor&lg=true&al=true&ab=on&t=2022-01-15-T04%3A00%3A00Z

jasontk19 (changes requested):
Actually, the ticket indicates that: "the eye icon should not appear in embed mode as you aren't allowed to turn on and off layers in embed mode." so I think this needs to be tweaked

minniewong (changes requested):
Agree with Jason K's comments - I think the eye should not display at all.

web/js/containers/sidebar/layer-row.js
```diff
@@ -307,6 +307,8 @@ function LayerRow (props) {
     return baseClasses;
   };
 
+  const getMainLayerClasses = () => (isAnimating ? 'layer-main layer-visible' : 'layer-main');
```
jasontk19, line 310: This could just be a static var instead of a function:  
```const getMainLayerClasses = isAnimating ? 'layer-main layer-visible' : 'layer-main';```

jasontk19 (approved):
Looks good!

PR #3836, opened 2022-03-02
WV-2194: Add long press functionality to context-menu.js

jwhurley1:
## Description

Provides the ability to do long press interactions on iOS to bring up the context menu.

## How To Test

After pulling down the branch, run `npm ci && npm run build && npm start`.

### With BrowserStack

1. Open BrowserStack and select an iOS device, open Safari or Chrome, and navigate to your computer's local IP address with `:3000` after the IP address.  Example: `192.168.0.123:3000`.
1. Click and hold with your mouse on the map for 1 second.
1. Context menu should appear.
1. Test each of the menu items to see if they work properly.
1. Press and hold again.
1. Click/tap off of the context menu to make it disappear.

### Without BrowserStack
1. Opening this link [localhost](http://localhost:3000) 
2. Open Chrome DevTools (F12 or CTRL/CMD+Shift+i). 
3. Click on mobile device icon on the top-left of the Devtool tools to toggle on the device toolbar.
4. Select any device that presents the WV app in portrait orientation.
5. Repeat steps 3 - 7 from the "With BrowserStack" section.

### DIFF
--- a/web/js/components/context-menu/context-menu.js
+++ b/web/js/components/context-menu/context-menu.js
@@ -18,17 +18,23 @@ function RightClickMenu(props) {
   const [pixelCoords, setPixelCoords] = useState({ pixel: [0, 0] });
   const [toolTipToggleTime, setToolTipToggleTime] = useState(0);
   const [formattedCoordinates, setFormattedCoordinates] = useState();
+  const [isHoldPress, setIsHoldPress] = useState(false);
   const {
     map, crs, unitOfMeasure, onToggleUnits, isCoordinateSearchActive, allMeasurements, measurementIsActive,
   } = props;
   const [getMap, setMap] = useState(map);
   const measurementsInProj = !!Object.keys(allMeasurements[crs]).length;
 
-  const handleClick = () => (show ? setShow(false) : null);
+  const handleClick = (event) => {
+    if (isHoldPress) return;
+    return show ? setShow(false) : null;
+  };
 
   function handleContextEvent(event, olMap) {
     if (measurementIsActive) return;
-    event.originalEvent.preventDefault();
+    const incomingEvent = event.type === 'contextmenu' ? event.originalEvent : event;
+    incomingEvent.preventDefault();
+
     const coord = olMap.getCoordinateFromPixel(event.pixel);
     const [lon, lat] = transform(coord, crs, 'EPSG:4326');
     setPixelCoords({ pixel: event.pixel });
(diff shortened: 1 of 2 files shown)


web/js/components/context-menu/context-menu.js
```diff
@@ -68,13 +74,38 @@ function RightClickMenu(props) {
     };
   };
 
+  let longPressTimer = null;
+
+  function setTimer(incomingEvent, incomingOlMap) {
+    longPressTimer = setTimeout(() => {
+      setIsHoldPress(true);
```
jasontk19, line 81: It seems odd that this gets set to true and never will be reset to false at any point.  There is an issue I am seeing that might be related.  Sometimes a click and hold mouse interaction to drag the map causes the context menu to appear.  And also, sometimes right-clicking after left click causes the context menu to appear and then immediately disappear.

jasontk19 (changes requested):

web/js/components/context-menu/context-menu.js
jasontk19: Here is a demonstration: https://drive.google.com/file/d/11mbMyl61-Ek3wj1Ug05n97IcQ0l-TVH8/view?usp=sharing

minniewong (approved):
Works well when I test in the browser using dev tools 👍

web/js/components/context-menu/context-menu.js
```diff
@@ -68,13 +74,43 @@ function RightClickMenu(props) {
...
+  }
+
   useEffect(() => {
     if (isCoordinateSearchActive) return;
     events.on('map:singleclick', handleClick);
     events.on('map:contextmenu', handleContextEvent);
+
+    events.on('map:touchstart', handleTouchStart);
+    events.on('map:touchend', clearTimer);
+    events.on('map:drag', handleDrag);
```
jasontk19, line 107: Maybe not a huge deal but, I noticed that this considers any mouse movement, with either left or right click down, to be a drag event.  That means that, if you move the mouse even slightly while right-clicking to show the context menu that it immediately gets hidden again.  On desktop this probably isn't likely but I'm wondering if this is a concern on mobile where it may be more likely that a user slightly moves their finger while doing a long press.  I wonder if we can listen for a different event that would only fire when a left-click + drag event occurs.

jasontk19 (changes requested):
Just a couple more things that need to be tweaked:

* This seems to break double-tap to zoom when I try it on an actual mobile device
* As mentioned in another comment, moving the map even slightly while long-pressing on mobile will hide the menu
* The menu items should be larger on mobile (refer to toolbar menus sizing for guidance)

PR #3839, opened 2022-03-09
WV-2214: Fix compare date labels

jasontk19:
## Description

- Make sure compare labels update even when no temporal layers
- Fix compare label styling when A & B dates are the same

## How To Test

1. Load app with same date on both sides of compare
2. Confirm date does not show in A/B label
3. Change date on either side
4. Confirm dates now show in A/B label


minniewong (approved):
🚀  looks good!

PR #3841, opened 2022-03-09
WV-1931: Provide ability to add multiple markers

jwhurley1:
## Description

Provide ability to add multiple markers with labels (using existing ESRI gazetteer labels and coordinates) on the map via right click context menu and are accessible via permalink.

## How To Test
1. Right click anywhere on the map.
2. Select "Add a marker"
3. Add two or three more markers
4. Minimize each marker.
5. Copy coordinates of each marker.
6. Close each marker.

## Acceptance Criteria
1. [ ] Via the right click context menu, a user will be able to add more than one place marker on the map. 
1. [ ] The marker locations are saved in the URL and are accessible via permalink.


jasontk19:
Placing a marker in a polar projection seems to cause an error: 
Screen Shot 2022-03-09 at 10 04 45 AM

Perhaps related to this issue, markers don't remain showing when switching between projections.  Basically, if I place something in geographic at a latitude where it would be visible in a polar projection, it should still show when I switch to that projection.

jasontk19:
Removing a marker from the map _does_ cause it to no longer be visible but, looking in the Redux store it looks like it's just _adding_ a `null` marker rather than removing one from the list.
Screen Shot 2022-03-09 at 10 11 31 AM

jwhurley1:
> There is a `isCoordinatesDialogOpen` property on `locationSearch` state in the Redux store. Previously this represented whether or not the single location marker had it's tooltip showing. Now that there are multiple, this is no longer valid. I don't see this being used outside of the actions/reducers so we might be able to do away with it, otherwise we may need to store this information per-marker. More investigation is needed.

Still working out the above.

jasontk19 (changes requested):
I can't seem to place markers any more on this branch.  I get this error:
Screen Shot 2022-03-15 at 9 45 49 AM

jwhurley1:
> I can't seem to place markers any more on this branch. I get this error: Screen Shot 2022-03-15 at 9 45 49 AM

Yeah I went to check again after our meeting this morning.  Thought it was working when I pushed up yesterday.  I have some changes incoming shortly...

jasontk19:
There is an issue with removing markers. To reproduce:

1. Place five location markers
2. Refresh the page
3. Start removing markers
4. At some point, clicking one may remove another or one marker cannot be removed

web/js/map/ui.js
```diff
@@ -282,25 +305,28 @@ export default function mapui(models, config, store, ui) {
...
     const state = store.getState();
-    const { locationSearch, proj, browser } = state;
-    const {
-      coordinates, reverseGeocodeResults,
-    } = locationSearch;
-    const results = geocodeResults || reverseGeocodeResults;
+    const { proj, browser } = state;
+    const results = geocodeResults;
+    if (!results) return;
+    // const coordinatesObject = coordinates.find((element) => element.latitude === results.location.x && element.longitude === results.location.y);
```
jasontk19, line 313: Can remove this commented out code

jasontk19 (changes requested):
The issue with not being able to remove a marker after page reload seems inconsistent, I will keep trying to see if I can track down why.

jasontk19:
There is still an issue removing pins after loading some from a URL and adding more.  See here how only three locations show in the store but there are more on the map.  The ones on the map that aren't showing in the store cannot be removed.
Screen Shot 2022-03-18 at 8 56 40 AM

web/js/components/location-search/coordinates-dialog.js
```diff
@@ -83,7 +85,7 @@ class CoordinatesDialog extends Component {
       <>
         <span
           id={closeButtonId}
-          className={`close-tooltip ${closeButtonId}`}
+          className={`close-tooltip ${closeButtonClassName}`}
           onTouchEnd={this.clearCoordinates}
```
jasontk19, line 89: Place markers can't be removed on mobile and I believe this is why:

It looks like a different function is called `onTouchEnd` for this `span` as compared to the `onClick` for the `<FontAwesomeIcon>` element contained within.  `this.clearCoordinates` is not even defined in this file anymore.  And I think we want to remove the `CLEAR_MARKERS` action/constant/reducer as well as where it's being picked up in `map/ui.js` and causes `removeAllCoordinatesMarkers()` to be called.

web/js/modules/location-search/actions.js
```diff
@@ -98,7 +102,7 @@ export function setPlaceMarker(coordinates, reverseGeocodeResults, isInputSearch
 export function clearCoordinates() {
```
jasontk19, line 102: This should no longer be needed unless we expect to add functionality to remove all markers at some point.

jasontk19 (approved):
Looks good, probably just want to rebase with the release branch to get the latest Chromedriver version so the tests can run properly so we can verify they pass before merging.

PR #3843, opened 2022-03-09
WV-2218: Make sure compare dates update properly

jasontk19:
## Description

* We were using stale references to `state` in several compare `update` functions which could cause the wrong dates to show sometimes.

## How To Test

* Try all three compare modes
* Try changing dates, changing to the same date, resizing, etc.


minniewong (approved):
Working now! 👍

PR #3844, opened 2022-03-10
WV-2219: Fix swipe line position on resize

jasontk19:
## Description

- destroy/recreate swipe line on resize


minniewong (approved):
great! working well

PR #3849, opened 2022-03-23
WV-2198: Implement updated design elements to PWA

jwhurley1:
## Description

This PR adds the new WorldView icons for the PWA and sets the app's theme color to white.

## How To Test

1. After pulling down the changes, run `npm run build && npm start`. 
2. Open [localhost](http://localhost:3000) in chrome and click the "install" icon in the URL.  Looks like an arrow pointing down into a box.
3. You should be prompted to install the app.  It should show the new Worldview globe icon with the white lines running through it.
4. The app should open with a white border around it.

To test on mobile:
1. After running `npm start`, go to your computer's IP address at `:3000` in your phone's browser.  Example:  `192.168.0.123:3000`.
2. Depending on your browser, you may eventually be prompted to download as a PWA.  If not, you can install by going to your browser's "More Options" menu (usually ellipses) and "Add to Home Screen" or similar.
3. Open the app after it has been added to your home screen and you should see a white splash screen with the white Worldview logo in the center.

PR #3850, opened 2022-03-24
Merge `feature-pwa` into `release-3.21.0`

jwhurley1:
Merge `feature-pwa` into `release-3.21.0`.


jasontk19:
Can we rebase this with the release branch?

jwhurley1:
> Can we rebase this with the release branch?

Done!

jasontk19:
So I opened this in Safari on my iPhone 12 and chose "Add to Home Screen" which, it did.  But, the new icon was not used and instead the icon is just a screenshot of the app (like it is by default)

jasontk19:
This PR doesn't appear to include the changes from https://github.com/nasa-gibs/worldview/pull/3849

jwhurley1:
Hmm ok.  Lemme try again

jasontk19:
Some things we should probably address before this goes into a release:

- No loading screen showing on iPhones
- Bottom buttons can be too close to the edges on some phones (particularly newer iPhones) and the swipe bar to switch apps can get in the way
- In landscape mode, if you open the product picker, the Layers/Events tabs and the chevrons to collapse this menu are all difficult to press since they are so small and right against the edge of the screen.  I think touch events that happen right at the edge of the screen are often interpreted as the beginning of a swipe down motion and so aren't passed through to the browser

jasontk19:
I just pushed the most recent updates to the GIBS UAT instances for `feature-pwa`.  Things that should be tested in both landscape and portrait orientations:

* Sidebar open/close
* Sidebar sizing with comparison mode on/off 
* Compare AB button placement
* Animation button placement
* Timeline date changes

Also it's likely that embed mode was affected so we should check that as well for any issues.

minniewong (commented):
Polar Rotate buttons are missing

jasontk19 (approved):

PR #3853, opened 2022-03-29
fix broken e2e tests

jasontk19:


Benjaki2 (approved):

galitniztad:
> You have a bug in the map view, see attachment, it does not show the surface temperature correctly, also stopped giving measurement comparisons from 2020 backwards.
nasa

PR #3854, opened 2022-04-01
WV-2199: Create e2e testing for right-click menu

jwhurley1:
## Description

This PR adds e2e testing for the right click menu. Every menu item has a test written for it except for the following:
- Copy to clipboard: Wasn't sure how to check the clipboard.  In retrospect, maybe we can add a npm package just for this?
- Testing for Firefox.  Was not able to get Nightwatch to execute clicks in Firefox.

## How To Test
- Run
`npx nightwatch -e chromeHeadless e2e/features/ui/context-menu-test.js` to run only in the CLI or 
`npx nightwatch -e chrome e2e/features/ui/context-menu-test.js` if you want to watch the browser automate live.


e2e/features/ui/context-menu-test.js
```diff
@@ -0,0 +1,106 @@
...
+    c.expect.element(contextMenu).to.be.present;
+  },
+
+  'Adding a marker after clicking "Add Place Marker" in context menu': (c) => {
+    if (c.options.desiredCapabilities.browserName === 'firefox') {
+      return;
+    }
+    c.moveToElement(contextMenuAddMarker, 10, 10);
+    c.mouseButtonClick('left');
+    c.waitForElementVisible('#marker-pin');
```
jasontk19, line 53: I think this _should_ fail if the marker doesn't show but, wouldn't be a bad idea to add some `assert` or `expect` statement here.

jasontk19:
For testing copy to clipboard, could you just verify that the "Copied to clipboard!" tooltip appears?  That at least confirms that some action was taken when clicking that item in the menu.

minniewong:
@jwhurley1 Looks like one of the checks is still not working. Could you look into it?

jwhurley1:
> @jwhurley1 Looks like one of the checks is still not working. Could you look into it? 

Yep I'm aware. I'll pick it up tomorrow after the my training course. Thanks!

jwhurley1:
So the checks for other tests seem to be inconsistent. They all pass on my local machine. Sometimes they pass, sometimes they fail in the CI/CD pipeline. This Firefox-Windows failure had never happened before, for example. Nothing I'm working on is related to this test that failed so it's like hit or miss. Not sure what's going on.

I'll keep checking it out but it's like playing whack-a mole....

jasontk19:
Looks like the last run of tests pass but now the branch is out of date with the release branch 😬

jasontk19 (approved):
:shipit:

PR #3858, opened 2022-04-06
WV-2013: Use UMM-T record for EDSC smart handoffs

jasontk19:
## Description

* Conditionally render smart handoff component only if data tab is active
   * Prevents having to make CMR API calls to validate layer data collections on every page load
   * Requires a brief loading period when first switching to data tab to make CMR API calls
     * Show loading spinner in tab while waiting for calls to return
     * Show informative error message if API calls fail
     * Metadata pulled from CMR now stored in Redux in order to support conditional rendering of SH component
* Restructure `features.json` so that the `smartHandoffs` property is an array of services that include a name and UMM-T record identifier
  * Absence of this property still means that no data tab will show when app is loaded
* Use selectors to build CMR request URLs and pull CMR base URL from `features.json` instead of hardcoding CMR URL in many places throughout the app
* Use constants for loading keys instead of duplicating strings
* Bring in url-template package to parse the URL template from UMM-T record for EDSC

### To test loading spinner

1. Load the app
2. Throttle connection to "Slow 3G" in devtools
3. Switch to Data tab
4. Confirm loading spinner shows

### To test failed API call message

1. Load app
2. Throttle connection to "Offline" in devtools
3. Switch to Data tab
4. Confirm error message shows after network timeout

### Test smart handoffs

* Smart handoffs behavior should otherwise be unchanged.  Try handing off to EDSC with various settings, dates, bounding boxes, and collections.


minniewong (commented):
If you go to the Data tab while you're online, then subsequently go offline, and select a radio button, it doesn't return the message that CMR cannot be reached, and the three dots under Available granules animate non-stop. Perhaps it should timeout after a certain amount of time?
I also suggest changing the error text slightly: "Data records from the Common Metadata Repository (CMR) could not be reached. Data downloads are not possible at this time."

Benjaki2 (approved):
Looks good!

PR #3861, opened 2022-04-14
Fix previewSnapshots config option

jasontk19:
## Description

Fixes an issue where this property MUST be present and defined as `false` to disable this feature:
 (e.g. `"previewSnapshots": false`)  

The absence of the property should be considered a `false` value rather than breaking the script.

## How to test

1. Open features.json
2. Remove "previewSnapshots" property entirely
3. `npm run build`
4. Confirm no errors


minniewong (approved):
Works as described!

PR #3865, opened 2022-04-20
WV-2284: Fix geostationary gifs in FireFox

jasontk19:
## Description

* Firefox can't translate the same datetime strings into date objects that Chrome can so, pass down a date object where needed and string elsewhere for display.

## How To Test

Create a GIF in FireFox with geostationary layers loaded


minniewong (approved):
Working well in Firefox! Also checked in Chrome and Safari and looks good there too

PR #3868, opened 2022-04-22
Open privacy policy link on new page

minniewong:
## Description

Update to open Web Privacy Policy link on new page. 

@nasa-gibs/worldview


jasontk19 (approved):
The `rel="noopener noreferrer"` is only needed if you want to block the page you're linking to from being able to see where incoming traffic is coming from.  Not sure if that was intentional or not here.

minniewong:
> The `rel="noopener noreferrer"` is only needed if you want to block the page you're linking to from being able to see where incoming traffic is coming from. Not sure if that was intentional or not here.

Not intentional - I just copied it from the other links :)

PR #3869, opened 2022-04-26
WV-2232: Snapshots for multiple markers

jasontk19:
## Description

- Handle multiple markers snapshots
- Fix issue with polar proj snapshots
- Swap misnamed lat/lon vars throughout


minniewong (approved):
awesome! works in all projections!

PR #3876, opened 2022-05-11
WV-2154: More accurate animation frame timing

jasontk19:
## Description

* Use a more accurate animation timing method

## How To Test

* Try animations at various lengths and speeds

### DIFF
--- a/web/js/components/animation-widget/play-queue.js
+++ b/web/js/components/animation-widget/play-queue.js
@@ -12,17 +12,23 @@ const CONCURRENT_REQUESTS = 3;
 const toString = (date) => util.toISOStringSeconds(date);
 const toDate = (dateString) => util.parseDateUTC(dateString);
 
-
+// Get the initiall buffer size, using a larger buffer for higher speeds
+const getInitialBufferSize = (numberOfFrames, speed) => {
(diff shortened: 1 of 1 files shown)


minniewong (approved):
Looks really good for most cases! Only issue I had was the first load Step 7 of the Geostationary tour story (https://go.nasa.gov/3FBYIok), it preloaded the buffer and played the first half pretty well, then started stuttering and you could occasionally see flashes of the preloading buffer bar. If you run through it again, it all loads well. This animation is quite long, so this issue is not a show stopper.

Benjaki2 (approved):

jasontk19:
Updated to do two things:

* Base the initial buffer size on the speed (larger initial buffer for higher speeds for more accurate loading time estimations)
* Don't divide the calculated total load time by the number of concurrent requests since, even though there can be three concurrently pending promises, they aren't actually queued up in batches of three at a time (meaning they don't exactly complete in batches of three at a time).  This often makes for slightly longer than necessary pre-loading times but minimizes the likelihood of the playback catching up to the buffer.

minniewong (approved):
🚀  really smooth now!

PR #3880, opened 2022-05-17
WV-2209: Faster concurrent requests during build

jasontk19:
## Description

Change the way Python makes http requests concurrently during build time in order to speed up the build process significantly.

* Bring in `httpx` lib
* Use Python 3 `async/await`, `asyncio` to run requests concurrently


minniewong (approved):
Worked for me after doing an `npm ci`

jasontk19:
Need to take a closer look at this when building on VPN, there appears to be an issue when pulling a bunch of things at once.

jasontk19:
Applied connection limits and appears to work well on VPN now

PR #3881, opened 2022-05-17
WV-2295: Latest overlapping granule on top

jasontk19:
## Description

- Order granule footprints on hover so that latest shows

## How To Test

* Checkout `granule-configs` branch
* Run `npm run build`
* Checkout this branch
* Run `npm run watch`

* Check overlapping granules and confirm that latest datetime appears on top.


minniewong (approved):
Appears the newest granules are on top!

PR #3883, opened 2022-05-18
Made Chlorophyll a layers inactive

minniewong:
## Description

Fixes WV-2308

Made MODIS_Aqua_Chlorophyll_A and MODIS_Terra_Chlorophyll_A inactive as they are no longer being processed. Added a note to the layer description to let users know to use the new Chlorohphyll a (L2) layers.

@nasa-gibs/worldview


jasontk19:
Should the corresponding layer-metadata configs have the `ongoing` value changes to `false` as well?  That would be the property that would come from the GIBS vis metadata to replace this, if we were including it.

jasontk19 (approved):

minniewong:
> Should the corresponding layer-metadata configs have the `ongoing` value changes to `false` as well? That would be the property that would come from the GIBS vis metadata to replace this, if we were including it.

I didn't think about that, so let me go change that too in the layer-metadata

minniewong:
> > Should the corresponding layer-metadata configs have the `ongoing` value changes to `false` as well? That would be the property that would come from the GIBS vis metadata to replace this, if we were including it.
> 
> I didn't think about that, so let me go change that too in the layer-metadata

I have opened a PR for that to be updated.

PR #3894, opened 2022-05-26
WV-2313: Fix issue with layers added on event selection

jasontk19:
## Description

There were a few problems related to this:

* The map `reloadLayers` function wasn't being called when event category layers were added so, even though they showed in the sidebar they weren't added to the map
* Since a lot of the `map/ui.js` functions were made async, there was a race condition between `updateDate` and `reloadLayers` that happened when selecting an event since we were trying to respond to two separate synchronous Redux actions by calling two async functions potentially in parallel, causing them to modify the same map object at the same time which created issues.
* Removed a `CHANGE_TAB` that appears to be leftover from the old data download approach (can't think of a reason why we would need to reload layers when switching to or from that tab anymore)
* Cleaned up a little bit of code around setting the date for an event

## How To Test

* Retry steps in ticket
* Try various event related behaviors to ensure nothing was inadvertently affected

### DIFF
--- a/web/js/map/natural-events/natural-events.js
+++ b/web/js/map/natural-events/natural-events.js
@@ -6,6 +6,7 @@ import * as olProj from 'ol/proj';
 import {
   getDefaultEventDate,
   validateGeometryCoords,
+  toEventDateString,
 } from '../../modules/natural-events/util';
 import util from '../../util/util';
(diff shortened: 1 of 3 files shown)


minniewong (approved):
awesome - I don't see any issues!

Benjaki2 (approved):

PR #3897, opened 2022-06-03
WV-2313: Fix event dates

jasontk19:
## Description

Make sure date formatting for events with multiple dates works right.


minniewong (approved):
All working well!

PR #3898, opened 2022-06-07
WV-2317: Add Ability To manually input image download extents

Benjaki2:
## Description
Add Ability To manually input image download extents

Fixes #wv-2317
Fixes #wv-2316

- [x] update react-image-crop version
- [x] make image download container state based on GEO lat-long values instead of pixel boundaries
- [x] Add input component
- [x] Hide input selection by default

[If this is a relatively large or complex change, start a discussion by explaining why you chose the solution you did and what alternatives you considered, etc...]

## How To Test

1. Open Image download dialog
2. expand the coordinates selection
3. Verify expected result
4. Verify works in polar projections

## PR Submission Checklist

@nasa-gibs/worldview

### DIFF
--- a/e2e/features/image-download/global-select.js
+++ b/e2e/features/image-download/global-select.js
@@ -1,9 +1,7 @@
-const { zoomIn, zoomOut } = require('../../reuseables/zoom');
 const { bookmark } = require('../../reuseables/bookmark');
 const { normalizeViewport } = require('../../reuseables/normalize-viewport');
 const {
   openImageDownloadPanel,
-  closeImageDownloadPanel,
 } = require('../../reuseables/image-download');
 
 const startParams = [
@@ -26,23 +24,18 @@ module.exports = {
     c.expect.element(globalSelectInput).to.be.present;
     c.expect.element(globalSelectInput).to.not.be.selected;
   },
-
   'Verify that checking checkbox updates bounding-box labels': function(c) {
     c.expect.element('#wv-image-top').text.to.not.contain('180.0000');
     c.click(globalSelectInput);
     c.expect.element('#wv-image-top').text.to.contain('180.0000');
+    c.expect.element('#wv-image-bottom').text.to.contain('180.0000');
+    c.pause(500);
   },
-  'Verify that checkbox is gone after zoom and bounding box is no longer around globe': function(c) {
-    closeImageDownloadPanel(c);
-    zoomIn(c);
-    openImageDownloadPanel(c);
-    c.expect.element(globalSelectInput).to.not.be.present;
+  'Verify that unchecking checkbox updates bounding-box to previous': function(c) {
+    c.click(globalSelectInput);
     c.expect.element('#wv-image-top').text.to.not.contain('180.0000');
-  },
-  'Verify that checkbox is back after zooming back out': function(c) {
-    closeImageDownloadPanel(c);
-    zoomOut(c);
-    openImageDownloadPanel(c);
-    c.expect.element(globalSelectInput).to.be.present;
+    c.expect.element('#wv-image-top').text.to.contain('35.1563');
+    c.expect.element('#wv-image-bottom').text.to.contain('35.1563');
+    c.pause(2000);
   },
 };
--- a/e2e/features/image-download/input-test.js
+++ b/e2e/features/image-download/input-test.js
@@ -1,0 +1,41 @@
+const { loadAndSkipTour } = require('../../reuseables/skip-tour');
+const {
+  openImageDownloadPanel,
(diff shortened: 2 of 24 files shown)


minniewong (changes requested):
I suggest swapping "Max Longitude" with "Max Latitude" as well as "Min Longitude" and "Min  Latitude" so that it matches the pattern of the existing lat-long labels at the corners of the dashed selection box. (as well as matching WVS)

jasontk19:
1. Open snapshot tool
2. Change any lat/lon input
3. Close snapshot tool
4. Open snapshot tool again
Result:  Crop box doesn't show, coordinate inputs are blanked out.  NaN values show for coordinates at edge of screen.

Screen Shot 2022-06-08 at 2 23 28 PM

jasontk19:
I think this gets the job done but could possibly be more user friendly.

* The "invalid input" messages are easy to miss and should probably be colored red
* Having a "snap to globe" type of button or something similar to quickly set max extents would be nice, not sure if we decided against that or not
* It looks a little odd with a grey background since it matches the "download" button coloring underneath. Perhaps just the background for the click toggle area and only an outline for the fields underneath?

A new bug:

1. Drag one of the crop box handles across the entire box area, the snapshot tool closes and there is a console error.

Benjaki2:
@jasontk19 I think that I have addressed your issues. The "snap to globe" implementation has its own ticket that I plan to tackle next.

minniewong (changes requested):
I tried to manually edit the coordinates, and expected the bounding box to update with it, but only the coordinate labels moved. Should the actual bounding box also update? in the screenshot, you can see that the coordinate labels moved, but the bounding box did not expand along with the labels. 

Screen Shot 2022-06-13 at 3 46 36 PM

minniewong (changes requested):
When you use the up and down arrows to change the coordinates or type in coordinates and use your mouse to go to each box, it does not update the bounding box on the map. Updates to the coordinates only occur if you use tab or enter. Should it update when you click between the boxes? Also, i wonder if we need to add an "i" note to let users know that the bounding box coordinates are limited to what's visible in the viewport i.e. if you are zoomed in a bit, you can't request the whole globe.

Benjaki2:
@minniewong  I agree that it should "try" to update when clicking from box to box and the "i" button sounds like a good idea. 👍  Let's talk a little bit more about where the "i" button should be and what exactly it should say.

I don't really think we should update the input while it's being edited. In my mind these arrows are different from the date arrows since they are actually part of the input. There is no distinction from typing and pressing this button event-wise in the DOM.  (We can come up with something to distinguish the arrowClick. Let me know if you feel strongly about implementing the click)

Right now the coordinate inputs don't really work over the dateline and I'm a little uncertain about how to handle this. Any ideas?

minniewong:
> 

Perhaps we just need a "submit" button instead of updating the box on the fly.

jasontk19:
Looks like some unit tests are failing

jasontk19:
Overall it's working well, just a few minor issues and UX observations to consider.

---

The dateline crossing warning shows when you snap to globe which doesn't seem right.  We probably just need to tweak the coordinates or conditional check used to determine if the dateline was crossed or not.

Screen Shot 2022-06-23 at 10 52 36 AM

---

I can enter invalid coordinates into the inputs resulting in a box that covers no area:

Screen Shot 2022-06-23 at 10 55 32 AM

---

There is something weird that happens when I click + drag to highlight the numbers in an input.  If I click + drag inside the input but, mouseUp with the pointer outside of the input, the selection gets cleared.  I frustratingly tried that a few times before I realized I would have to double-click or be more precise on where I release the click.  Not sure what would cause that or if it's an easy fix.

---

If you aren't zoomed out far enough to be allowed to check the "Select Entire Globe" checkbox, you have to close the snapshots tool to zoom out and then open it again, which can be annoying if you have to do it multiple times because you didn't zoom out far enough the first time (I didn't 😅 ).  I still think it would be nice if it just zoomed out for you when you made the choice to select the whole globe.

---

A checkbox input for selecting the whole globe feels weird since you can't un-check it once the selection is made.  Maybe a button that disables if the coordinates are exactly the full globe would feel less out of place?

minniewong (approved):
I think it looks good and it all seems to work well! 🚀

jasontk19:
Updated this to fix the tests and now, clicking the checkbox will zoom the map out to fit the globe instead of just telling you it can't be done.  Unchecking the checkbox will revert to previously selected coordinates.

minniewong (approved):
Snazzy! I like how checking the box goes to global view 🌎

PR #3900, opened 2022-06-13
WV-2331: Fix incorrect date for imagery when toggling visibility

jasontk19:
## Description

Make sure we call updateDate when toggling visibility


minniewong (approved):
thanks for the quick fix!

PR #3913, opened 2022-06-27
Removed orbit tracks from multi day MODIS layers

minniewong:
## Description

Fixes WV-2344.

Removed orbit tracks from some 8-day and Monthly MODIS layers that were previously missed.

@nasa-gibs/worldview


jasontk19 (commented):
This shouldn't be necessary as long as the GIBS metadata configs are correct.  These properties are going to be removed entirely from these configs.

PR #3920, opened 2022-07-05
WV-2289: Make layer retract button more prominent on mobile

jasontk19:
## Description

* Fix CSS selectors
* Increase width/height


minniewong (approved):
Looks good!

PR #3924, opened 2022-07-06
WV-2318: Smart handoffs for granule layers

jasontk19:
## Description

- Smart handoffs handling for granule layers

## How To Test

1. Checkout `granule-configs` branch, pull latest
2. `npm run build`
3. Checkout this branch
4. `npm run watch`

Note:  The granule test layers only have NRT collections associated with them but, the imagery from GIBS is from 2019.  Therefore, when handing off to Earthdata Search, no results will show.  Similarly, the CMR request for a granule count will always show none.


minniewong (approved):
👍

PR #3930, opened 2022-07-14
WV-2345, WV-2364

jasontk19:
## Description

- fix thresholding issue on stepped pallettes
- add padding to bottom of double palette container


minniewong (approved):
Looks great!

PR #3936, opened 2022-07-19
WV-2333: Upgrade ol and ol-mapbox-style

jasontk19:
## Description

* Upgrade ol and ol-mapbox-style to latest versions

## How to test

The map should be thoroughly tested.  Specific things to check would be:

* WMS and WMTS layers
* Vector layers and interactions
* Layers with palettes


minniewong (approved):
Looks good!

PR #3939, opened 2022-07-25
WV-2202: Disallow context menu outside map extent

jasontk19:
## Description
- dont show context menu outside visible extent
- allow location marker in map wings

## How To Test

* Attempt to open context menu outside extent in all projections
* Place location marker in wings, confirm coordinates show correctly (e.g. longitude not greater than 180 or less than -180)
* Place location markers in polar projections
* Confirm coordinates in context menu match coordinates in location marker tooltip
* Confirm location markers cannot be placed outside visible extent
* Confirm location markers are placed properly on page refresh (url params)


minniewong (changes requested):
I noticed two issues with the markers in the wings: When you put a point in the wings, change projections, and change back again, the point disappears. Also, if you put a point in the wings, save the URL then open the URL in a new tab, the marker does not appear. I am currently waffling between whether this behavior should be allowed or not. We may want to ping @rboller on his thoughts.

minniewong:
Let's allow the markers in the map wings!

minniewong (approved):
🎸

PR #3940, opened 2022-07-27
WV-2359 Disable Set Area of Interest Without Selection

ryanweiler92:
## Description

Fixes # .

Edited containers/sidebar/smart-handoff.js to check for a selected collection before enabling the earthdata button and rendering the set area of interest button.

## How To Test

When visiting this URL 

http://localhost:3000/?v=-157.50605321419718,-124.98135679600885,138.05460701253722,86.99656135638496&sh=VIIRS_SNPP_CorrectedReflectance_TrueColor,C1344298008-LANCEMODIS&t=2021-10-18-T12%3A27%3A23Z

The download earthdata button should be disabled and the set area of interest button should be hidden until you select a collection.


web/js/containers/sidebar/smart-handoff.js
```diff
@@ -580,7 +580,7 @@ class SmartHandoff extends Component {
           <hr />
           {this.renderLayerChoices()}
           <hr />
-          {this.renderCropBox()}
+          {!selectedCollection ? undefined : this.renderCropBox()}
```
jasontk19, line 583: This works great! Can we move the conditional check to the return statement for the `renderCropBox()` function on line 437?

For example:
`return selectedLayer && selectedCollection && proj.id === 'geographic' && ( ... )`

That way all the logic for determining if this shows or not lives in the same place.

jasontk19 (changes requested):

web/js/containers/sidebar/smart-handoff.js
ryanweiler92: Thanks, I've consolidated the conditional check for the cropbox. It still works the same.

jasontk19 (approved):

PR #3943, opened 2022-08-03
WV-2290: Changing Location of Loading Spinner on Mobile

ryanweiler92:
## Description

Fixes # .

The loading spinner was covering the toolbar on mobile views. I moved the loading spinner on mobile devices so that it no longer covers the toolbar on devices 530px & smaller. The styling is created in a JS object so I had to create a mediaQuery variable that uses window.matchMedia to create a conditional statement for mobile devices.

## How To Test

Load a layer on a mobile view that is 530px or smaller and the loading spinner will render in an empty space next to the toolbar.

### DIFF
--- a/web/js/components/map/loading-spinner.js
+++ b/web/js/components/map/loading-spinner.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import { Spinner } from 'reactstrap';
 
 function LoadingIndicator({ msg, isLoading }) {
-  const mediaQuery = window.matchMedia("(max-width: 530px)");
+  const mediaQuery = window.matchMedia('(max-width: 530px)');
 
   const spinnerStyle = {
     position: 'absolute',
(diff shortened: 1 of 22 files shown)


web/js/components/map/loading-spinner.js
```diff
@@ -4,21 +4,38 @@ import PropTypes from 'prop-types';
...
   };
 
+  const mobileSpinnerStyle = {
+    position: 'absolute',
+    top: 10,
+    left: 130,
+    zIndex: 999,
+  };
+
+  if (mediaQuery.matches) {
```
jasontk19, line 23: Instead of duplicating the return statements, can you do the conditional check where the style is declared? For example:

```const spinnerStyle = mediaQuery.matches ? {  mobile styles } : { default styles };```

or

```<div styles={mediaQuery.matches ? mobileSpinnerStyle : spinnerStyle}>```

jasontk19 (changes requested):
Would you mind updating the PR title to match the format used in the release notes (and include a descriptive title)?

minniewong (changes requested):
Is it possibly to move the loading spinner directly to the right of the logo? It might be easier to spot if it's anchored to something.
Screen Shot 2022-08-08 at 1 52 28 PM (2)

minniewong (commented):
Location looks good in both Chrome and Firefox, but on Safari it's over to the right. (though not a terrible location, so if it's difficult to move it in safari, we can leave it)
Screen Shot 2022-08-09 at 1 38 56 PM (2)

minniewong (approved):
Looks good in Chrome, Firefox and Safari!

PR #3944, opened 2022-08-03
WV-2217: Updating Coordinate Format on Location Markers

ryanweiler92:
## Description
When a user placed a location marker, and then pressed the change coordinate format button, the coordinates in the location marker were not updating.

## Fixes 

This PR updates the coordinates format in the placeable location marker when the user presses the change coordinate format button. I have hooked up the change coordinate format button to Redux under settings. Unfortunately, after a lot of research and discussion, I don't think it's possible, with the way we are currently rendering location markers, to hook up the coordinates-dialog component to Redux directly. I have created a simple solution to update the coordinates format with the current utility functions we have in place. I am going to leave the work I have completed for hooking up the coordinate format change button to Redux as it maybe helpful in the future to have this setting already hooked up to Redux. (I have another ticket, wv-2373, to create a selection in the settings tab to change the format where I will use the Redux work I have set up here.)

## How To Test

Add a place marker and then press the Change Coordinate Format button and the coordinates in the location markers will update in real time.


web/js/components/location-search/coordinates-dialog.js
```diff
@@ -22,8 +26,13 @@ class CoordinatesDialog extends Component {
     setTimeout(() => {
       this.setState({ showTooltips: true });
     }, 200);
+    events.on('location-search:coordinate-format', this.updateCoordinateFormat);
```
jasontk19, line 29: We should also deregister this event when the component unmounts.  The call to deregister is basically the same as this just with `event.off()` instead of `event.on()`.

web/js/modules/settings/actions.js
```diff
@@ -29,3 +30,15 @@ export function changeDatelineVisibility(value) {
     value,
   };
 }
+
+export function changeCoordinateFormat(value) {
+  if (!value) {
```
jasontk19, line 35: Is there a case where we don't pass a `value` to this action creator?  I think there always has to be _some_ coordinate format to reference so, maybe we don't need the conditional and only need the `setItem()` call. 🤔

jasontk19 (changes requested):

minniewong:
This work is captured in this PR: https://github.com/nasa-gibs/worldview/pull/3946

web/js/components/location-search/coordinates-dialog.js
ryanweiler92: Added a componentWillUnmount method to deregister the event.

web/js/components/location-search/coordinates-dialog.js
```diff
@@ -22,8 +26,13 @@ class CoordinatesDialog extends Component {
...
   }
 
+  updateCoordinateFormat = () => {
+    this.setState({ updateCoords: true });
```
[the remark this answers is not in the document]
ryanweiler92: removed this state variable and replaced with a formattedCoordinates variable which the render() and renderCopyToClipboardButton() functions read.

web/js/components/location-search/coordinates-dialog.js
```diff
@@ -174,6 +184,7 @@ class CoordinatesDialog extends Component {
           <CopyClipboardTooltip
             tooltipToggleTime={tooltipToggleTime}
             clearCopyToClipboardTooltip={this.clearCopyToClipboardTooltip}
+            updateCoords={updateCoords}
```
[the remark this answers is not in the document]
ryanweiler92: removed this variable entirely

web/js/modules/settings/actions.js
ryanweiler92: You're right, I was able to remove the conditional statement as there are no situations where we are not passing a value to the action creator.

PR #3945, opened 2022-08-05
WV-2285: Don't allow shortening URLs that are too long

Tomcariello:
## Description

Fixes # .

A URL over 2048 characters cannot be shortened via the link shortener. I have added code to check the length of the current URL & suppress the Shorten Link component if the URL length exceeds 2048.

## How To Test

To test you need to exceeds 2048 characters in the URL. To do so, load a fresh instance & select "Add Layers" at the bottom of the window on the left side. If you enable every layer option in Carbon Monoxide, Corrected Reflectance, Dust, and then just 1 option in Fire & Thermal anomalies the URL will exceed 2048. (You can check the URL length as you add layers with window.location.href.length in the console) Once you have exceeded 2048, close the layers window & click the "Share" option on the top-right of the screen. "Shorten URL" & the checkbox will not be rendered.

Now, you can delete any layer (click the 'x' on any item in the left window) & the 'Shorten Link" option will render once the URL is again below or equal to 2048 characters.

You can re-add layers to confirm that the rendering is behaving appropriately as the URL length changes.

### DIFF
--- a/e2e/features/share/share-test.js
+++ b/e2e/features/share/share-test.js
@@ -64,6 +64,24 @@ module.exports = {
     c.expect.element('#rd-share').to.be.present;
     c.expect.element('#email-share').to.be.present;
   },
+  'Clicking Shorten link works with links less than 2049 characters': (c) => {
+    const shortQueryString = '?l=Reference_Labels_15m,Reference_Features_15m,Coastlines_15m,VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2022-08-10-T15%3A15%3A05Z';
+    c.url(c.globals.url + shortQueryString);
+    c.waitForElementVisible(shareToolbarButton, TIME_LIMIT);
+    c.click(shareToolbarButton);
+    c.waitForElementVisible('#wv-link-shorten-case', TIME_LIMIT);
+    c.click('#wv-link-shorten-case');
+    c.assert.cssClassPresent('#wv-link-shorten-case', 'checked');
+  },
+  'Clicking Shorten link is refused with links greater than 2048 characters': (c) => {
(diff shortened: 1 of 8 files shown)


web/js/containers/share.js
```diff
@@ -243,15 +246,15 @@ class ShareLinkContainer extends Component {
               Copy URL to share link.
             </p>
             {' '}
-            {urlShortening && (
+            {urlShortening && showShortenUrlCbx ? (
```
jasontk19, line 249: I think you should be able to omit the tertiary here and just check for 
```{urlShortening && showShortenUrlCbx && (...)}```

**Review by jasontk19 — CHANGES_REQUESTED** [review]
The PR titles will get pulled verbatim to generate the release notes ([seen here](https://github.com/nasa-gibs/worldview/releases/tag/v3.29.0)) so it's helpful to format them consistently and provide a very brief description of the fix as the title, eg:
```WV-2285: Don't allow shortening URLs that are too long```

It may be useful to simply disable the checkbox and show a tooltip on hover that explains why it is disabled (e.g. `URL is too long to shorten`) rather than hide it entirely.  If so, look for uses of `<UncontrolledTooltip>` from `reactstrap` throughout the app for examples of how to set this up.  I will leave it up to @minniewong to make the final call on this.

File: web/js/containers/share.js (hunk 1 above)
Comment (Tomcariello) [reply web/js/containers/share.js]: Agreed & addressed in b583bc219db4798d42179f84898a0be041efc8f1.

File: web/js/containers/share.js [hunk 2]
```diff
@@ -247,9 +250,10 @@ class ShareLinkContainer extends Component {
               <Checkbox
                 label="Shorten link"
                 id="wv-link-shorten"
-                onCheck={this.onToggleShorten}
+                onCheck={!preventShorten ? this.onToggleShorten : null}
                 checked={isShort}
-                disabled={!shortLink.isLoading}
+                disabled={!shortLink.isLoading && preventShorten}
+                title="Link cannot be shortened at this time"
```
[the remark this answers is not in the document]
Tomcariello: Agreed & addressed in 669ec943e10e863a3a74f8f5d2d1a5684071126b. Thanks!

jasontk19 (approved):
Looks good to me!

minniewong (approved):
looks good!

jasontk19:
When merging, please use the "Squash and merge" option.

e2e/features/share/share-test.js
```diff
@@ -64,6 +64,24 @@ module.exports = {
...
+  'Clicking Shorten link works with links less than 2049 characters': (c) => {
+    const shortQueryString = '?l=Reference_Labels_15m,Reference_Features_15m,Coastlines_15m,VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2022-08-10-T15%3A15%3A05Z';
+    c.url(c.globals.url + shortQueryString);
+    c.waitForElementVisible(shareToolbarButton, TIME_LIMIT);
+    c.click(shareToolbarButton);
+    c.waitForElementVisible('#wv-link-shorten-case', TIME_LIMIT);
+    c.click('#wv-link-shorten-case');
+    c.assert.cssClassPresent('#wv-link-shorten-case', 'checked');
+  },
+  'Clicking Shorten link is refused with links greater than 2048 characters': (c) => {
```
jasontk19, line 76: Nice!

PR #3947, opened 2022-08-10
WV-2310: Add facet for imagery type (granule, vector, raster)

jasontk19:
## Description

- Remove unused 'variable' layer type
- Update layer config schema
- Add imagery type facet

## How To Test

* Checkout `granule-configs` branch
* `npm run build`
* Checkout this branch
* `npm run watch`
* Open product picker
* Confirm new "Imagery Type" facet shows at the bottom and layers can be filtered by their type: Raster, Vector, Granule


minniewong (approved):
Looks good - again, we may have to change the word "Granule" to something else at some point, but I'm still not sure we've decided yet what it should be. So leave it as it is for now!

PR #3950, opened 2022-08-12
WV-2229: Fixing animation issues

ryanweiler92:
## Description

When an animation is created, the number of frames is determined by the function getNumberOfSteps which loops through the selected start and end date and returns a number based on the interval. (IE. January 1st to March 1st with Month intervals returns 3). This has to be equal to the number of dates that are pushed to the buffer array in the playQueue file. With the way that the getNumberOfSteps() function was set up, it was adding an extra frame when the next date started before but went beyond the end date. (IE. If you selected January 1st to March 2nd, it would return 4 frames instead of 3 where the last date would be April 1st) In the play-queue file, we filter out any dates that extend beyond the end date when we push to the bufferArray of dates. This would mean that in the previous example that we were expecting 4 frames but only had 3 dates in the bufferArray. This is what lead to the 'determining buffer size' loader being stuck on the screen. I made a simple change to the getNumberOfSteps function in modules/animation/util.js. It now includes a check in each loop for a nextDate variable to see if it extends beyond the end date. This way we always have the same number of frames and number of dates in the buffer array. 

The last issue that remained was if a user selected an interval like month, but selected 2 dates that were less than one month apart. If a user pressed the play animation button in this scenario, the determining buffer size loader would pop up with no explanation. I changed the getNumberOfSteps to return 1 frame if the start and end date did not have a separation that matched the selected interval. The animation-widget component was already set up to disable the animation play button if the number of frames was 1.

*This is the PR that behaves like how we outlined in the dev meeting*

## How To Test

Open the animation selector and select the monthly interval. Then select a date range such as January 1st 2022 through March 3rd 2022. Now the animation will work instead of stalling at the determining buffer size loader. You can set the date range to January 1st to January 15th while keeping the monthly interval. The play button will be disabled since there is less than a month between these dates.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3259,7 +3259,6 @@
     },
     "node_modules/@webassemblyjs/ast": {
       "version": "1.9.0",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@webassemblyjs/helper-module-context": "1.9.0",
@@ -3269,22 +3268,18 @@
     },
     "node_modules/@webassemblyjs/floating-point-hex-parser": {
       "version": "1.9.0",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/@webassemblyjs/helper-api-error": {
       "version": "1.9.0",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/@webassemblyjs/helper-buffer": {
       "version": "1.9.0",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/@webassemblyjs/helper-code-frame": {
       "version": "1.9.0",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@webassemblyjs/wast-printer": "1.9.0"
@@ -3292,12 +3287,10 @@
     },
     "node_modules/@webassemblyjs/helper-fsm": {
       "version": "1.9.0",
-      "dev": true,
       "license": "ISC"
     },
     "node_modules/@webassemblyjs/helper-module-context": {
       "version": "1.9.0",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@webassemblyjs/ast": "1.9.0"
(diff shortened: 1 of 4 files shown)


jasontk19 (changes requested):
It looks like you have conflicts due to updating `chromedriver` directly rather than rebasing with develop where it has been updated.  I'd suggest undoing that last commit and rebasing with develop.

The `snappedCurrentDate` variable represents the frame date in the animation that is closest to the current date the user has selected on the timeline.  I believe it is used to determine where to start playback so that a user can begin playback from any point within the animation start/end date range.  It appears that, after your changes, playing an animation will always "rewind" (so to speak) to the beginning start day before beginning playback.  We should find a solution that preserves this behavior if possible.

ryanweiler92:
I was able to reimplement the snapped date property so that users can play an animation from custom points in their date range. I had to add an additional check to the conditional in the getStartDate() function of playqueue to check that the next date after the snapped date was not beyond the end date.

ryanweiler92:
> It's working well on most circumstances but I noticed an issue with leap years. If you create an [animation starting 2012 JAN 29 to 2013 DEC 31](http://localhost:3000/?z=2&i=2&as=2012-01-29-T16%3A00%3A00Z&ae=2013-12-31-T16%3A00%3A00Z&ab=on&t=2012-01-29-T17%3A59%3A59Z) where 2012 is a leap year and 2013 is not a leap year. When you reach 2013 FEB, it changes the date to 2013 FEB 28 and the animation continues using the 28th. If you then instead use the right arrow button to progress through the dates monthly, starting 2012 JAN 29, when you reach 2013 FEB "29" the timeline changes to 2013 MAR 01. In this instance, the animation and the timeline are behaving differently. I believe we should be defaulting to the latter behaviour as it is covered in the [Test Suite -> Timeline -- Ranges section](https://wiki.earthdata.nasa.gov/display/WOR/S-06+%5Bdate%5D%3A+Timeline). We can discuss this at a meeting if it doesn't make sense/or see if we want to revise the behaviour.

We have decided to move this issue to another ticket due to the scope and behavior pre-existing before my changes. I will handle this on ticket WV-2385.

minniewong (approved):
Looks great!

ryanweiler92:
I have reviewed this again and this is behaving like we wanted it to based on our discussion from the dev meeting. A timeframe of January 31st to March 31st will return 3 frames of animation on Jan 31st, Feb 28th & March 28th. This PR also includes the underlying changes that actually makes the animation work.

jasontk19 (approved):
👍🏻

PR #3951, opened 2022-08-17
WV-2281: Standalone about page

jasontk19:
## Description

- Add a URL param to indicate if about page is open or not
- Clean up some places where "Worldview" was hard-coded into the about sections

## How To Test

* Load http://localhost:3000/?abt=on
* Confirm about page shows
* Close about page
* Refresh page
* Confirm about page does not show


minniewong (commented):
Looks good! Do you think we need to include a link somewhere in the app about page, that say something like "open in a new page"?

jasontk19:
> Looks good! Do you think we need to include a link somewhere in the app about page, that say something like "open in a new page"?

I don't think so but, I don't fully understand this feature request in the first place. 🤷🏻

minniewong (approved):
Looks like it works as expected! I will share a screenshot of it to the chat to double check it meets their needs

PR #3952, opened 2022-08-17
WV-2385: Fixing date interval logic in animation to be consistent with timeline date interval logic

ryanweiler92:
## Description

Clicking through timeline dates at a monthly interval would return different dates than playing an animation at a monthly interval when including the month of February. This is because the date interval logic for these components are separate. They had different ways of handling the month of February which only has 28 or 29 days. I updated the animation date interval logic to match the timeline date interval logic.  

## How To Test

This requires the fix from branch WV-2229. If you start on January 31st 2017 and change the interval selector to monthly, the next date will March 3rd. Now if you play an animation at a monthly interval starting on January 31st 2017, the next date will also be March 3rd. Prior to this fix, if you played an animation on January 31st 2017, the next date would be February 28th.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3259,7 +3259,6 @@
     },
     "node_modules/@webassemblyjs/ast": {
       "version": "1.9.0",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@webassemblyjs/helper-module-context": "1.9.0",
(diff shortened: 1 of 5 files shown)


jasontk19:
I think we need to talk through expected behavior for this, maybe at the dev meeting tomorrow.  Jumping from January 31st to March 3rd does not feel like a "month" to me.

ryanweiler92:
This PR is the one where I made changes so that the animation date intervals will act the same as the timeline. From what we talked about in the dev meeting, this is NOT the behavior we want. Please refer to https://github.com/nasa-gibs/worldview/pull/3950.

minniewong:
> This PR is the one where I made changes so that the animation date intervals will act the same as the timeline. From what we talked about in the dev meeting, this is NOT the behavior we want. Please refer to #3950.

Should we delete this PR then?

PR #3953, opened 2022-08-19
WV-2386: Fix adding layers

jasontk19:
## Description

* clone layers object that gets modified for product picker so the `type` property used by the layer builder is correct

## How To Test

* Open product picker
* Add layer(s)
* Confirm imagery shows for layer(s)


minniewong (approved):
👍

PR #3955, opened 2022-08-19
WV-2380: Update event strings with constants

Tomcariello:
## Description
All usage of the event service (/web/js/util.events.js) have been updated from passing strings as arguments to instead pass constants which are defined in the new file /web/js/util/constants.js. 

This update does not change any behavior. It is a case of applying appropriate coding standards.

## How To Test
Since no behavior changed, I expect that the standard test suite will expose any errors.
Each update can be reviewed via "Files changed" to confirm that the old string has been appropriately replaced with the corresponding constant value.


jasontk19 (approved):
Looks good!

PR #3956, opened 2022-08-23
WV-2388

jasontk19:
## Description

Prevent measure vectors from processing as running data

## How To Test

* Create a measurement
* Hover over the measurement line(s)
* See console error


minniewong (approved):
🚀

PR #3959, opened 2022-08-25
WV-2203: Prevent starting or completing a measurement when out of map bounds

Tomcariello:
## Description
Fixes WV-2203

Prior to this update, if a user clicked outside of the map bounds during a line or polygon measurement the measurement would fail with NaN. This update discards invalid clicks during a measurement which allows the measurement to continue properly.

## How To Test
Load WV, click the ruler on the bottom right, select either "Measure Distance" or "Measure Polygon".
- Test a measurement while staying within map bounds.
- Test a measurement with click(s) outside of the map bounds.
- Try to start/end a measurement with click(s) outside of the map bounds.

These steps should be repeated for both "Measure Distance" or "Measure Polygon" AND should be tested in each of the 3 projections (Arctic, Geographic, Antarctic).

@nasa-gibs/worldview


minniewong (commented):
Measurements look great in Geographic; I had a question about polar measurements. I am able to ATTEMPT to draw an area measurement outside of the map bounds, but not distance measurements. Should it be disabled similarly for area? 

Screen Shot 2022-08-30 at 9 55 25 AM (2)

Tomcariello:
@minniewong Good eye! This is addressed is the latest update.

web/js/components/measure-tool/measure-tooltip.js
```diff
@@ -101,16 +104,34 @@ export default function MeasureTooltip(props) {
...
-    <div className={`tooltip-measure tooltip-custom-black ${activeStaticClass}`}>
-      <span dangerouslySetInnerHTML={{ __html: getMeasurementValue() }} />
-      {!active && (
-        <span className="close-tooltip" onClick={onRemove} onTouchEnd={onRemove}>
-          <FontAwesomeIcon icon="times" fixedWidth />
-        </span>
-      )}
-    </div>
-  );
+  const checkGeographicCoordValidity = (val) => val.indexOf('NaN') < 0;
```
jasontk19, line 107: Instead of having to add all this code here, couldn't you just modify the `drawChangeListener` in `ol-measure-tooltip` to check the `tooltipCoord` in a similar way to how you are checking the mouse coordinates in the `condition` function that was added to the draw interaction?  For example: 
```
// ol-measure-tool line 200
    drawChangeListener = feature.getGeometry().on('change', (e) => {
      const geom = e.target;
      if (geom instanceof OlGeomPolygon) {
        tooltipCoord = geom.getInteriorPoint().getCoordinates();
      } else if (geom instanceof OlLineString) {
        tooltipCoord = geom.getLastCoordinate();
      }

     // checking if the tooltip coordinate is in the extent before we render the tooltip and update its position
      const tCoord = transform(tooltipCoord, crs, 'EPSG:4326');
      if (!areCoordinatesWithinExtent(proj, tCoord)) return;

      renderTooltip(feature, tooltipOverlay);
      tooltipOverlay.setPosition(tooltipCoord);
    });
  ```

jasontk19: I took a closer look at this and it wasn't as straightforward as I had hoped so I think we are good to stick with your current approach!

Tomcariello: Sounds good. Once the CI tests all pass I will merge. Thanks.

jasontk19: You may want to re-request review from Minnie and get her approval on the PR first

jasontk19 (approved):

minniewong (approved):
Looks great! 👍

PR #3963, opened 2022-08-31
Remove OL graticule code

jasontk19:
## Description

Removes dead code from when we used OpenLayers to render a graticule rather than using a GIBS layer.

## How To Test

- Try image download with the new graticule loaded
- Try gif creation with the new graticule loaded

@nasa-gibs/worldview


minniewong (approved):
yay! Look good

PR #3965, opened 2022-09-07
WV-2395: Update link for MODIS_True_Color.pdf

minniewong:
## Description

Fixes WV-2395.
With the Earthdata website change, the former URL for the MODIS_True_Color.pdf no longer worked. The URL was updated for "Why are some layers not available?" in the data download tab, as well as some of the layer descriptions. 

@nasa-gibs/worldview


Tomcariello (approved):
These links all load correctly now. Looks good to me!

PR #3966, opened 2022-09-07
WV-2393 Setting area of interest without moving bounding box

ryanweiler92:
## Description

When adding a layer, checking the 'Set Area of Interest' box and then clicking 'Download via Earthdata Search' without moving the box, you would be linked to a new page that would present an error saying an unknown error had occurred. This was happening because the SmartHandoff component was only rendering the coordinates and boundaries necessary to generate a working download link when the bounding crop box was moved or resized. More specifically, the currentExtent state property was being left blank if you pressed the download link before moving or resizing the bounding box. This was fixed by adding calling the `onBoundaryChange` function within the `onCheckboxToggle` function. A conditonal check within `onBoundryChange` was necessary since the `boundaries` object already had the correct properties. 

## How To Test

1. Load any aerosol layer such as 'Deep Blue Aerosol Angstrom Exponent (Land and Ocean)'.
2. Click the data tab
3.  Select the Near Real-Time selection of the Aerosol layer.
4. Check the 'Set Area of Interest' box but do (!not!) move or resize the bounding box.
5. Click the 'Download via Earthdata Search' button'.
6. Click continue on the 'Leaving Worldview' modal.
7. Verify that the link does not provide an error message.

### DIFF
--- a/web/js/containers/sidebar/smart-handoff.js
+++ b/web/js/containers/sidebar/smart-handoff.js
@@ -229,7 +229,7 @@ class SmartHandoff extends Component {
     }
   }
 
-  initalBoundary(boundaries, setExtent) {
+  setInitialExtent(boundaries, setExtent) {
     const { proj, map, selectedCollection } = this.props;
 
     const lonlats = imageUtilGetCoordsFromPixelValues(
@@ -269,7 +269,7 @@ class SmartHandoff extends Component {
         event: 'smart_handoffs_toggle_true_target_area',
       });
       this.setState({ showBoundingBox: true });
-      this.initalBoundary(boundaries, true);
+      this.setInitialExtent(boundaries, true);
     } else {
       this.setState({
         showBoundingBox: false,
--- a/web/js/containers/sidebar/smart-handoff.js
+++ b/web/js/containers/sidebar/smart-handoff.js
@@ -158,19 +158,24 @@ class SmartHandoff extends Component {
 
     if (!selectedCollection) return;
 
-    const {
+    let newBoundaries = boundaries;
+
+    let {
       x,
       y,
       width,
       height,
-    } = boundaries;
+    } = boundaries
 
-    const newBoundaries = {
-      x,
-      y,
-      x2: x + width,
-      y2: y + height,
-    };
+    if(boundaries.hasOwnProperty('width')){
+
+      newBoundaries = {
+        x,
+        y,
+        x2: x + width,
+        y2: y + height,
+      };
+  }
 
     const lonlats = imageUtilGetCoordsFromPixelValues(
       newBoundaries,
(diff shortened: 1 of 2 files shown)


web/js/containers/sidebar/smart-handoff.js
```diff
@@ -229,16 +229,47 @@ class SmartHandoff extends Component {
     }
   }
 
+  initalBoundary(boundaries, setExtent) {
+    const { proj, map, selectedCollection } = this.props;
+
+    const lonlats = imageUtilGetCoordsFromPixelValues(
```
jasontk19, line 235: A lot of this code looks like the same as what we have in `onBoundaryChange()`.  Do you think you could find a way to refactor this a little so it can be re-used?

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -229,16 +229,47 @@ class SmartHandoff extends Component {
     }
   }
 
+  initalBoundary(boundaries, setExtent) {
```
jasontk19, line 232: This is minor but, it might be helpful to call this `setInitialBoundary` (or even `setInitialExtent` since it ends up calling `this.updateExtent`).  The `set` prefix helps make it more obvious at a glance that the end result of this fn is to set a state property.

jasontk19 (changes requested):
This is working well and I think after talking through it earlier, this definitely looks like the right approach 👍🏻   It might also be worth considering some small changes in `web/js/modules/smart-handoff/util.js` to make sure we never send over parameter values of `undefined` to Earthdata Search since it (reasonably) didn't seem to like that very much.  If you do, I think it would be worth also throwing an error to make sure we don't just silently skip over something that could represent a bug.

web/js/containers/sidebar/smart-handoff.js
ryanweiler92: I ended up finding a way to simply add a conditional check to `onBoundaryChange` so that it does not update the boundaries object if it already has the correct properties. Now in the `onCheckboxToggle` I just call `onBoundaryChange`. I also added a check in the util file of smart-handoffs in the modules directory. I looped through the options object and consoled any keys that were undefined or empty objects.

web/js/containers/sidebar/smart-handoff.js
```diff
@@ -158,19 +158,23 @@ class SmartHandoff extends Component {
...
       height,
     } = boundaries;
 
-    const newBoundaries = {
-      x,
-      y,
-      x2: x + width,
-      y2: y + height,
-    };
+    if (Object.prototype.hasOwnProperty.call(boundaries, 'width')) {
```
jasontk19, line 170: I don't think you want to use the [`call` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) here.  You could check if the `boundaries` object has a `width` property with `Object.prototype.hasOwnProperty` but, I think the simplest way to check if `boundaries.width` is defined would just be to evaluate for truthiness, which should be easy to do inline with the `newBoundaries` object definition:

```
const newBoundaries = {
  ...boundaries
  x2: width ? x + width : boundaries.x2,
  y2: height ? y + height : boundaries.y2,
}
```

ryanweiler92: I was originally just using `Object.prototype.hasOwnProperty` but linter did not like it unless I added the `.call`. I think you are right though, the best way seems to be to do the check within defining the object. I've updated like you suggested above.

jasontk19 (approved):
Looks great!

PR #3969, opened 2022-09-12
WV-2298: Hide datelines and text if not in geographic projection

jasontk19:
## Description

* Updated to check the projection to determine when to hide lines and text
* Use map.un to deregister listeners (map.off is not a thing)

## How To Test
* Switch between projections and try to get datelines to show in polar projections


minniewong (approved):
looks like it's working!

PR #3973, opened 2022-09-13
WV-2297: Improve granule dateline crossing behavior

jasontk19:
## Description

* Don't request as many granules from CMR
* Search granules more efficiently for dates that have matching imagery to fix UI sluggishness
* Change dateline crossing behavior so that if granules for more than one date are visible, they always get shifted to the right
* Increase number of simultaneously loading map tiles to 32 (from 16)

## How to test

- Checkout `granule-configs` branch
- `npm run build`
- Checkout this branch
- `npm run watch`


minniewong (approved):
Looks good ⚡

PR #3976, opened 2022-09-20
WV-2396: Dateline tooltip hover delay

ryanweiler92:
## Description

When browsing granules it can be difficult to view granules very close to the dateline because every time you move the mouse over the dateline the tooltip will move to your current mouse position. This update changes the logic so that the tool tip only moves to your mouse position if you hover over the dateline for at least 500ms. 

I felt that the Line component that this logic is located in was a good component to update from a class component to a function component since it can utilize several modern hooks. This will improve performance and make it easy to make future changes to this component. 

I have also create a customHook file located in the util directory. I added a custom hook for getting the previous value of a prop or state property. This is very useful as unlike class components, function components do not have the built in ability to pass `prevProps` in a `componentDidUpdate` lifecycle function.

## How To Test

Go to the global settings and set the dateline to `always` show. Hover over the dateline for 500ms or more and observe how the tooltip does not move right away. If you move your mouse over the dateline without hovering over the tooltip, the tooltip should not move at all.

### DIFF
--- a/web/js/components/dateline/line.js
+++ b/web/js/components/dateline/line.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef} from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import OlOverlay from 'ol/Overlay';
 import util from '../../util/util';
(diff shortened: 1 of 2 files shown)


web/js/components/dateline/line.js
```diff
@@ -15,136 +16,126 @@ const lineStyles = {
...
 
-  mouseOutHidden = () => {
-    this.setState({ textActive: false });
-  }
+  const mouseOverHidden = (e) => {
+    const coords = map.getCoordinateFromPixel([e.clientX, e.clientY]);
+    timerRef.current = setTimeout(() => {
+      setTextCoords(coords);
+      toggleTextActive(true);
+    }, 500);
```
jasontk19, line 73: After playing around with this some I think we can drop this timeout to 300ms.  Otherwise this looks good to go 👍🏻

jasontk19 (approved):

PR #3977, opened 2022-09-20
WV-2441: Fix layer picker modal scrolling

ryanweiler92:
## Description

Previously the entire layer picker modal would scroll when trying to scroll the contents. With this update you can now individually scroll the facets contents and the layer contents of the layer picker modal.  

## How To Test

Click add layer, press the filter button, observe that you can scroll the facets section and layers section separately.

### DIFF
--- a/web/scss/components/facets.scss
+++ b/web/scss/components/facets.scss
@@ -4,7 +4,6 @@
   display: flex;
   flex-direction: column;
   overflow-y: auto;
-  height: 90vh;
 
   & .bag-o-chips {
     background: #333;
@@ -101,7 +100,6 @@
   max-width: 100%;
   display: flex;
   flex-direction: column;
-  height: 100% !important;
 
   & .filter-chip {
     font-size: 16px;
--- a/web/scss/components/facets.scss
+++ b/web/scss/components/facets.scss
@@ -3,6 +3,8 @@
   max-width: 270px;
   display: flex;
   flex-direction: column;
+  overflow-y: auto;
+  height: 90vh;
 
   & .bag-o-chips {
     background: #333;
@@ -99,6 +101,7 @@
   max-width: 100%;
   display: flex;
   flex-direction: column;
+  height: 100% !important;
 
   & .filter-chip {
     font-size: 16px;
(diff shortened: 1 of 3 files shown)


jasontk19 (changes requested):
This looks better but, the initial categories view still does not scroll within the modal like it used to.  Also, we used to keep a little extra space at the bottom, presumably to see the timeline.  I don't know if that was intentional or not but, check out how it looks in production.  I could see making the height fill the window as being useful as well though.

Screen Shot 2022-09-20 at 3 42 49 PM

web/scss/components/facets.scss
```diff
@@ -99,6 +101,7 @@
   max-width: 100%;
   display: flex;
   flex-direction: column;
+  height: 100% !important;
```
jasontk19, line 104: In general, we should try to avoid using `!important` as much as possible and consider it a last resort.  It usually just makes changes things later more difficult.

web/scss/features/layers.scss
```diff
@@ -386,9 +386,14 @@
   color: #ccc;
 }
 
+.layer_picker_component .modal {
+  max-height: 100vh !important;
```
jasontk19, line 390: Same thing here about the `!important`.  I would look at how the styles are set up in prod and determine what style changes caused this to break and see if they can be reversed.  That way we can avoid having to add additional styles and forcing them into place with `!important`

ryanweiler92: My bad I totally forgot to fix the initial selection screen on the modal. It should be identical to the way it looks in production now. I went through the layers and facets scss files and tried to remove as many `!important` as I could.

minniewong (approved):
I didn't even notice the issue before, so it seems to look fine :)

ryanweiler92:
Ok apologies for the repeated breakpoint issue. I finally found the root of the issue which was how I was setting the class name for the modal in footer-content.js. I made sure to only set mobile classes for mobile devices (not smaller desktops) that way the breakpoints work correctly on all sizes for desktop.

jasontk19 (changes requested):
There is now some extra horizontal space in the portrait mobile view:

Screen Shot 2022-09-21 at 4 27 22 PM
Screen Shot 2022-09-21 at 4 26 55 PM

jasontk19 (approved):

PR #3978, opened 2022-09-21
WV-2439: Map zoom buttons showing on mobile

ryanweiler92:
## Description

The zoom buttons were showing in mobile views when they should be hidden. This was a side effect of changing the mobile detection library. The zoom buttons are now hidden in mobile views. I also fixed the placement of the rotation buttons on polar projections.

## How To Test

Open a mobile view and observe that the zoom buttons are not visible.
View a polar projection and observe the placement of the rotation buttons.


minniewong (approved):
No more zoom buttons and rotation buttons look good!

jasontk19 (approved):

PR #3983, opened 2022-09-22
WV-2448: Change default Corrected Reflectance layer & initial map position

Tomcariello:
## Description
Terra will be adjusting its orbital configuration on October 12, 2022 at which point it will not produce data. We need to alter the following in WV:

- Change the default Corrected Reflectance layer from Terra/MODIS to Aqua/MODIS.
- Adjust the default map position to be more eastward to compensate for Aqua/MODIS later overpass time compared to Terra.
- Update default Natural Event preference from Terra/MODIS to Aqua/MODIS.
- Tour stories: change the default layer for the Intro to Worldview story from Terra/MODIS to Aqua/MODIS.

## How To Test
- Load the PR.
- Run `npm run build:config` to update the configuration.
- Load WV & confirm:
  - The default Corrected Reflectance Layer is loading Aqua/MODIS properly.
  - The map view is sufficiently eastward, meaning graphics fill the majority of the visible map on initial load
  - All "Events" are defaulting to Aqua/MODIS in the layer selector
  - Tour Stories: Intro To Worldview is using Aqua/MODIS Layer

@nasa-gibs/worldview

### DIFF
--- a/config/default/common/config/wv.json/defaults.json
+++ b/config/default/common/config/wv.json/defaults.json
@@ -3,11 +3,11 @@
         "projection": "geographic",
         "startingLayers": [
             {
-                "id": "MODIS_Terra_CorrectedReflectance_TrueColor"
+                "id": "MODIS_Terra_CorrectedReflectance_TrueColor",
+                "hidden": "true"
             },
             {
-                "id": "MODIS_Aqua_CorrectedReflectance_TrueColor",
-                "hidden": "true"
+                "id": "MODIS_Aqua_CorrectedReflectance_TrueColor"
             },
             {
                 "id": "VIIRS_SNPP_CorrectedReflectance_TrueColor",
--- a/config/default/common/config/wv.json/naturalEvents.json
+++ b/config/default/common/config/wv.json/naturalEvents.json
@@ -56,7 +56,7 @@
                 "Wildfires": [
                     [
                         "MODIS_Terra_CorrectedReflectance_TrueColor",
-                        true
+                        false
                     ],
                     [
                         "MODIS_Terra_CorrectedReflectance_Bands721",
@@ -68,7 +68,7 @@
                     ],
                     [
                         "MODIS_Aqua_CorrectedReflectance_Bands721",
-                        false
+                        true
                     ],
                     [
                         "VIIRS_NOAA20_CorrectedReflectance_TrueColor",
@@ -142,11 +142,11 @@
                     ],
                     [
                         "MODIS_Terra_CorrectedReflectance_TrueColor",
-                        true
+                        false
                     ],
                     [
                         "MODIS_Aqua_CorrectedReflectance_TrueColor",
-                        false
+                        true
                     ],
                     [
                         "VIIRS_NOAA20_CorrectedReflectance_TrueColor",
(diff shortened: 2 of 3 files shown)


jasontk19:
Some things to consider, probably for @minniewong to answer:

- Do we change the defaults for event layers?
- Do we change the defaults for tour story layers?

minniewong:
> Some things to consider, probably for @minniewong to answer:
> 
> * Do we change the defaults for event layers?
> * Do we change the defaults for tour story layers?

Sigh. We probably should change the default layer for events that use Terra CR; for Tour stories, luckily we'd only have to change the default layer for the Intro to Worldview story.

jasontk19:
@minniewong  It looks like most of the tour stories have `MODIS_Terra_CorrectedReflectance_TrueColor` in their permalinks.  Sometimes it's hidden but sometimes not.  Is it because they usually only point to dates in the past that we don't need to update these?

minniewong:
> @minniewong It looks like most of the tour stories have `MODIS_Terra_CorrectedReflectance_TrueColor` in their permalinks. Sometimes it's hidden but sometimes not. Is it because they usually only point to dates in the past that we don't need to update these?

Yep - the outage will start 12 October 2022 and last over 10 days, so past dates will still show imagery, so it's fine.

jasontk19:
@minniewong How far east should the default be shifted?  What's here now looks good to me but also, at this time of day a lot more imagery coverage is available.  I think this is fine but I defer to you on the positioning.

minniewong (approved):
Everything looks good. I checked the map timing last night and this morning, and they both seem far east enough at the relevant UTC time. Thanks also for updating the tour story and events imagery.

jasontk19:
For reference here are the same links in prod:
[0000Z](https://worldview.earthdata.nasa.gov/?now=2022-09-27T00%3A00%3A00Z&t=2022-09-26-T00%3A00%3A00Z)

[0300Z](https://worldview.earthdata.nasa.gov/?now=2022-09-27T03%3A00%3A00Z&t=2022-09-26-T00%3A00%3A00Z)

Tomcariello:
I have adjusted the values to better align with the needs of the Aqua/MODIS Corrected Reflectance layer. I have discarded the old values, which were tuned for the Terra/MODIS Corrected Reflectance layer. 

The comments at lines 113-119, describing an intention to adjust based on the time of day, are no longer applicable/accurate. The goal of this function is to return values that result in a map with current imagery & a minimal amount of missing/upcoming 
imagery. There is no longer any manipulation of the map location based on the time of day.

jasontk19:
I think this looks ok now but, I am not sure if we do want to keep the behavior that shifts the map westward when loading at 0000Z.  Will defer to @minniewong

minniewong:
> I think this looks ok now but, I am not sure if we do want to keep the behavior that shifts the map westward when loading at 0000Z. Will defer to @minniewong

We'd only want the shifting behaviour during the time that Terra is not working (10+ days after 10 October). When Terra comes back, we'd want to return to how it used to work.

Tomcariello:
There was time based behavior applied to Terra (at 00:00 UTC, start at far eastern edge of map; at 23:00 UTC, start at far western edge of map) which is not being applied to the Aqua CR layer. The only shifting applied to the Aqua layer is moving the map to the east to ensure that the map imagery is populated.

Once we switch back to Terra all of this code will be reverted & the time based adjustments will be re-applied & the eastward shift removed.

minniewong (approved):
Thanks for the info! That sounds good

PR #3984, opened 2022-09-23
WV-2427: Drop python 2 check

jasontk19:
## Description
* Drop python 2 check, assume python 3
* Update pipenv version to latest

## How To Test

* `npm run python` 
* Confirm latest pipenv and dependencies install succesfully
* `npm run build`
* `npm run start`
* Confirm app loads and works as expected


ryanweiler92 (approved):
This works for me.

Tomcariello (approved):
This works for me on my system which does not have Python 2 installed. Looks good!

PR #3992, opened 2022-09-30
Remove tile transition time except for granule layers

jasontk19:
## Description

Only fade in tiles for granule layers


minniewong (approved):
Looks great! Thanks

Issue #555, opened 2017-09-28
Timeline: better support for differing temporal ranges (e.g., hourly, monthly, 8-day)


minniewong:
To get it to work with sub-daily products (#460, #461)

rboller:
Includes #610, #460, #461

ZachTRice:
#610 is a duplicate of this ticket. Since that ticket has a long thread of information, I am closing this ticket and will use that one instead. @minniewong , @rboller feel free to re-open this ticket if I am misreading this as a duplicate.

PR #3993, opened 2022-10-03
WV-2145: Mobile Animation Feature

ryanweiler92:
## Description

This update allows for animation creation within mobile views. The mobile animation frame limit have been set to 50 frames to account for limited bandwidth on mobile devices. The animations in mobile views have limited features and do not include GIF creation. 

Added a mobile animation button which is limited to mobile views that brings up a new mobile animation widget. After selections on this widget are selected, it can be minimized and the animation can be played with the collapsed animation widget. 

Overhauled the conditional styling for the animation button, animation play button (collapsed animation widget) and date arrows so that they render in proper positions depending on breakpoints and subdaily mode. 

## How To Test

1. Launch WV
2. Open the DevTools
3. Select any mobile view
4. The new mobile animation button should appear which brings up the new mobile animation widget.
5. The mobile animation widget should work in any mobile view including smaller desktop views.
6. To test the animation frame limit, go to [localhost:3000/?v=-127.54084611130202,-31.196051270164425,-36.29880238885806,62.96630766900102&z=4&ics=true&ici=5&icd=10&as=2021-12-03-T16%3A00%3A00Z&ae=2021-12-03-T17%3A10%3A00Z&l=GOES-East_ABI_GeoColor,Coastlines_15m&lg=true&al=true&ab=on&t=2021-12-01-T20%3A10%3A00Z](url)
7. Open the animation widget and decrease the year by 1 to 2021.
8. Observe that the mobile animation play button to  play the animation does not appear when the frame limit is exceeded.


Tomcariello (commented):
I see a "Set up animation" tooltip after tapping the camera to trigger the mobile animate feature and when clicking the "Loop" button. Do either have a hover effect/tooltip? These could well be an artifact of browser mobile emulation; will need to test on a real mobile device. 

image

Tomcariello:
There are some active console.log() calls:

- Remove console.log() at line 72 of mobile-date-picker.js
- Remove console.log() at lines 175-179 of play-queue.js

Tomcariello:
UI thoughts:

The Interval Selector (Day/Month/Year) select menu:
- Should have a min-width. It currently adjusts the size to accommodate the length of the word selected which causes the UI to jump around a bit.
- Should have rounded corners to match the "Loop" button.
- Should be horizontally level with the Interval Selector inputs. They are very close to aligned but off by a few pixels with the Loop button being _slightly_ lower.

Remove `border: 4px solid #555;` from `.wv-animation-widget-wrapper-mobile` to fix the barely visible rounded corner issue.

image

Tomcariello:
There is sufficient space to reclaim screen space without impacting the usability of the animation feature. Reclaiming this space will show more of the map to the user. Something like this (apologies for the poor graphics work here)

image

Tomcariello:
Can we style this window to match Search/Share/Global Settings/etc.? The top of the window should be a slightly different gray with a thin line separator. Perhaps insert "Animation Setting" as the header & then the down arrow on the right side?

image

web/js/containers/animation-widget.js
```diff
@@ -292,6 +317,13 @@ class AnimationWidget extends React.Component {
       + `${hasSubdailyLayers ? 'subdaily ' : ''}`
       + `${isMobile ? 'mobile ' : ''}`
       + `${isLandscape ? 'landscape ' : ''}`;
+    const subdailyID = hasSubdailyLayers ? '-subdaily' : '';
+
+    const widgetIDs = (isMobilePhone && isPortrait) || (screenWidth < 670 && hasSubdailyLayers) || (screenWidth < 575 && !hasSubdailyLayers) ? `collapsed-animate-widget-phone-portrait${subdailyID}`
```
jasontk19, line 322: This is another very verbose nested definition that could be broken down and simplified.  Are any of these checks the same across components?  Can a helper function be created somewhere else to avoid redefining these conditions?

web/scss/features/anim-widget.scss
```diff
@@ -357,3 +408,227 @@ a .wv-animation-widget-icon {
...
+      }
+
+      .wv-icon-case {
+        height: 45px;
+        width: 52px;
+      }
+
+      & .mobile-date-picker-select-btn {
+        position: relative;
+        background-color: #295f92;
```
jasontk19, line 509: This button color should probably be a variable.

jasontk19 (changes requested):
A few behavior and style related things:

- There is no feedback to let the user know they have exceeded the limited number of frames allowed, the play button just disappears.  We need to find a way to convey this information.  
---
Screen Shot 2022-10-06 at 9 14 54 AM

If these two items are going to be side by side, I think the corners shouldn't be rounded where they touch so that they appear as a "button group".  Also, the number input should not be colored the same as the buttons.  It should probably have a dark grey/black background.

---

Screen Shot 2022-10-06 at 9 17 45 AM
I'm not sure the transparent background works well for this after all, it looks too busy in this example when you can see all the UI elements behind it.

---

Screen Shot 2022-10-06 at 9 29 45 AM
Screen Shot 2022-10-06 at 9 30 10 AM
The collapse icon we use here should be the same we use elsewhere on mobile for consistency.  Also, this is minor but, the border line separating the header from the body should match other modals.  If we are going to make them look similar we should use the exact same styling where possible.

e2e/features/animation/mobile-animation-test.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+
+  'Minimizing mobile animation widget opens collapsed animation widget': (c) => {
+    c.useCss().click('.wv-minimize');
+    c.waitForElementVisible('#collapsed-animate-widget-phone-portrait', TIME_LIMIT);
+  },
+
+  'Playing the animation changes the date of the mobile date picker': (c) => {
+    c.useCss().click('#collapsed-animate-widget-phone-portrait');
+    c.pause(3000);
```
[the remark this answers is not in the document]
ryanweiler92: I was able to apply all of the recommended changes from this thread but I did leave this one instance of a long pause. This one is because I need to wait for the animation to load and play and I'm trying to account for poor connections when it may take longer to load despite the fact that it is a smaller animation.

jasontk19:
Screen Shot 2022-10-11 at 10 21 52 AM
Screen Shot 2022-10-11 at 10 22 00 AM
Screen Shot 2022-10-11 at 10 23 18 AM


If we are going to style the modal after other modals we should be more consistent with how we use the header and how the styles match up.  The header in other modals is used to display a descriptive title of the modal contents, and the "X" to close is much smaller and thinner.  I don't think the header is where we want to display an error message, perhaps below the header inside the modal body is better?

ryanweiler92:
I've updated the position of the frame limit warning message to move it from the header to within the modal content. I've also updated the styling of the widget header to be more in line with our other modals within the app.

Tomcariello:
When testing in Chrome, if I set the viewport to Samsung Galaxy S8+ or iPhone SE the FPS slider extends off the side of the screen. Galaxy Fold also does not render properly but in different ways.

web/js/containers/timeline/timeline.js
```diff
@@ -941,6 +971,29 @@ class Timeline extends React.Component {
...
+            screenWidth={screenWidth}
+            isMobilePhone={isMobilePhone}
+            isMobileTablet={isMobileTablet}
+            isLandscape={isLandscape}
+            isPortrait={isPortrait}
+            clickAnimationButton={this.clickAnimationButton}
+            hasSubdailyLayers={hasSubdailyLayers}
+            disabled={animationDisabled}
+            label={
+isCompareModeActive
```
Tomcariello, line 988: Fix indenting

web/js/components/timeline/timeline-controls/animation-button.js
```diff
@@ -1,38 +1,75 @@
...
+    } if (isMobilePhone && isLandscape) {
+      return `phone-landscape${subdailyID}`;
+    } if ((isMobileTablet && isPortrait) || (!isMobilePhone && screenWidth < breakpoints.small)) {
+      return `tablet-portrait${subdailyID}`;
+    } if (isMobileTablet && isLandscape) {
+      return `tablet-landscape${subdailyID}`;
+    }
+  };
+
+  const className = getButtonClassName();
```
Tomcariello, line 37: I would prefer the "className" variable have a different name since "className" is a JSX term. btnClassName? btnClassSuffix?

Tomcariello (commented):
Just some very minor feedback here. Everything seems to work correctly. Looks good!

jasontk19:
I think this looks good.  The only comment I have is that the "X" to close is not the same style as what's in other modals.  Can we use the same thing?

Screen Shot 2022-10-19 at 1 43 42 PM
Screen Shot 2022-10-19 at 1 40 05 PM

jasontk19 (approved):

minniewong (approved):
Looks great and seems to work well!

PR #3995, opened 2022-10-03
WV-2464: Dateline text initially active on hover setting

ryanweiler92:
## Description

When the `Show Antimeridian / Approximate Date Line` setting was set to `On Hover`, the dateline text was still showing on the map on the initial load. Updated the logic so that if the setting is set to `On Hover`, the text field only appears when the dateline is hovered.

## How To Test

1. Open the global settings menu and select `On Hover` for the setting `Show Antimeridian / Approximate Date Line` setting.
2. Zoom out to see where the datelines would appear.
3. Observe that the dateline text fields do not appear unless you have hover over the dateline.

### DIFF
--- a/web/js/components/dateline/datelines.js
+++ b/web/js/components/dateline/datelines.js
@@ -9,7 +9,7 @@ import { CRS } from '../../modules/map/constants';
 
 function DateLines(props) {
   const {
-    map, proj, date, isCompareActive, mapIsRendered, alwaysShow, hideText,
+    map, proj, date, isCompareActive, mapIsRendered, alwaysShow, hideText, isMobilePhone, isMobileTablet,
   } = props;
 
   if (!mapIsRendered) return null;
@@ -89,6 +89,8 @@ function DateLines(props) {
         date={date}
         textCoords={textCoords}
         setTextCoords={setTextCoords}
+        isMobilePhone={isMobilePhone}
+        isMobileTablet={isMobileTablet}
       />
       <Line
         id="dateline-right"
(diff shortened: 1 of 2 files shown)


Tomcariello (approved):
This solves the problem in the ticket so I am approving. However, there are a few other minor issues with this feature. I'm not sure if they should be rolled into this ticket or new ticket(s) created...?

1) You need to hover directly over the Antimeridian to see the date indicators but if you hover a few pixels to the right of the Antimeridian the dates do not show while the Antimeridian does. I think we should show the Antimeridian & dates in tandem on hover; there should not be any separation between the two.

2) If you have the Antimeridian setting to Always show the dateline on initial page load & then change the setting to Hover the Antimeridian is removed but the date indicators remain. The dates should be removed once the user changes this setting.

ryanweiler92:
> This solves the problem in the ticket so I am approving. However, there are a few other minor issues with this feature. I'm not sure if they should be rolled into this ticket or new ticket(s) created...?
> 
> 1. You need to hover directly over the Antimeridian to see the date indicators but if you hover a few pixels to the right of the Antimeridian the dates do not show while the Antimeridian does. I think we should show the Antimeridian & dates in tandem on hover; there should not be any separation between the two.
> 2. If you have the Antimeridian setting to Always show the dateline on initial page load & then change the setting to Hover the Antimeridian is removed but the date indicators remain. The dates should be removed once the user changes this setting.

I was able to fix both of these issues. The text was activating when the mouse hovered over the hidden line which caused the disparity. I also fixed one of the useEffect functions so that the date text goes away when you switch settings.

Tomcariello:
Looks good to me!

jasontk19 (approved):
👍🏻

PR #3997, opened 2022-10-05
WV-2398: Fix iOS browser crashes

jasontk19:
## Description

Changes were introduced in OL 6.14.2 that causes browsers to crash on iOS for us after reaching a certain number of layers rendered.  See https://github.com/openlayers/openlayers/issues/14171

The inclusion of the `className` property when creating layers on our end prevents canvas re-use between layers which causes a large performance hit, leading to the crash on iOS devices.  Removing `className` fixes the issue.

## How To Test

* Pull changes
* `npm ci`
* Open the app on an actual iOS device
* Load any fire event (this will add many layers)
* Confirm the app does not crash

You will need to find a way to hit your local dev server from an iOS device, such as using a reverse proxy like [ngrok](https://ngrok.com/).  This error did not seem to happen in simulated devices and definitely won't occur in a desktop browser set to mobile size.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "worldview",
-      "version": "3.32.0",
+      "version": "3.34.0",
       "hasInstallScript": true,
       "license": "NASA-1.3",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


ryanweiler92 (approved):
I was able to verify that this works on my personal iOS mobile device going through ngrok.

PR #4001, opened 2022-10-05
update chromedriver and point nightwatch.conf files to animation folder

ryanweiler92:
github tests


jasontk19:
Does this need to be a PR or can it just be a branch?

PR #4002, opened 2022-10-12
WV-1995: Prioritize References measurement group in Layer List

ryanweiler92:
## Description

This updates the order in which layers are added to give priority to the reference layers group. Now by default, new layers will always populate beneath the last reference layer if the `group similar layers` option is off or under the reference group if the option is on. If a user reorders a non-reference layer between or before the reference layers, any new layers added will still populate under the last reference layer.

## How To Test

1. Starting with the default reference layers, ensure that the `group similar layers` option is off and add any non-reference overlay layer. Observe that the layer is added under the last reference layer. 
2. Add another non-reference overlay layer and observe that this layer has also been added directly after the last reference layer. 
3. Move the most recent layer that you added to the 2nd position (behind coastlines) and then add another non-reference overlay layer. Observe that this layer is added to the 5th position (behind place labels). 
4. Check the `Group Similar Layers` checkbox and observe that each of the layer groups is ordered beneath the reference layer group.
5. Uncheck the `Group Similar Layers` checkbox and observe that the original layer order has been preserved.

### DIFF
--- a/e2e/features/image-download/layers-test.js
+++ b/e2e/features/image-download/layers-test.js
@@ -16,7 +16,7 @@ module.exports = {
   },
 
   'List layers in draw order': function(c) {
-    bookmark(c, startParams.concat(['l=MODIS_Terra_CorrectedReflectance_TrueColor,Reference_Features_15m,MODIS_Terra_Aerosol']));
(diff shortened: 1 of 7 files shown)


jasontk19 (approved):
I like how simple this solution is and I think it solves the problem well without overcomplicating the code or the behavior.

Tomcariello (approved):
Looks good.

minniewong (approved):
Looks great!

PR #4006, opened 2022-10-18
Increase timeouts for GC request during build

jasontk19:
## Description

This increases the timeout on requests made for the GetCapabilities documents, and related linked metadata such as colormaps, from GIBS during the build process which will occasionally cause build issues or failures.  This is even more likely to occur when on an EDGATE VPN or when using more than the three default GC documents (as when building against worldview-options-gibs environments).

## How to test

Attempt to build while on the VPN or with a throttled connection.


ryanweiler92 (approved):
Worked while on the VPN & using Network Link Conditioner with the setting `Very Bad Network`.

PR #4008, opened 2022-10-21
WV-2403: Show All Event Tracks Feature

ryanweiler92:
## Description

This feature adds the ability to view event tracks for all of the events on the map at the same time. This also adds a default layer (blue marble) when the events are activated in the sidebar. Clustering was removed on geographic projections for better visibility of tracks when several are grouped closely together. Event specific layers are now removed when deselecting an event as well. 

## How To Test

After checking out to the feature branch you will need to do a `npm run build` as the config files were. updated to include a default events layer.

The best way to test this feature would be to select events from both the map and sidebar while sometimes deselecting events from the sidebar and to make sure the following criteria is always met:

1. If events tab is active but no single event is selected, the default blue marble layer should appear.
2. No layer should appear above the blue marble layer unless it was specifically selected in the layers tab prior to activating the events tab and selecting an event.
3. Clicking any event from the map or sidebar should always show the event specific event layers and never the default blue marble layer.
4. Individual tracks should not disappear if the show all tracks option is selected and you unselect an individual event.
5. Leaving the events tab of the sidebar should make the blue marble layer invisible. 
6. Leaving the events tab with a single event selected should keep the event specific layers on.


minniewong (commented):
This works really well! Only one very minor nit, the checkboxes seem like they could be closer together, but I'm OK with them being like this also.

ryanweiler92:
> This works really well! Only one very minor nit, the checkboxes seem like they could be closer together, but I'm OK with them being like this also. Screen Shot 2022-10-21 at 10 47 22 AM

I've updated the spacing between the event checkbox options so that they are more close together.

minniewong (approved):

jasontk19 (commented):
If I have an overlay turned on before I switch to the events tab, it is not hidden:
Screen Shot 2022-10-21 at 2 43 40 PM

jasontk19:
There is an existing issue which is sort of highlighted by showing multiple tracks at once which is that, while the map is animating from one event to another, all the tracks and arrows become unattached from their reference point on the map and shift about visually in space.  It may be worth trying to hide them while the animation happens (like we do when you zoom in/out) if we can't find some other way to fix this.

Screen Shot 2022-10-21 at 2 46 59 PM

ryanweiler92:
@jasontk19 

In regards to the requested changes I have been able implement solutions to each.

1. I've updated a couple components so that layers will never show above the blue marble layer when the events tab is active even if the layer was selected before activating. Users can still see non-event layers over events when an individual event is selected. 
2. I updated the function that compares the event layers and original layers in the layers/selectors file to loop fewer times.
3. I found a good way to hide the event tracks & points while animating to an event in the event-tracks.js file.

jasontk19 (approved):
Looks good to me

PR #4009, opened 2022-10-21
WV-2479: Fix Reference Layer Ordering

ryanweiler92:
## Description

This ensures that the default reference layers are added in their original and correct order while still adding new layers under the reference layers or reference layer group.

## How To Test

Open an instance of WV and check that the reference layers are loaded in the following order:
1. Place Labels,
2. Coastlines / Borders / Roads
3. Coastlines

Add a non-reference layer and verify that the layer is rendered below the reference layers.


minniewong (approved):
Looks good!

jasontk19 (approved):

PR #4012, opened 2022-10-24
WV-2485: Fix baselayer ordering/visibility 

ryanweiler92:
## Description

This fixes the logic for adding layers so that new overlays will be visible above base layers when there are no reference layers selected.

## How To Test

1. Start with either only the top baselayer visible (see link below)
2. Add any overlay (e.g. an Aerosol Optical Depth layer)
3. Observe that the newly added overlay is visible

[(http://localhost:3000/?v=-457.6410612408573,-294.48706705194684,480.0464387591427,457.01293294805316&l=VIIRS_NOAA20_CorrectedReflectance_BandsM11-I2-I1,VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_NOAA20_CorrectedReflectance_BandsM3-I3-M11(hidden)&lg=false&t=2022-10-07-T00%3A03%3A00Z)]

### DIFF
--- a/e2e/features/image-download/layers-test.js
+++ b/e2e/features/image-download/layers-test.js
@@ -16,7 +16,7 @@ module.exports = {
   },
 
   'List layers in draw order': function(c) {
-    bookmark(c, startParams.concat(['l=MODIS_Terra_CorrectedReflectance_TrueColor,MODIS_Terra_Aerosol,Reference_Features_15m']));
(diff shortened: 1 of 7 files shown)


jasontk19 (approved):
👍🏻

Tomcariello (approved):
Looks good.

PR #4013, opened 2022-10-24
WV-2397: Update date intervals for date arrows on mobile views

ryanweiler92:
## Description

The date arrows were not updating by custom intervals when adding a geostationary layer while in a mobile view. This update fixes this so that the arrows will increase/decrease time by the custom value determined by the Geostationary layer.

## How To Test

1. Open WV in a mobile view
2. Add any geostationary layer (ex. GeoColor)
3. Press the back arrow on the date arrows and verify that the date is decreasing by 10 minute intervals.
4. Remove the geostationary layer
5. Verify that the date increases/decreases by days again.

### DIFF
--- a/web/js/containers/timeline/timeline.js
+++ b/web/js/containers/timeline/timeline.js
@@ -43,7 +43,7 @@ import {
   selectDate as selectDateAction,
   changeTimeScale,
   selectInterval,
-  changeCustomInterval,
+  changeCustomInterval as changeCustomIntervalAction,
   updateAppNow,
   toggleCustomModal,
   triggerTodayButton,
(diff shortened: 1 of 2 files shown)


minniewong (approved):

Tomcariello (approved):
👍

web/js/containers/timeline/timeline.js
```diff
@@ -217,6 +222,19 @@ class Timeline extends React.Component {
...
+    const subdailyRemoved = !hasSubdailyLayers && prevProps.hasSubdailyLayers;
+    const subdailyInterval = customInterval > 3 || interval > 3;
+
+    if (subdailyRemoved && subdailyInterval) {
+      changeCustomInterval();
+      selectInterval(1, TIME_SCALE_TO_NUMBER.day, false);
+    }
+
+    if (subdailyAdded && !customSelected) {
+      changeCustomInterval(10, TIME_SCALE_TO_NUMBER.minute);
```
jasontk19, line 235: This also happens [here](https://github.com/nasa-gibs/worldview/blob/main/web/js/components/timeline/custom-interval-selector/custom-interval-selector.js#L49) which means that, at least on desktop, we dispatch this same action twice when a subdaily layer is added.  I'm guessing we can take the code out of that file and dispatching it here will work on both desktop and mobile?

jasontk19 (changes requested):
This works though it causes some duplicate actions to fire.  We should see if we can avoid that.

web/js/containers/timeline/timeline.js
ryanweiler92: I've updated the CustomIntervalSelector component to remove the duplicate logic for changing intervals. The actions were no longer necessary in this component since it is always happening in the timeline parent component now. There should be no more duplicate actions.

jasontk19 (approved):

PR #4019, opened 2022-10-31
wv-2496

Tomcariello:
## Description
Revert wv-2448 (reset default layer to Terra/MODIS).

Fixes #
Flip back to having Terra/MODIS as the default layer on load
Also flip back the default events layers that were changed to Aqua/MODIS
Change the map positioning back to the original location based on time.

## How To Test
Clone, build, run WorldView. Confirm that the default layer is TERRA/MODIS. Also check & confirm that natural events are using Terra/MODIS as the layer.

### DIFF
--- a/config/default/common/config/wv.json/defaults.json
+++ b/config/default/common/config/wv.json/defaults.json
@@ -3,11 +3,11 @@
         "projection": "geographic",
         "startingLayers": [
             {
-                "id": "MODIS_Terra_CorrectedReflectance_TrueColor",
-                "hidden": "true"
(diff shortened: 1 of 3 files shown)


jasontk19 (approved):

minniewong (changes requested):
The Fire events didn't revert back to Terra

jasontk19:
Looks like there is an additional commit to be reverted: https://github.com/nasa-gibs/worldview/commit/a3c1d2b2a10c4d7f15b36b6cfe02bab8da5a9b7d

minniewong (approved):
Looks good now :)

PR #4020, opened 2022-11-02
WV-2478: Dynamic granule layer start/end date handling

jasontk19:
### WV-2477
* make sure dates show properly in coverage panel
* make sure dates don't line break in layer info panel

### WV-2478
* Remove handling for projection-specific date ranges (was needed when GIBS served date ranges for each individual granule)
* Assume granule layers without a defined end date are ongoing and have available imagery up until current time

## How to Test

* Checkout this branch
* Run `NODE_ENV=development npx webpack` to build the UI based on this branch
* Place [this pre-built configuration file](https://github.com/nasa-gibs/worldview/files/9924051/wv.json.zip) into the `web/config/` directory (overwrite what is there) after the build
* `npm run watch` to serve the build with the pre-built config file in place

### WV-2477
* Check timeline date coverage panel and layer info panel
* Confirm start-end dates fit into a single line and there are no line breaks (as seen in the ticket)

### WV-2478
This is a little trickier to really put to the test since, if you just built, the dates in the GC will be fairly recent.  To really test this, you probably want a modified `wv.json` with start/end dates and date ranges that are somewhat in the past in order to test the 15 day rolling date window more effectively.  Note that, on the configuration branch, the rolling window was set to 14 days since going back 15 days introduced a bit of time with no imagery.

* Open the app with a granule layer
* Browse around the dates where coverage begins 
* Browse around the dates where coverage ends (most current time)
* Confirm imagery works almost to the end of the coverage limits
* Note that there are some limitations and that the limits cannot be exact at this time (see internal ticket comments for more details)


ryanweiler92 (approved):

minniewong (approved):
🚀

PR #4025, opened 2022-11-04
WV-2094: Show layer collection version and type in sidebar

jasontk19:
## Description

- pull collection version/type from headers
- display in layer row

## Todo

- [x] How do we handle data in the map "wings"?  Since we load and tiles for technically three days at a time for some layers, is there a chance that a user looks at some data in the left wing which is from an NRT layer while middle or right wing is a STD layer?
- [ ] How do we handle when tiles are pulled from the browser cache?  If you load the app and all the tiles for a layer are loaded from the cache, there is no way to know what the `layer-identifier-actual` header would have been were it requested from the server.

## How To Test

1. Build against the BEST endpoint in an environment that supports the custom tile headers (GITC UAT) 
2. Add a layer that we know should should sometimes return NRT and sometimes STD depending on the date selected
  * An example layer could be `VIIRS_NOAA20_Clear_Sky_Confidence_Day`
4. Choose the most recent possible date of imagery which should be NRT
  * For the example layer, `2022-07-07` appears to be NRT (as of 2022-11-9)
5. Confirm that the version and "NRT" show in the layer info row of the sidebar
6. Change the date to something earlier that still has coverage
7. Confirm that the sidebar now shows "STD" (the version may or may not change depending on layer


minniewong (approved):
Great! Point to develop though

PR #4038, opened 2022-12-06
WV-2507: Implementing mobile animation widget updates

ryanweiler92:
## Description

Implementing the suggested updates to the mobile animation widget.

## How To Test

-`git checkout wv-2507`
-`npm run watch`
- Open the dev tools and switch to one of the iPhone views.
- Open the animation widget.
- Verify that the body of the widget is better centered and looks like the suggestions from the JIRA ticket WV-2467.


Tomcariello (changes requested):
iPhone SE (375x667) does not render well. Every other device I tested looks great.

Tomcariello (approved):

minniewong (approved):
Looks great!

PR #4053, opened 2022-12-13
WV-2528: Removing embed tab and updating related tests

ryanweiler92:
## Description

This removes the embed tab from the share modal. Also updates all of the tests related to embed.

## How To Test

1. `git checkout wv-2528`
2. `npm run watch`
3. Click the share button.
4. Verify that the embed tab is not there.


minniewong (commented):
Verified that I don't see the embed tab anymore; we may also want to update the text in the "Introduction to Worldview" tour story to remove the reference to the embed mode. I can do it and push to your branch if you would like.

ryanweiler92:
> Verified that I don't see the embed tab anymore; we may also want to update the text in the "Introduction to Worldview" tour story to remove the reference to the embed mode. I can do it and push to your branch if you would like.

Good call, I just removed the text in the last commit.

minniewong (approved):
thanks!

minniewong:
Should we also hide/remove: https://github.com/nasa-gibs/worldview/blob/main/doc/embed.md?

ryanweiler92:
> Should we also hide/remove: https://github.com/nasa-gibs/worldview/blob/main/doc/embed.md?

I added that markdown file to our `gitignore` file and removed it from the cache. So we should be able to just remove that line from the `gitignore` file if we want to bring it back. Worst case we can just make a new one.

minniewong:
> > Should we also hide/remove: https://github.com/nasa-gibs/worldview/blob/main/doc/embed.md?
> 
> I added that markdown file to our `gitignore` file and removed it from the cache. So we should be able to just remove that line from the `gitignore` file if we want to bring it back. Worst case we can just make a new one.

nice!

PR #4027, opened 2022-11-16
WV-2374: Creating React Components for Map UI

ryanweiler92:
## Description

This update replaces the files previously called `ui.js` and `combine-ui.js` with React components in the new MapUI folder. The behavior of WV should be entirely the same. Changing these into React components will make the logic in these files much easier to maintain and read when it requires changes or updates.

## How To Test

- Checkout to branch `wv-2445`
- `npm run watch`
- Test as much functionality as you can to make sure that WV still operates entirely the same. The scope of these files is very wide ranging and includes many aspects of WV such as updating projections, dates, markers, layers, opacity etc...

PR #4069, opened 2022-12-21
WV-2374: React Map Component 

ryanweiler92:
## Description

This is an updated branch of [the original React map component PR](https://github.com/nasa-gibs/worldview/pull/4027) that includes the latest changes from our develop branch. I made a new branch so that I can preserve the old one in case we run into any issues. I tested this pretty thoroughly but it's probably worth one more look. 

## How To Test
1. `git checkout wv-2374-updated`
2. `npm ci`
3. `npm run build`
4. `npm run watch`
5. Everything should work the same. You should be able to see all of our changes we made in develop. There should be no performance issues.


minniewong (approved):
Everything seems to work really well!

PR #4072, opened 2022-12-28
WV-2536 Fixing granule layers getting stuck

ryanweiler92:
## Description

This fixes a bug that was causing the initial granule layer that gets added to the map when you add a granule layer to stay on the map after you changed dates. So there would be multiple granule layers visible at one time (one would be with incorrect dates). 

1. `git checkout wv-2536-fixing-granules`
2. `npm run watch`
3. Add any granule layer
4. Move the date backwards by one day. 
5. Verify that the map rerenders and removes the previous layer granule layer. 

This is an image of what it would look like after adding a granule layer and changing the date. It should no longer appear like this.
stuck-granules-example

### DIFF
--- a/web/js/mapUI/components/layers/addLayer.js
+++ b/web/js/mapUI/components/layers/addLayer.js
@@ -27,7 +27,7 @@ const AddLayer = (props) => {
     if (action.type === layerConstants.ADD_LAYER) {
       const def = lodashFind(action.layers, { id: action.id });
       if (def.type === 'granule') {
-        granuleLayerAdd(def);
+        return granuleLayerAdd(def);
       }
       clearPreload();
       addLayer(def);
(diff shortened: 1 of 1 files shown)


Tomcariello (changes requested):
This is not working for me. I decrement by DAY & I get the new imagery but the old imagery persists.

Is there something I need to do aside from running 'npm run build' prior to testing?

ryanweiler92:
> This is not working for me. I decrement by DAY & I get the new imagery but the old imagery persists.
> 
> Is there something I need to do aside from running 'npm run build' prior to testing?
@Tomcariello 
Hmm I just tried it again and it seems to be working for me. Can you try one more time with these instructions? 
1. `git checkout wv-2536-fixing-granules`
2. `npm ci`
3. `npm run build`
4. `npm run watch`
5. Change date to December 31st 2022
6.  Add any granule layer
7. Move the date backwards by one day. (Careful that you are selecting day because the default value for the date arrows will be 10 minutes for this layer). 
8.  Verify that the map rerenders and removes the previous layer granule layer.

Tomcariello (approved):
I have re-tested on 2 different machines & everything is working as expected.

PR #4104, opened 2023-02-21
WV-2604: Reimplement embed mode with <object> tag

ryanweiler92:
## Description

This reimplements the embed element creation in the share menu. The iFrame tag has been replaced with the object tag and its properties updated accordingly.

## How To Test

1. `git checkout wv-2604`
2. `npm run watch`
3. Click the `share` button in the top right.
4. Select the embed tab. 
5. Verify that the element that is created looks similar to:
`<object type="text/html" data="http://localhost:3000/?lg=false&t=2023-02-21-T12%3A37%3A06Z&em=true" width="100%" height="100%" role="application"></object>`


ZachTRice (approved):
Looks good. Let's wait until after v4.0.0 is released to merge this into develop

PR #4106, opened 2023-02-21
WV-2605: Fixing layer ordering in compare mode after removing layers

ryanweiler92:
## Description

This fixes a bug that was causing layers to be added out of order when they were removed while compare mode was active and added again.

## How To Test

1. `git checkout wv-2605` 
2. `npm run watch`
3. Start compare mode
4. Remove all of the layer groups with the `Remove Group` button. 
5.  Add the Aerosol Optical Depth MODIS/Terra layer.
6. Add the Corrected Reflectance MODIS/Terra layer.
7. Click on the B side tab.
8. Remove all of the layer groups with the `Remove Group` button. 
9. Add the Aerosol Optical Depth MODIS/Aqua layer.
10. Add the Corrected Reflectance MODIS/Aqua layer.
11. Verify that the layers are shown in the correct order and you can still perform the swipe compare action.

### DIFF
--- a/web/js/containers/sidebar/layer-list.js
+++ b/web/js/containers/sidebar/layer-list.js
@@ -18,7 +18,7 @@ import {
 } from '../../modules/layers/selectors';
 import {
   reorderLayers as reorderLayersAction,
-  removeLayer as removeLayerAction,
+  removeGroup as removeGroupAction,
   toggleGroupVisibility as toggleGroupVisibilityAction,
 } from '../../modules/layers/actions';
 
@@ -43,7 +43,7 @@ function LayerList(props) {
     available,
     groupId,
     title,
-    removeLayers,
+    removeGroup,
     toggleVisibility,
     toggleCollapse,
     isMobile,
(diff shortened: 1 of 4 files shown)


Tomcariello (approved):

minniewong (changes requested):
If you do the steps for the A side - remove all layers, then add AOD, then add Corrected Reflectance (Where the AOD is correctly on top of the corrected reflectance) THEN replicate the same steps for the B side, the layers are also not ordered correctly and an error is returned. 
AssertionError
Call Stack
 Collection.assertUnique_
  wv.js?v=@BUILD_NONCE@:254325:15
 Collection.insertAt
  wv.js?v=@BUILD_NONCE@:254206:12
 addLayer
  wv.js?v=@BUILD_NONCE@:65365:31

ryanweiler92:
> If you do the steps for the A side - remove all layers, then add AOD, then add Corrected Reflectance (Where the AOD is correctly on top of the corrected reflectance) THEN replicate the same steps for the B side, the layers are also not ordered correctly and an error is returned. AssertionError Call Stack  Collection.assertUnique_   wv.js?v=@BUILD_NONCE@:254325:15  Collection.insertAt   wv.js?v=@BUILD_NONCE@:254206:12  addLayer   wv.js?v=@BUILD_NONCE@:65365:31

This actually exposed a different bug where the remove group button was dispatching the incorrect action. Instead of utilizing the `removeGroup` action for removing multiple layers, the `removeLayer` action was being used to map through the layers to remove. This was not being relayed correctly to the `removeLayer` component so the layers were being removed from the sidebar but not correctly removed from the open layers map object. I think this fix fits perfectly well in this branch. I have updated the testing description to account for this as well.

minniewong (approved):
Thanks for working through the bugs!

ZachTRice (approved):
Working as expected

PR #4107, opened 2023-02-21
WV-2602: Fix accessing links with 0 layers

ryanweiler92:
## Description

Currently when accessing UAT from the test [link perm.in.layers.3](https://worldview.uat.earthdata.nasa.gov/?l=&now=2013-07-01T12 ) which tests the ability to access links with 0 layers, WV will throw an error. This fixes this prevents this error so that you can access these links with 0 layers. 

## How To Test

1. `git checkout wv-2602`
2. `npm run watch`
3. Access this [link](http://localhost:3000/?l=&now=2013-07-01T12)
4. Verify that the map loads without errors and displays without any layers.


Tomcariello (commented):
The provided link loads correctly, but un-checking "Group Similar Layers" causes the app to crash.

PR #4108, opened 2023-02-22
WV-2607 Granule zoom issues on mobile

Tomcariello:
## Description
With granules loaded, zooming in on the map will sometimes result in incorrect pinch/swipe behavior. For instance, if a user tries to pan by swiping left-right with their finger the app will instead zoom. This behavior is most noticeable on my iPhone; Android seemed to work OK prior to the fix.

Steps to recreate: 

- Load a granule layer
- Turn off any other baselayer
- Two finger pinch and zoom FAR into the granules
- Two finger pinch to zoom and moving one finger to pan behave incorrectly.

Fixes WV-2607.

## How To Test

- git checkout wv-2607_granule_zoom_issues
- Connect your iPhone to access localhost ([for Windows](https://stackoverflow.com/questions/3132105/how-do-you-access-a-website-running-on-localhost-from-iphone-browser))
- Follow the steps to recreate above & confirm that the pinch/pan behavior is improved

@nasa-gibs/worldview

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -147,7 +147,7 @@
         "yargs": "^17.6.2"
       },
       "engines": {
-        "node": ">= 16"
+        "node": ">= 18"
       }
     },
     "node_modules/@ampproject/remapping": {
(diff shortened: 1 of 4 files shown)


ryanweiler92 (approved):
I ended up having to use [ngrok](https://ngrok.com/) to be able to access my local host instance. If anyone else has issues accessing localhost from your phone in the traditional way, ngrok was fairly simple to set up.

ZachTRice (approved):

PR #4132, opened 2023-03-07
WV-2433 On mobile, "Add Layer" and "Start Comparison Mode" should not stack

victorsh:
## Description

On mobile, in landscape view "Add Layer" and "Start Comparison Mode" buttons should not stack

These buttons shouldn't stack in landscape view because they end up consuming way too much vertical space.  The layout should change to show them side-by-side.

Fixes # WV-2433

## How To Test

- Open Chrome and turn on developer mode.
- Click the button on the top left of development mode to switch to mobile view.
- Click on the "layers" icon.

Screen Shot 2023-03-06 at 12 30 07 PM

@nasa-gibs/worldview

### DIFF
--- a/web/scss/features/sidebar.scss
+++ b/web/scss/features/sidebar.scss
@@ -359,7 +359,7 @@ footer .compare-toggle-button {
   }
 
   #products-holder footer .product-buttons {
-    height: 120px;
+    height: 70px;
   }
 }
 
--- a/web/scss/mobile.scss
+++ b/web/scss/mobile.scss
@@ -40,13 +40,11 @@
 
     button#layers-add,
     button#compare-toggle-button {
-      width: 100%;
-      position: relative;
-      left: 0;
+      width: 50%;
       height: 50px;
 
       span {
-        font-size: 18px;
+        font-size: 14px;
       }
     }
 


Tomcariello (approved):
We're talking about **portrait** mode, not **landscape** mode here, right? Assuming that's the case, this looks good to me.

minniewong:
> We're talking about **portrait** mode, not **landscape** mode here, right? Assuming that's the case, this looks good to me.

It's for landscape mode

minniewong (commented):
In Portrait mode, they also don't stack anymore. I can be convinced that we should go forward with this look :) 
Portrait: 
Screen Shot 2023-03-07 at 3 52 56 PM
Landscape: 
Screen Shot 2023-03-07 at 3 53 06 PM

victorsh:
Based on how I saw the SCSS configured. The layout of the two buttons is dependent on the width of the display. I took a bit of creative liberty here and made it so that the buttons are side by side. If this doesn't work it's no problem to change it so that the buttons are spaced opposite of each other.

Tomcariello:
> 

I see. My phone screen is large enough that the buttons are not stacked in production in landscape mode. With a smaller screen size I see the change impacts both portrait & landscape.

Tomcariello (approved):

minniewong (approved):
Thanks for making that edit!

PR #4136, opened 2023-03-08
WV-2437: On mobile, category label should not show on "Featured" tab

victorsh:
## Description

When the "Featured" tab is chosen in a mobile layout, the header element which would display the category title in other tabs (e.g. Air Quality seen below under Hazards and Disasters tab) should not be visible.  Right now it just shows "All"

Fixes # WV-2437

## How To Test

- Click on "Add Layers"
- Click the header dropdown and selected "Featured"
- "All" should no longer be there.

Screen Shot 2023-03-08 at 2 12 35 PM

@nasa-gibs/worldview

### DIFF
--- a/web/js/modules/product-picker/reducers.js
+++ b/web/js/modules/product-picker/reducers.js
@@ -1,4 +1,5 @@
 import googleTagManager from 'googleTagManager';
+import { get as lodashGet } from 'lodash';
 
 import {
   SAVE_SEARCH_STATE,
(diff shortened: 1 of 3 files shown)


web/js/modules/product-picker/reducers.js
```diff
@@ -166,7 +166,8 @@ export function productPickerReducer(state = productPickerState, action) {
       return {
         ...state,
         categoryType: 'featured',
-        category,
+        // category,
```
Tomcariello, line 169: I would remove this comment at line 169.

Tomcariello (approved):
Works as described.

web/js/modules/product-picker/reducers.js
victorsh: Removed

minniewong (approved):
Works now as expected!

PR #4147, opened 2023-03-13
WV-2438 On mobile, categories layout landscape could be improved

victorsh:
## Description

Now that we support mobile landscape view for larger devices, the categories layout could be updated to show two categories side by side since there appears to be plenty of horizontal real estate to support this.  Currently a single category takes up an entire row.

Fixes # WV-2438

The solution provided effects the styling of the `react-masonry-component`. The specific class that was modified was `.layer-category`. The CSS was changed so that the width of a category containing box resizes with the width of its container.

The main difference here between mobile view and desktop view is that the category modal is set to the width of screen on mobile, while the width in desktop is static except for breakpoints > 768 and > 1024

## How To Test

1. Open Chrome and turn on the developer console.
2. Set the developer console to mobile view.
3. Set the mobile device rotation to landscape.
4. Change the width of the device to <= 768px.
5. Category Items should be side-by-side.

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/layer/product-picker/browse/browse-layers.js
+++ b/web/js/components/layer/product-picker/browse/browse-layers.js
@@ -210,7 +210,7 @@ function BrowseLayers (props) {
               <CategoryGrid width={width} />
             </div>
           ) : renderContent()
-        }
+      }
     </>
   );
 }
--- a/web/js/components/layer/product-picker/browse/category-grid.js
+++ b/web/js/components/layer/product-picker/browse/category-grid.js
@@ -27,13 +27,9 @@ function CategoryGrid(props) {
     width,
   } = props;
   const setColumnWidth = (width) => {
-    console.log(width)
-    if (width >= 656) {
-      return 310;
-    } else {
-      return 0;
-    }
-  }
+    if (width >= 656) return 310;
+    return 0;
+  };
   const masonryOptions = {
     transitionDuration: '0.6s',
     columnWidth: setColumnWidth(width),
--- a/web/js/components/layer/product-picker/browse/category-grid.js
+++ b/web/js/components/layer/product-picker/browse/category-grid.js
@@ -26,9 +26,17 @@ function CategoryGrid(props) {
     categoryType,
     width,
   } = props;
+  const setColumnWidth = (width) => {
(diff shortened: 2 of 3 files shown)


minniewong:
Looks great at iPhone SE size:
iphoneSE

But iPad mini and iPhone 12 Pro seems to only show one column to the left. I made sure to refresh the page after changing the orientation, so I assume this is what it would look like on those devices. I know the breakpoints for those are > 768, so it seems it should be more likely to have the categories side-by-side?
ipadmini
iphone12pro

victorsh:
@minniewong I believe I've fixed the issue. `react-masonry-component` had options set to 310px column width. After I set this to 0, the width resizes correctly. After the device width passes 1024px, the category elements switch from a percentage width to a fixed width of 310px, thus the column width on the masonry component is unnecessary.

Tomcariello:
This may be out of scope of this ticket, but there is a different animation depending on how the user navigates the menus (testing iPhone 12 Pro).
If you go between 'Hazards & Disasters' & 'Science Disciplines' there is a grow style animation.
But if you go from 'Featured' to 'Hazards & Disasters' or 'Science Disciplines' the right column slides a bit from the right to the left.

Tomcariello (approved):

victorsh:
@Tomcariello I think what's happening with the animation is `react-masonry-component` is being unmounted/mounted when switching between 'Featured' and ('Hazards & Disasters' and 'Science Disciplines') or when first opening the category modal. I tried experimenting with some of the options for the component but nothing seems to have an effect on the animation. I tried `initLayout`, `resize`, `originLeft`, `originTop`.

```
export interface MasonryOptions {
    columnWidth?: number | string | HTMLElement | null;
    itemSelector?: string;
    gutter?: number | string;
    percentPosition?: boolean;
    horizontalOrder?: boolean;
    stamp?: string;
    fitWidth?: boolean;
    originLeft?: boolean;
    originTop?: boolean;
    containerStyle?: Object;
    transitionDuration?: number | string;
    resize?: boolean;
    initLayout?: boolean;
}
```

minniewong (approved):
Looks great with the new fixes

PR #4149, opened 2023-03-14
WV-2491 Fix granule highlight behavior on mobile

victorsh:
## Description

Currently, some weird behavior exists that causes map rotation.  Granule footprints should show and hide on a tap gesture.

Fixes # WV-2491

Granule selection was set to `onMouseMove`. The weird behavior was caused by the method checking for granule selection only triggering when the mouse was moved or the touch press was dragged.

## How To Test

Steps to recreate:

- Open Worldview on your phone
- Change to a polar projection, Arctic or Antarctic
- Tap on the Layers icon
- Tap on "Add Layers"
- Type in "granule" in the search bar, select the first one "Granule Corrected Reflectance (True Color) Suomi NPP/VIIRS"
- Tap the x in the upper right corner to close the Layer List
- Tap on a section of the map that has imagery, it should show a blue outline with the date and time inside the granule box.
- Tap on a different section of the map with imagery
- See error - it is difficult to select a different granule to highlight and show the blue outline and date and time. It does no always select and highlight a new granule that has been tapped on.

@nasa-gibs/worldview

### DIFF
--- a/web/js/containers/map-interactions/ol-vector-interactions.js
+++ b/web/js/containers/map-interactions/ol-vector-interactions.js
@@ -153,7 +153,7 @@ export class VectorInteractions extends React.Component {
 
   mouseMove({ pixel }, map, crs) {
     const {
-      isCoordinateSearchActive, measureIsActive, granuleFootprints, isMobile
+      isCoordinateSearchActive, measureIsActive, granuleFootprints, isMobile,
     } = this.props;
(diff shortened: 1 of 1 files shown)


minniewong (approved):
Can't test the tap behaviour until we get to UAT

Tomcariello (approved):

PR #4152, opened 2023-03-16
WV-2434 On mobile, sidebar and product picker landscape UI element sizing

victorsh:
## Description

Now that larger mobile devices use a mobile layout in landscape, there are a number of UI elements that should be tweaked for optimal sizing.  This is probably as simple as changing the media query breakpoint for some of these items:

most layer picker elements included facet/filter elements
sidebar layer elements
layer group headers 
layer row elements
etc

Fixes # WV-2434
Screen Shot 2023-03-17 at 9 19 08 AM
Screen Shot 2023-03-17 at 9 19 15 AM

## How To Test

- Open developer console
- Switch to device mode
- Try device Iphone 12 pro and Iphone Mini
- Also test window width and height at breakpoints 768px and 1024px

@nasa-gibs/worldview


minniewong (approved):
Looks great!

ryanweiler92 (approved):

PR #4159, opened 2023-03-21
WV-2621 On mobile, there are different animations depending on how the user navigates the menus

victorsh:
## Description

There appears to be different animation depending on how the user navigates the menus (testing iPhone 12 Pro).
If you go between 'Hazards & Disasters' & 'Science Disciplines' there is a grow style animation.
But if you go from 'Featured' to 'Hazards & Disasters' or 'Science Disciplines' the right column slides a bit from the right to the left.

more info: https://github.com/nasa-gibs/worldview/pull/4147 

Fixes # WV-2621

Set styling of GridCategory component based on which tab in the layer picker is selected. The styling sets the visibility and the display of div wrapping GridCategory. This way, GridCategory does not re-render so the animation triggers between 'Hazards & Disasters' & 'Science Disciplines' to/from 'Featured' and 'Recent'.

## How To Test

- npm ci
- npm run build
- In Chrome Dev Console, switch to mobile view and select 'Iphone 12'
- Open the layer picker
- Select 'Featured' then 'Hazards & Disasters' or 'Science Disciplines'
- Select 'Recent' then 'Hazards & Disasters' or 'Science Disciplines'
- The animation show trigger between tabs as each Category Item expands.

@nasa-gibs/worldview


minniewong (approved):
Looks good!

ryanweiler92 (approved):

victorsh:
@Tomcariello I fixed the error occurring when switching to 'Recent' then closing 'Add Layers'.

In terms of the animation, it looks like this is exhibiting the same behavior as what Worldview has in production currently.
I have yet to find a way to manually control the animations on the Masonry component. I was able to fix the animations here by keeping CategoryGrid mounted while switching between tabs. When first opening 'Add Layers' 'CategoryGrid' is mounted for the first time and the slide animation displays.

Tomcariello (approved):
Nice work!

PR #4148, opened 2023-03-13
WV-2578 Update (front-end) dependencies including React / Redux

victorsh:
## Description

As part of the build overhaul ticket, Zach updated the devDependencies (build dependencies) but the dependencies (frontend) were not updated.

Major packages that need updated include react / redux. To get a list of all outdated packages run `npm outdated`

Fixes # WV-2578

## Update Description

2 packages to note that required code changes are...
History (SCSS)
- added .nav-item a and .nav-item a:hover to set the <a></a> color to white.
@khanisak/temperature-converter (JS) https://github.com/nasa-gibs/worldview/blob/wv-2578/web/js/modules/settings/util.js
- The package updated the temperature conversion functions
rc-slider (JS)
- css import was changed
- Functions are now part of the Slider object. ex. Slider.createSliderWithTooltip, Slider.Range

## How To Test

Since this update touches on every aspect of the application, overall extensive testing would be required.
I've done my due diligence to touch on as many aspects of the application as possible but there may be bugs
that have slipped through the cracks.

@nasa-gibs/worldview

PR #4173, opened 2023-03-24
WV-2627 mobile warning alignment

Tomcariello:
## Description

Fixes # wv-2627

On mobile, the warning ("!" in the triangle) is misaligned in the "Recent" tab.

image

## How To Reproduce
- Add Layers --> Recent --> Clear List --> close (this isn't required, but will make it easier to see the problem layers)
- Add Layers --> Search for "Reflectance" & scroll down until you find some with the warning mark
- Click to add a couple options
- Go to "Recent" to see the layers you selected with the warning misaligned

## How To Test
- Follow the steps above & confirm the warning mark is properly aligned

Note: This error seems to be caused by a difference in the installed react-swipe-to-delete package (0.5.4) & the CDN react-swipe-to-delete CSS (current version, 1.0.8). This updates react-swipe-to-delete to 1.0.5 & removes the CDN link in favor of a local copy of the current CSS. This may need to be revisited/confirmed when [wv-2578 ](https://github.com/nasa-gibs/worldview/pull/4148) is approved as that also bumps this package version.

@nasa-gibs/worldview


minniewong (approved):
Looks great, thanks!

ryanweiler92 (approved):
The bug fix looks good to me. I think that your package-lock might be out sync though. When I pulled your branch I had to do an `npm i` before I could do an `npm ci`. It looks like this is why the end-to-end tests did not run either. I think you should be able to just go back into this branch and do an `npm i` (and `npm ci` again to be safe) and push the changes because that should modify the package-lock.

PR #4181, opened 2023-03-28
WV-2648 Modernize Measure Button

victorsh:
## Description

The MeasureButton component is currently written as a React Class component. Today's best practices for React suggest we should be using functional components whenever possible. Using the modern best practices for React, convert the MeasureButton component into a functional component.

1. Study the current behavior of the component so you know exactly what needs to be replicated.
2. Research [the differences between class & functional components and how to replace lifecycle hooks with useEffect hooks](https://www.scaler.com/topics/react/react-functional-vs-class-components/).
3. Use the [React docs](https://react.dev/reference/react) to find the best practices for updating state, using hooks & props.
4. Use other components from the WV repo that are currently written as functional components for reference.
5. If there are obvious opportunities to refactor code to be more concise or easier to read do so.
6. After refactoring into a functional component, test the component to make sure the behavior is still the same.

Fixes # wv-2648

## Changes
- Class based component was converted into a functional component
- Using redux hooks instead of redux function to map dispatch and props.

## How To Test

- `git fetch --all`
- `git checkout wv-2648-modernize-measure-button`
- `npm ci`
- `npm run watch`
- Click the measure button on the bottom left
- Click measure distance. Click on the map to create the first point. Click on the map again in a different location to create another point.
- Right click to cancel
- Click on the measure button
- Click on measure area. Click on the map to create the first point. Click on the map again in a different location to create another point.
- Right click to cancel
- Click on the measure button.
- Click the toggle from km to mi. Repeat steps for measuring distance and measuring area.
- Open Chrome developer console. Change to mobile view and select Iphone 12 pro.
- Follow the same steps as above but using touch input instead of mouse clicks.
- An alert should open explaining how to control each measure tool in mobile.

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/measure-tool/measure-button.js
+++ b/web/js/components/measure-tool/measure-button.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, memo, useMemo } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { Button, UncontrolledTooltip } from 'reactstrap';
 import googleTagManager from 'googleTagManager';
@@ -20,7 +20,7 @@ const MEASURE_MENU_PROPS = {
 const mobileHelpMsg = 'Tap to add a point. Double-tap to complete.';
 const helpMsg = 'Click: Add a point. Right-click: Cancel. Double-click to complete. ';
 
-const MeasureButton = function () {
+const MeasureButton = memo(function () {
   const dispatch = useDispatch();
   const [showAlert, setShowAlert] = useState(true);
   const [isTouchDevice, setIsTouchDevice] = useState(false);
@@ -42,18 +42,24 @@ const MeasureButton = function () {
     setShowAlert(false);
   };
 
-  const [isActive, isDistractionFreeModeActive, isMobile] = useSelector((state) => [
-    state.measure.isActive,
-    state.ui.isDistractionFreeModeActive,
-    state.screenSize.isMobileDevice,
-  ]);
+  const isActive = useSelector((state) => state.measure.isActive);
+  const isDistractionFreeModeActive = useSelector((state) => state.ui.isDistractionFreeModeActive);
+  const isMobile = useSelector((state) => state.screenSize.isMobileDevice);
 
-  const faSize = isMobile ? '2x' : '1x';
-  const shouldShowAlert = isActive && showAlert;
-  const message = isTouchDevice ? mobileHelpMsg : helpMsg;
+  const memoizedSelectors = useMemo(() => {
+    return {
+      isActive,
+      isDistractionFreeModeActive,
+      isMobile,
+    };
+  }, [isActive, isDistractionFreeModeActive, isMobile]);
+
   const buttonId = 'wv-measure-button';
   const labelText = 'Measure distances & areas';
-  const mobileMeasureButtonStyle = isMobile ? {
+  const faSize = memoizedSelectors.isMobile ? '2x' : '1x';
+  const shouldShowAlert = memoizedSelectors.isActive && showAlert;
+  const message = isTouchDevice ? mobileHelpMsg : helpMsg;
+  const mobileMeasureButtonStyle = memoizedSelectors.isMobile ? {
     bottom: '20px',
     fontSize: '14.3px',
     height: '44px',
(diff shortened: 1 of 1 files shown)


ryanweiler92 (changes requested):
Nice job on this one. I like your choice to go with the `useDispatch` and `useSelector` hooks for the Redux variables. There are only a few small and opinionated things I would like to see changed so that we can set a standard for going forward & modernizing other components. 

For the useSelector hook, lets use the object literal syntax. I think that this is easier to read will be useful if we need to pull two variables that have the same name. For example, we have several Redux variables called `isActive` so using the object literal syntax will allow us to rename these as we destructure them.  
`const { isActive, isDistractionFreeModeActive, isMobile } = useSelector(state => ({
    isActive: state.measure.isActive,
    isDistractionFreeModeActive: state.ui.isDistractionFreeModeActive,
    isMobile: state.screenSize.isMobileDevice,
  }));`
  
  I think we should also consolidate all of our Redux hooks at the top of the component followed by our local state variables. This way we always know where to look for these. 
  
```
const MeasureButton = function () {
  const dispatch = useDispatch();

  const { isActive, isDistractionFreeModeActive, isMobile } = useSelector(state => ({
    isActive: state.measure.isActive,
    isDistractionFreeModeActive: state.ui.isDistractionFreeModeActive,
    isMobile: state.screenSize.isMobileDevice,
  }));
  
  const [showAlert, setShowAlert] = useState(true);
  const [isTouchDevice, setIsTouchDevice] = useState(false);
```

victorsh:
@ryanweiler92 I've moved the Redux Selectors to the top as suggested and used the object literal syntax. I've also added memoization around the entire component as well as memoizing the Redux selectors. This might be a bit overkill for this component but I thought we could use this as an example when modernizing the rest of the components. This way, we would reduce the number of renders as much as possible and improve the overall performance of the app. It think we should have a discussion around this tomorrow during office hours.

I also agree with moving the redux selectors to the top of the component. Also tempted to move the rest of the variables above the function definitions so that there's a clear separation between the function and the variables.

victorsh:
Made one more update here. Using `shallowEqual` instead of `useMemo` since it looks much cleaner. It seems to work better in this case as well. I tested the number of renders by logging `render` within the component. Now it only renders once when the page is loaded, once when a the button is disabled and once when the coordinate points are finalized.

ryanweiler92 (approved):
I like the use of shallowEquals with the useSelector hook but I don't think that the memoization of this React component is necessary here. It is a good thought but we usually only want to memoize components if they were class components using the PureComponent subclass (this can be seen in ShareToolTips & CopyClipboardTooltip) or if you have a larger component that is constantly rendering the same props and we can see a measurable benefit to memoizing, otherwise it can lead to unwanted side effects. This is a really good article about memo https://dmitripavlutin.com/use-react-memo-wisely/#:~:text=2.1%20Component%20renders%20often%20with,render%20by%20a%20parent%20component. Other than that I am good with this one.

web/js/components/measure-tool/measure-button.js
```diff
@@ -21,110 +20,81 @@ const MEASURE_MENU_PROPS = {
 const mobileHelpMsg = 'Tap to add a point. Double-tap to complete.';
 const helpMsg = 'Click: Add a point. Right-click: Cancel. Double-click to complete. ';
```
Tomcariello, line 21: I would delete the space at the end of this string.

web/js/components/measure-tool/measure-button.js
```diff
@@ -21,110 +20,81 @@ const MEASURE_MENU_PROPS = {
...
+    isMobile: state.screenSize.isMobileDevice,
+  }), shallowEqual);
+
+  const [showAlert, setShowAlert] = useState(true);
+  const [isTouchDevice, setIsTouchDevice] = useState(false);
+
+  const dismissAlert = () => setShowAlert(false);
+
+  const onButtonClick = (evt) => {
+    const touchDevice = evt.type === 'touchend';
```
Tomcariello, line 38: I would prefer to see these lines reordered as below. I know this functionally is the same but I find it easier to parse code if we handle one concept at a time. By moving line 38 down 2 lines we address button behavior first & then move on to device type instead of having them interwoven. 

```
const onButtonClick = (evt) => {
    evt.stopPropagation();
    evt.preventDefault();
    const touchDevice = evt.type === 'touchend';
    MEASURE_MENU_PROPS.touchDevice = isTouchDevice;
```

Tomcariello (approved):
You can take or leave my suggestions as you see fit. I think this looks good. Nice job.

victorsh:
Thanks for the feedback guys. I've updated the PR with the recommendation that were made.

PR #4185, opened 2023-03-29
WV-2649 Modernize Measure Menu

victorsh:
## Description

The MeasureMenu component is currently written as a React Class component. Today's best practices for React suggest we should be using functional components whenever possible. Using the modern best practices for React, convert the MeasureMenu component into a functional component.

Study the current behavior of the component so you know exactly what needs to be replicated.
Research [the differences between class & functional components and how to replace lifecycle hooks with useEffect hooks](https://www.scaler.com/topics/react/react-functional-vs-class-components/).
Use the [React docs](https://react.dev/reference/react) to find the best practices for updating state, using hooks & props.
Use other components from the WV repo that are currently written as functional components for reference.
If there are obvious opportunities to refactor code to be more concise or easier to read do so.
After refactoring into a functional component, test the component to make sure the behavior is still the same.

Fixes # WV-2649

## Changes
- Converted component from class based to functional
- Use of Redux hooks and shallowEquals

## How To Test
- `git fetch --all`
- `git checkout wv-2649-modernize-measure-menu`
- `npm ci`
- `go to http://localhost:3000`
- On the bottom right corner of the app, click on the measure button
- Select measure distance. Click on a location on the map, then drag the mouse to a different location and double click to complete.
- The measure menu should update with an option to `Remove Measurements` and to `Download as GeoJSON`
- Clicking on `Remove Measurements` should delete the measured distance drawn on the map.
- Clicking on `Download as GeoJSON` should start a download
- The same behavior should occur when clicking `Measure Menu`

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/measure-tool/measure-menu.js
+++ b/web/js/components/measure-tool/measure-menu.js
@@ -1,4 +1,4 @@
-import React, { memo } from 'react';
+import React from 'react';
 import { useSelector, useDispatch, shallowEqual } from 'react-redux';
 import { Form } from 'reactstrap';
 
--- a/web/js/components/measure-tool/measure-menu.js
+++ b/web/js/components/measure-tool/measure-menu.js
@@ -48,7 +48,7 @@ const MeasureMenu = memo(() => {
   const dispatch = useDispatch();
 
   const {
-    isMobile, isTouchDevice, unitOfMeasure, measurementsInProj
+    isMobile, isTouchDevice, unitOfMeasure, measurementsInProj,
   } = useSelector((state) => ({
     isMobile: state.screenSize.isMobileDevice,
     isTouchDevice: state.modal.customProps.touchDevice,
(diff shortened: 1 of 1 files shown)


web/js/components/measure-tool/measure-menu.js
```diff
@@ -68,11 +73,6 @@ const MeasureMenu = function () {
     dispatch(changeUnits(units));
```
Tomcariello, line 73: Can we write out this dispatch call to it's own function as we did for onToggle()?

```
const changeUnits = (units) => { dispatch(changeUnitsAction(units)); };
```
And then just
```
changeUnits(units);
```

victorsh: Good catch. I just updated it to wrap a function around the dispatch.

Tomcariello (approved):

PR #4191, opened 2023-03-30
Updating Dependencies v4.3.0

ryanweiler92:
4/10/2022 UPDATE:

I fixed several more bugs that I found as a result of updating a ton of dependencies here. I also updated dependencies from the 4/7 dependabot suggestions. 
- I removed the portal-react library which was replaceable with a built in feature of React v17. 
- Updated the visibility icons to more closely match the ones we were already using
- All of the issues below should be resolved now as well.

I have not been able to find any more bugs since implementing these fixes. I would like to get this into develop after the release on Tuesday so we have a while to test it while working on other things. 

Testing:
I would just give this a brief look over by running through some basic functionality. I would not expect to see any breaking bugs, possibly just a UI issue here or there but again, I have not been able to find anything else. 

 ---------------------------------------------------------------------------------------------------

This updates most of our dependencies and includes the changes required to accommodate the updates.

Known issues that still need to be fixed:

- [x] Clicking outside of modals don't close the modal (layer options modal, global settings modal)
- [x] Filter events modal 'apply' & 'cancel' buttons render outside of the modal on the top left of the screen.

- [x] Clicking on events marker causes tooltip to stay on screen. 
- After clicking on the events tab in the sidebar and then clicking on an on map event marker, a zoom animation is played which will cause the tooltip that displays the event marker name to be stuck on screen. This is likely happening in the `map/natural-events/event-markers.js` file. 

- [x] There is an issue with converting palette values that produces an error message `Cannot read properties of undefined (reading 'toFixed')`
- This error is occurring in `modules/settings/util.js`. This error can be reproduced by adding the sea surface temperature layer. It seems like when the `@khanisak/temperature-converter` package was updated, there was some refactoring required to pull in the `Celcius, Fahrenheit & Kelvin` converting functions respectively. This is causing an issue in the `convertPaletteValue` function which looks to stem from the `unitConvert` function.  

- [x] Miles/Kilometers switch in the measure menu is broken


ryanweiler92:
@victorsh 
The functionality of the custom tooltip for the event icons is looking much better, but there are some things we still need to clean up in regards to this component. 

1. Remove all of the commented code. If we need to go back and implement this back to the old way we can just go back into the commit history. 
2. We need to find a way to simplify this function. 
2-A. There is far too much code inside the useEffect hook on line 81. 
```javascript
  useEffect(() => {
    const divTooltip = document.createElement('div');
    divTooltip.setAttribute('id', tooltipId);
    divTooltip.innerHTML = title || category;
    divTooltip.style.setProperty('max-width', '160px');
    divTooltip.style.setProperty('background', 'black');
    divTooltip.style.setProperty('color', 'white');
    divTooltip.style.setProperty('position', 'fixed');
    divTooltip.style.setProperty('z-index', '1000');
    divTooltip.style.setProperty('border-radius', '5px');
    divTooltip.style.setProperty('text-align', 'center');
    divTooltip.style.setProperty('padding', '5px');
    divTooltip.style.setProperty('visibility', 'hidden');

    const divTooltipArrow = document.createElement('div');
    divTooltipArrow.setAttribute('id', tooltipArrowId);
    divTooltipArrow.style.setProperty('background', '#000');
    divTooltipArrow.style.setProperty('border-radius', '2px');
    divTooltipArrow.style.setProperty('z-index', '1000');
    divTooltipArrow.style.setProperty('width', '20px');
    divTooltipArrow.style.setProperty('height', '20px');
    divTooltipArrow.style.setProperty('rotate', '45deg');
    divTooltipArrow.style.setProperty('position', 'fixed');
    divTooltipArrow.style.setProperty('visibility', 'hidden');

    let markerContainer = document.getElementById('marker-container');
    if (markerContainer === null) {
      markerContainer = document.createElement('div');
      markerContainer.setAttribute('id', 'marker-container');
      document.body.appendChild(markerContainer);
    }
    markerContainer.appendChild(divTooltipArrow);
    markerContainer.appendChild(divTooltip);

    return () => {
      divTooltip.remove();
      divTooltipArrow.remove();
    };
  });
  ```
I think that we can move most of this to another function. You really want React hooks to be calling functions that will execute code instead of just putting logic inside hooks. 

2-B: I think if it is possible that we would like to see the custom tooltip be imported from another file entirely. In the components directory we have a subdirectory called `util` which houses some of our custom elements like `checkbox` and `button`. I think that a lot of this code regarding the tooltip itself can be lifted from this component and placed in a file in this subdirectory. 

2-C: We should find a way to not repeat ourselves as much here.
```javascript
    const divTooltipArrow = document.createElement('div');
    divTooltipArrow.setAttribute('id', tooltipArrowId);
    divTooltipArrow.style.setProperty('background', '#000');
    divTooltipArrow.style.setProperty('border-radius', '2px');
    divTooltipArrow.style.setProperty('z-index', '1000');
    divTooltipArrow.style.setProperty('width', '20px');
    divTooltipArrow.style.setProperty('height', '20px');
    divTooltipArrow.style.setProperty('rotate', '45deg');
    divTooltipArrow.style.setProperty('position', 'fixed');
    divTooltipArrow.style.setProperty('visibility', 'hidden');
```
If you have to you can probably create an array of css properties ex `[{'background', '#000'}, {'border-radius', '2px'}]` and then map through these. We want to avoid repeating ourselves whenever possible. 

2-D: Any other ways you find that might be able to cut down on the verbosity of this component would be great too (ex. moving styles to css files) but I understand that the way we are creating the element here might make that sort of difficult. 

3. Updating the end-to-end `event-filter-test`. You can run this test in chrome using the command `npx playwright test event-filter-test.spec.js --headed`. If you get a message about binaries not being installed you should be able to run `npx playwright install` to install those. All of this can be found in the `e2e_testing.md` file in the repository. You should be able to trace back to the exact line that is failing from the error message in the end-to-end test. It will likely just require updating some class names that are being used in the test.

Tomcariello:
This looks really good. I only noticed 2 things when comparing the desktop versions:

- On initial load with "Welcome to Worldview", the background no longer has reduced brightness.
- I know a lot of icons have been changed, but the "copy" icon on a place marker might be worth checking for a closer match.

I'll check mobile this afternoon.

Tomcariello:
I don't know how crazy we want to get, but in the layer picker on mobile, if you click "Hazards And Disasters" to bring up the menu it's a few pixels higher than in production. Prod is on the right in the pic.

image

Tomcariello:
Last one...in the LAYER COVERAGE panel the question mark in circle icon is blue when it used to be white. Again, production on the right side.
image

PR #4207, opened 2023-04-03
WV-2662 modernize mobiledatepicker

victorsh:
## Description

The MobileDatePicker component is currently written as a React Class component. Today's best practices for React suggest we should be using functional components whenever possible. Using the modern best practices for React, convert the MobileDatePicker component into a functional component.

1. Study the current behavior of the component so you know exactly what needs to be replicated.
2. Research the differences between class & functional components and how to replace lifecycle hooks with useEffect hooks.
3. Use the React docs to find the best practices for updating state, using hooks & props.
4. Use other components from the WV repo that are currently written as functional components for reference.
5. If there are obvious opportunities to refactor code to be more concise or easier to read do so.
6. After refactoring into a functional component, test the component to make sure the behavior is still the same.

Fixes # wv-2662

## How To Test

1. `git fetch --all`
2. `git checkout wv-2662-modernize-mobiledatepicker`
3. `npm ci`
4. `npm run watch`
5. Open Chrome and switch to mobile view
6. Click on the date on the bottom left of the screen
7. Switch the date to something other than the current date is see if everything updates

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/timeline/mobile-date-picker.js
+++ b/web/js/components/timeline/mobile-date-picker.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import DatePicker from 'react-mobile-datepicker';
 import { getDisplayDate, getISODateFormatted } from './date-util';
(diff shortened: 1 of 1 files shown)


ryanweiler92 (changes requested):
This looks pretty good both functionally and syntactically, there are just a few things I would like to see changed.

1. Lets change the component declaration to:
`function MobileDatePicker = (props)`
I am getting a soft warning on the declaration here and we declare almost all of our other functional components this way. 

2. Lets destructure the props separately like this:
```javascript
function MobileDatePicker = (props) { 
const { 
date,
startDateLimit,
endDateLimit,
onDateChange,
hasSubdailyLayers,
} = props;
```
I think that this is a little easier to read. 

3. This one is optional but personally I like to keep the JSX as clean as possible so I would handle the logic for displayDate and headerTime outside of the jsx like this: 
``` 
const displayDate = getDisplayDate(date, hasSubdailyLayers) 
const headerTime = getHeaderTime(time, hasSubdailyLayers) 
.... 
<div className="mobile-date-picker-select-btn-text">
 <span>{displayDate}</span> 
</div> 
... 
<DatePicker 
.... 
customHeader={headerTime}
 /> 
``` 
If you or Tom feel like the way you wrote it as it is now is easier to read/understand then I would be okay with that too. 

Other than that I think this looks good. I really like how you were able to replace the `componentDidUpdate` hook with the useEffect hook with multiple variables in the dependency array.

Tomcariello:
I agree on all points. Readability is the priority in my eyes.

ryanweiler92 (changes requested):
Can you make sure to go back and update #1 from my previous comment. I am still getting the soft syntax warning on the component declaration. We want to use the syntax 
```javascript
function MobileDatePicker (props) {
```

Tomcariello (approved):

ryanweiler92 (approved):

PR #4219, opened 2023-04-07
WV-2671 slow down tour animations

Tomcariello:
## Description
Make a global edit to slow down transition time in tour story map transitions.

Fixes #WV-2671

## Notes
The production code does a calculation using the distance from one point to the next when determining the duration of the transition. This PR alters the calculation in 2 ways:

- Doubles the length of transitions relative to the production code
- Set a minimum of 5 seconds on any transition

It is simple to tweak the values based on the opinions of the team to get this dialed in.

## How To Test
- git checkout wv-2671-slow-down-tour-animations
- npm run watch
- Load any tour story
- Proceed through the story & confirm the transitions are running slower relative to production

@nasa-gibs/worldview


ryanweiler92 (approved):
This looks good to me. I've updated the event end-to-end test because the event animations were taking longer than the test was willing to wait.

minniewong (approved):

PR #4223, opened 2023-04-10
WV-2672: Step Backwards on Tile Error

ryanweiler92:
## Description

When making tile request on load, if there are errors on the tile requests, step imagery back to date with full imagery. 
- Max Daily Step Back: `7 Days`
- Max Subdaily Step Back: `23 Hours`
- Can handle daily & subdaily tiles together

This adds:
- A `TileErrrorHandler` component in `mapUI` that handles when and how to step back through dates.
- A `kiosk` parameter that can be included in the permalink to trigger the error handling on load.
- Updates to the `tileLoadFunction` in `layerbuilder` to track and push each error tile matching the current app date to the Redux store.
- Redux actions/reducer to handle entering/exiting `kiosk` mode and tracking error tiles. 

## Testing

1. Daily Layer Only
- Use this [permalink](http://localhost:3000/?v=-282.4650237431051,-148.10845457420226,321.36078019016395,121.09721634604688&kiosk=true&l=VIIRS_SNPP_Aerosol_Type_Deep_Blue_Best_Estimate&lg=false) to test the logic for daily tiles only. This link includes no time parameter. 
- This uses only the `Deep Blue Aerosol Type` layer which is usually not fully available for the current date for a full day and a half. So using this link should move the date back 1 to 2 days. 

2. Subdaily Layer Only
- This [permalink](http://localhost:3000/?v=-515.488162993126,-274.9757744402694,195.51183700687395,206.5242255597306&z=4&ics=true&ici=5&icd=10&kiosk=true&l=GOES-West_ABI_GeoColor(hidden),GOES-East_ABI_Band2_Red_Visible_1km(hidden),Himawari_AHI_Band3_Red_Visible_1km(hidden),GOES-West_ABI_Band13_Clean_Infrared(hidden),Himawari_AHI_Band13_Clean_Infrared(hidden),GOES-West_ABI_Air_Mass(hidden),Himawari_AHI_Air_Mass(hidden),GOES-East_ABI_Air_Mass(hidden),GOES-East_ABI_Band13_Clean_Infrared(hidden),GOES-West_ABI_Band2_Red_Visible_1km(hidden),GOES-East_ABI_GeoColor(hidden)&lg=false&t=2023-04-17-T13%3A38%3A34Z) will load the most recent time (no date parameter) with all of the subdaily layers. Each layer will have it's visibility turned off. You will need to find one of these layers that has no imagery at the current time and then refresh. This is because the logic is only designed to work on load for newly added layers, and there are no subdaily layers are consistently not available at the current time. 
- Usually the second layer in the list `Red Visible (0.64 um, Band 2, 10 minute)` will work.
- Once you find a subdaily layer without imagery and refresh, the error handler should typically find the latest available imagery from 10-30 minutes earlier and adjust the date.

3. Daily & Subdaily Layers
- This [permalink](http://localhost:3000/?v=-125.47454977968671,-109.44678084846323,355.7198942945144,105.08574213461809&z=4&kiosk=true&l=VIIRS_SNPP_Aerosol_Type_Deep_Blue_Best_Estimate,Himawari_AHI_Band3_Red_Visible_1km&lg=false&t=2023-04-15-T13%3A47%3A36Z) will load the `Deep Blue Aerosol Type` daily layer and the `Red Visible (0.64 um, Band 2, 10 minute)` subdaily layer at the most recent date. 
- When there are daily & subdaily layers present, the logic is designed to sort the daily layer error tiles so that they take precedence over the subdaily error tiles, since it would not make sense to check the subdaily tiles first. 
- These layers should move back to the same date as test 1 from above, except you show see both the daily and subdaily layer visible.


web/js/map/layerbuilder.js
```diff
@@ -102,6 +109,35 @@ export default function mapLayerBuilder(config, cache, store) {
...
+  const handleTileError = (tile, layer, sourceURL) => {
+    const state = store.getState();
+    const { selected: reduxDate } = state.date;
+    const { isKioskModeActive } = state.ui;
+
+    const { id, layerPeriod } = layer;
+
+    const isSubdailyLayer = layerPeriod === 'Subdaily';
+
+    if (isKioskModeActive) {
```
Tomcariello, line 122: We can save a (tiny) bit of processing if we check for kiosk mode earlier in this function. 

```
    const state = store.getState();
    const { isKioskModeActive } = state.ui;
    if (isKioskModeActive) {
       const { selected: reduxDate } = state.date;
       const { id, layerPeriod } = layer;
       const isSubdailyLayer = layerPeriod === 'Subdaily';
```

web/js/mapUI/components/tile-errors/tile-error-handler.js
```diff
@@ -0,0 +1,171 @@
...
+// gets a week ago from real time date
+function weekAgo(realTimeDate) {
+  const inputDate = new Date(realTimeDate);
+  const oneWeekInMilliseconds = 7 * 24 * 60 * 60 * 1000;
+  const earlierDate = new Date(inputDate.getTime() - oneWeekInMilliseconds);
+  const earlierDateString = earlierDate.toString();
+  return earlierDateString;
+}
+
+// safe gaurd agaisnt stepping back too far, we only want to go back 7 days for daily and 24 hours for subdaily
```
Tomcariello, line 58: I like your comments overall but I would remove the reference to subdaily here as it is handled directly below & contradicts between 23/24 hours.

web/js/mapUI/components/tile-errors/tile-error-handler.js
```diff
@@ -0,0 +1,171 @@
...
+  const hourlySafeguardCheck = compareSubdailyDates(lastDateToCheck, selectedDate);
+
+  useEffect(() => {
+    if (isKioskModeActive && errorTiles.length && dailySafeguardCheck && hourlySafeguardCheck && !isLoading) {
+      sortErrorTiles();
+    }
+  }, [action]);
+
+  // sorting tiles for daily & subdaily
+  function sortErrorTiles() {
```
Tomcariello, line 114: Are there perf concerns here? Would we be better off storing separate daily/subdaily in state & avoid sorting all error tiles each time?

Tomcariello (commented):
This looks really good & works as described.

PR #4227, opened 2023-04-12
wv-2670: Modify distraction free mode to display distance scale & minimal timestamp

Tomcariello:
## Description
Add "minimally distracting timestamp" and scale bar to UI while in distraction free mode.

Fixes # WV-2670

## How To Test
- git checkout wv-2670-kiosk-distraction-mode
- npm run watch
- Reset Worldview to default layers
- Enter "Distraction Free" mode via Info button on the top right
- Confirm that scale is displayed in desktop view
- Confirm date (Year Month Day) is displayed in desktop & mobile views
- Exit Distraction free mode
- Add a geostationary (or otherwise subdaily) layer
- Enter "Distraction Free" mode via Info button on the top right
- Confirm date & time (Year Month Day Hours Minutes) is displayed in desktop & mobile views

@nasa-gibs/worldview


minniewong:
Looks pretty nice but I think we could possibly remove some of that padding around the date. Also, if we have a geostationary layer loaded, I think we should have the time included as well.

ryanweiler92 (changes requested):
I would probably agree with Minnie that the date element's inner spacing could be reduced. You will probably want to change this section `<section id="timeline" className="timeline-inner clearfix">` since we are using the `timeline` id above already. After you create a unique ID I would remove the `timeline-inner` class since this is what is affecting the spacing inside the element and just move your styles to your new ID.

PR #4283, opened 2023-04-28
wv-2706: restore collapsed layer list styles

Tomcariello:
## Description
When the layer list is collapsed the font is black instead of white, making it hard to read.
This was caused by the recent bootstrap update. You can read more here: https://github.com/twbs/bootstrap/issues/24140

Fixes # wv-2706

## How To Test
- git checkout wv-2706-layer-list-font-color
- npm run watch
- Load worldview & click the up arrow/triangle to the right of the Data tab to collapse the layers
- Confirm that the styles are correct (white font on dark gray background, black font on gray background when hovered)

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/sidebar/collapsed-button.js
+++ b/web/js/components/sidebar/collapsed-button.js
@@ -23,6 +23,7 @@ class CollapsedButton extends PureComponent {
           id={buttonId}
           aria-label={labelText}
           onClick={onclick}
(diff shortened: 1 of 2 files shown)


ryanweiler92 (approved):

minniewong (approved):
Looks great!

PR #4285, opened 2023-04-28
wv-2704 layer info url underlining

Tomcariello:
## Description
Hyperlinks are underlined in the layer descriptions by default, when they should not be. 

NOTE: This is because the latest bootstrap has a generic `a { text-decoration: underline}` property. This same problem may appear elsewhere so we should keep an eye out.

Fixes # wv-2704

## How To Test
- git checkout wv-2704-url-underline
- npm run watch
- Click on the "info" icon (view description) for any layer
- Confirm the hyperlinks are not underlined by default & are underlined when hovered.

@nasa-gibs/worldview

### DIFF
--- a/web/scss/components/modal.scss
+++ b/web/scss/components/modal.scss
@@ -16,6 +16,11 @@
 
   a {
     color: #7bf;
+    text-decoration: none;
+  }
+
+  a:hover {
+    text-decoration: underline;
   }
 
   .modal-header {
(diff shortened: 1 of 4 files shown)


ryanweiler92 (approved):

minniewong:
Screen Shot 2023-04-28 at 9 45 20 AM
In the Data tab, I still see a couple of urls underlined

Tomcariello:
These URLs are fixed as well. Thanks.

minniewong (approved):

PR #4294, opened 2023-05-03
wv-2710: CSS adjustments to fix compare mode hover palette

Tomcariello:
## Description
The text for the active tab in compare mode should be a contrasting color when the cursor hovers over it. Currently it's white on white when you hover over the text with the cursor. 

Fixes # wv-2710

## How To Test

- git checkout wv-2710-hover-color
- npm run watch
- Enter Compare Mode
- Hover over the active tab and notice (in production) that it is white on white
- Hover over the active tab and notice (in this branch) that it is white on a contrasting gray. The icon also swaps colors & the bottom border is more cohesive with these changes.

@nasa-gibs/worldview

### DIFF
--- a/web/scss/features/compare.scss
+++ b/web/scss/features/compare.scss
@@ -165,6 +165,25 @@
   span {
     padding-bottom: 2px;
   }
+
+  &.nav-link {
+    &.first-tab, &.second-tab {
+      background-color: #28292a;
+    }
+
+    &.active {
+      background-color: #fff;
+
+      // Toggle layer icon color on hover by repositioning the SVG
+      &:hover > i {
+        background-position: 0 -10px;
+      }
+    }
+
+    &:hover {
+      background-color: #404542;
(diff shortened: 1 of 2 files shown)


ryanweiler92 (approved):

minniewong:
Screen Shot 2023-05-04 at 11 06 14 AM
In mobile, there is a long grey bar to the right of the tabs, which doesn't seem like it should be there. Also, when you select a tab, it seems slightly confusing as to know which tab is active as the tab is only outlined, and not a different color. Also, if you click on a tab twice, it seems to first select the new tab with a white outline, but upon second click, the first tab turns white and becomes the active tab. Not sure if this is just a problem with using dev tools for testing mobile.

Tomcariello:
@minniewong these issues have been addressed in the latest update. Thanks.

minniewong (approved):
⚡

PR #4310, opened 2023-05-08
Set static loopdelay

Tomcariello:
## Description
From Slack: 

Set the loop delay to a fixed value when FPS >= 1 . A fixed value of 1500 seemed to work fine, i.e.

const loopDelay = speed === 0.5 ? 2000 : 1500

Let’s just go ahead and make it a static amount.

## How To Test

- git checkout Set_static_loopdelay
- npm run watch
- Start an animation. Notice that with any FPS above 0.5 the amount of time between the last frame of the animation & the start of the next loop remains a constant 1.5 seconds.

@nasa-gibs/worldview


minniewong (approved):
Looks good

PR #4340, opened 2023-05-12
WV-2708: Animation Date Range Check

ryanweiler92:
## Description

This provides the ability to test an entire date range of an animation for missing or blank tiles. With this new feature we will be able to scan the latest date range to be used in an animation in kiosk mode and determine whether the animation should play or abort the animation and remain on the current date with static imagery. This works very well since it waits for the initial error tile check process, so the current imagery should always be something we've already considered satisfactory.

Currently this will scan the entire date range of daily or subdaily animations in kiosk mode and it will play the animation only if less than 40% of the dates in the range have missing or blank tiles. This value can be changed to tighten or loosen the constraints on whether to play the animation. In my opinion the current 40% value is very generous/loose and it will play the animation in most situations. 

There are currently 2 console statements in `date-range-tile-check.js` that I would like to keep and push into SIT and possibly UAT before we push to production. Since this will not be released into the first iteration of kiosk mode, the console statements will provide valuable context to what we are seeing when we test in SIT & UAT. 

The modularity handicap for this feature is that any layers that need to be checked in an animation will need to manually be added and measured to `date-range-tile-check.js`. I have provided a detailed process for doing this in comments of this file.

## How To Test

1. `git checkout wv-2708-animation-range-check`
2. `npm ci`
3. `npm run watch`
4. Open the browser console
5.  Use this [URL](http://localhost:3000/?v=-223.38671531740192,-111.23416482365595,252.6798741684782,101.75604370153731&z=4&ics=true&ici=5&icd=10&kiosk=true&df=true&eic=sa&l=Reference_Features_15m,GOES-West_ABI_GeoColor,GOES-East_ABI_GeoColor,VIIRS_SNPP_CorrectedReflectance_TrueColor(opacity=0.61)&lg=true) to open the original subdaily kiosk scenario we were considering.
6. In the console you will see either a success or failure message depending on the amount of missing tiles which will also provide a detailed message about how many dates were found with missing tiles in the animation range. 
7. If the message is success, verify that the animation plays. If the message is failure, verify that the animation is aborted and that static imagery is displayed. 
8. If you would like to force a scenario where the animation is aborted, open up `date-range-tile-check.js` and update line 235 to `if (percentMissing < 1) {`
9. Repeat the process for the [daily animation scenario](http://localhost:3000/?v=-9215416.788865805,-4212995.281243633,9489665.699466601,4155580.686192584&p=arctic&df=true&kiosk=true&eic=da&l=Land_Mask,AMSRU2_Sea_Ice_Concentration_12km(palette=blue_6)&lg=true)


web/js/modules/ui/actions.js
```diff
@@ -74,16 +80,30 @@ export function clearErrorTiles() {
...
+// Displays a static image map stored locally in repo.
 export function toggleStaticMap(isActive) {
   return {
     type: DISPLAY_STATIC_MAP,
     isActive,
   };
 }
 
+// When kiosk mode is active && (eic === 'sa' || eic === 'da') this action is dispatched once we step the date back to full imagery.
+// This signals that the animation availability to ready to be checked.
```
Tomcariello, line 93: Can you clarify line 93? It's not clear to me what this means...

Tomcariello (approved):
This works as expected & looks good. I just had the 1 minor comment (about a comment). Nice work.

minniewong (approved):

PR #4364, opened 2023-05-25
WV 2679 layer picker menu header bug

PatchesMaps:
## Description

This PR fixes behavior in the "Add Layer" workflow where the header scrolls off-screen. The header should remain at the top of the screen while scrolling through the available layers.

## How To Test

1. `git checkout WV-2679-layer-picker-menu-header-bug`
2. `npm ci`
3. `npm run watch`
4. Got to `http://localhost:3000`
5. Use dev tools to emulate a mobile device.
5. Click on `+Add Layers`
6. Scroll through the available layers.
7. Confirm that the category header remains at the top of the screen.


Tomcariello (approved):
This looks good. Simple & solves the problem.

I really want to fix the spelling of "outter" here...

ryanweiler92 (approved):
This looks great! Pulling in develop should fix the e2e tests now.

minniewong (commented):
I noticed this when in desktop mode in Chrome - I'm not sure if this is the same thing Tom was talking about. It should be 3 across


Screenshot 2023-05-31 at 10 55 10 AM

minniewong (approved):
Looks great!

PR #4380, opened 2023-05-30
WV-2726 "Running data value" feature breaks under certain circumstances

PatchesMaps:
## Description

Fixes bug where the "Running data value" feature breaks under certain circumstances

## How To Test

1. Select the Events tab
2. Select a cyclone event
3. Select the Layers tab
4. Expand the "Precipitation Rate" layer group
5. Hover over the map where there is precipitation rate imagery; the running data value indicators should appear in the legend

All other functionality should remain the same.

## Merging

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/sidebar/paletteLegend.js
+++ b/web/js/components/sidebar/paletteLegend.js
@@ -146,7 +146,7 @@ class PaletteLegend extends React.Component {
         if (this[ctxStr]) {
           const newWidth = this[ctxStr].current.getBoundingClientRect().width;
           // eslint-disable-next-line react/destructuring-assignment
-          if (newWidth !== this.state.width) {
+          if (newWidth && newWidth !== this.state.width) {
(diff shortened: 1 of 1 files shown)


Tomcariello (commented):
This fix makes sense however I am unable to reproduce the bug & therefore unable to confirm the fix.

 Is there a reproducible method to see the problem?

PatchesMaps:
The steps above should be able to replicate the bug. However, it is finicky, doing anything like toggling the visibility of the layer will cause the bug not to occur.

Tomcariello:
> The steps above should be able to replicate the bug. However, it is finicky, doing anything like toggling the visibility of the layer will cause the bug not to occur.

Ahh, I see. I needed to have "Group Similar Layers" checked before selecting the cyclone event. Thanks.

Tomcariello (approved):

ryanweiler92 (approved):

PR #4409, opened 2023-06-08
WV-2725: Fix story crashes

ryanweiler92:
## Description

When stepping through a story we allow users to enter & exit the events tab. When incrementing or decrementing into a step of a story that uses compare mode, the app would crash if the events tab was still open. This updates the tour logic to check each incremental or decremental step to see if the events tab is active, and if so, switch back to the layers tab. 

## How To Test

1. `git checkout `wv-2725-story-compare-mode`
2. `npm ci`
3. `npm run watch`
4. Start the flood tour story
5. Increment to step 2.
6. Activate the events tab
7. Increment to step 3.
8. Verify that A. The app has not crashed. B. The events tab is no longer active C. Compare mode is working. 
9. Increment to step 4.
10. Activate the events tab.
11. Decrement to step 3.
12. Verify that A. The app has not crashed. B. The events tab is no longer active C. Compare mode is working.

### DIFF
--- a/web/js/containers/tour.js
+++ b/web/js/containers/tour.js
@@ -35,7 +35,7 @@ import {
   startTour as startTourAction,
 } from '../modules/tour/actions';
 import { resetProductPickerState as resetProductPickerStateAction } from '../modules/product-picker/actions';
-
+import { changeTab as changeTabAction } from '../modules/sidebar/actions';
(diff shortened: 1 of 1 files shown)


minniewong (approved):
Looks great!

Tomcariello (approved):

PR #4446, opened 2023-06-22
WV-2638 Dynamic Data Visualization 

ryanweiler92:
## Description

NOTE: This is not a final draft, I would just like to get a few preliminary reviews on the functionality since this is such a complex project. Eventually within one of the upcoming dev meetings I will present this to get some idea's on any final changes we need to make.

This is the implementation of the UI and further functionality for the dynamic data visualization layers that we are utilizing through the FIRMS API. This adds some layers with preset RGB band combinations and also a landsat customizable layer. This customizable layer allows for users to select custom RGB combinations. 

This imagery follows orbit tracks that can be added through the associated layers settings. In the instructions below I've added a URL with a good starting point. But if you want to find imagery on other dates, simply locate the orbit tracks on the date of interest (United States Only). 

## How To Test

1. `git checkout wv-2638`
2. `npm ci`
3. `npm run build`
4. `npm run watch`
5. For simplicity, use this [URL](http://localhost:3000/?v=-77.27333274057173,38.77605197284668,-76.83722700850396,38.97116386026658&l=Reference_Labels_15m,Reference_Features_15m,MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2023-04-17-T12%3A21%3A36Z) as a starting point. This will place you over Washington DC on April 17th where any of the new DDV layers will be visible. 
6. Go into the layer picker and add and remove each preset DDV layer (HLS Natural Color, HLS Color Infrared, HLS False Color Vegetation, HLS False Color Urban, HLS Shortwave Infrared). For each layer, wait for the tiles to load and add the associated orbit track in the layer options. 
7. Add the 'HLS Customizable Landsat' layer. 
8. Open the layer options and verify that the current band selection defaults to RED: B07 GREEN: B05 BLUE: B04
9. Click on the customize button within the layer settings and confirm the new customization menu opens. 
10. Create a few custom combinations by changing the values of R, G & B in the bottom selection and then hitting the 'Confirm Band Selection' button. NOTE: Some combinations do not exist! Especially when using the B8A band and higher. We will have to figure out a way to tell what bands can actually be used in any combination. This was _not_ a problem before adding bands above B08.
11. This should close the customization menu and you should shortly be able to visualize your changes on the map. 
12. After testing a few custom combinations, try selecting and confirming the preset values from the customizable menu. 
13. Verify that the preset preview images look similar to the visualization you see on the map.


Tomcariello:
This is looking great & working as described.

PatchesMaps (approved):

PR #4476, opened 2023-07-05
WV-2724 error during animations

PatchesMaps:
## Description

Fixes: Safari error gets thrown during fast-playing animations

>When playing EIC [scenario #9](https://worldview.earthdata.nasa.gov/?v=-4171187.01916682,-4238559.180936186,4142545.3786389525,4188800.5389352706&p=antarctic&df=true&kiosk=true&eic=da&l=Land_Mask,AMSRU2_Sea_Ice_Concentration_12km(palette=blue_6)&lg=true) (animation of Antarctic sea ice concentration) in Safari, I’m often getting the following error after a few loops:
SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds

>{{}}

>… which seems to be due to stricter limits in Safari for updating the URL bar (and browser history) from what I can tell after a [quick search](https://stackoverflow.com/questions/38419867/os-x-ios-safari-history-replacestate-limit-throws-securityerror-dom-exception). Fortunately the kiosk display software seems to use Chromium which has looser limits, so we shouldn’t run into the problem there.  But we should still address this for regular ole playback of fast animations in Safari.

## How To Test

1. `git checkout WV-2724-error-during-animations`
2. `npm ci`
3. `npm run watch`
4. Got to `http://localhost:3000` (in Safari if you want to check for the original bug)
5. Create a new animation or start a tour that has one.
6. Select a high frame rate.
7. Click the play button.
8. The currently selected date or 't' parameter should not be present during the animation but should update as soon as the animation is paused. If in Safari, the error should not be logged.

## PR Submission Checklist

## Merging

@nasa-gibs/worldview


ryanweiler92 (approved):
Awesome! I'm glad you got this to work with minimal changes

Tomcariello (approved):

PR #4501, opened 2023-07-17
WV-2764: Modernizing custom-interval-selector component

christof-wittreich:
## Description
This converts the custom-interval-selector component from a React Class component to a function component, per best modern practices for React. This process included replacing the lifecycle hooks with `useEffect`, and replacing Redux's `connect` with `useSelector` and `useDispatch`.

## How To Test
1. `git checkout wv-2764-modernize-customintervalselector-component`
2. `npm ci`
3. `npm run watch`
4. Locate the custom-interval-selector component in the bottom-left of the screen, as part of the Timeline component.
5. Verify that the new custom-interval-selector component functions the same as the existing custom-interval-selector component.

### DIFF
--- a/web/js/components/timeline/custom-interval-selector/custom-interval-selector.js
+++ b/web/js/components/timeline/custom-interval-selector/custom-interval-selector.js
@@ -2,7 +2,6 @@ import React, { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import usePrevious from '../../../util/customHooks';
 import DeltaInput from './delta-input';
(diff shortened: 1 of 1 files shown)


PatchesMaps (approved):

Tomcariello (approved):

ryanweiler92 (changes requested):
Just for the sake of conciseness we should try to declare our dispatch functions after we declare `const dispatch = useDispatch()`. 

For example, instead of calling `dispatch()` in different functions, we can just write single functions to dispatch our actions like these.

```
  const changeCustomInterval = (delta, timeScale) => dispatch(changeCustomIntervalAction(delta, timeScale));
  const closeModal = () => dispatch(toggleCustomModal(false, undefined));
```

ryanweiler92 (approved):

PR #4505, opened 2023-07-20
WV-2766: Modernizing Rotation Component

christof-wittreich:
## Description
This converts the rotation component from a React Class component to a function component, per best modern practices for React. This process included replacing the class state with `useState`, and replacing Redux's `connect` with `useSelector` and `useDispatch`.

## How To Test
1. `git checkout wv-2766-modernize-rotation-component`
2. `npm ci`
3. `npm run watch`
4. Switch the projection mode to polar projection (either Arctic or Antarctic), then locate the rotation component on the right of the screen.
5. Verify that the new rotation component functions the same as the existing rotation component.

### DIFF
--- a/web/js/components/map/rotation.js
+++ b/web/js/components/map/rotation.js
@@ -4,13 +4,14 @@ import { useSelector, useDispatch, shallowEqual } from 'react-redux';
 import { debounce } from 'lodash';
 import { saveRotation } from '../../map/util';
 import HoverTooltip from '../util/hover-tooltip';
+import { refreshRotation } from '../../modules/map/actions';
 
 const duration = 500;
 
 function Rotation() {
   const [intervalId, setIntervalId] = useState(null);
 
-  const map = useSelector((state) => state.map, shallowEqual);
+  const map = useSelector((state) => state.map.ui.selected, shallowEqual);
   const proj = useSelector((state) => state.proj, shallowEqual);
   const rotation = useSelector((state) => state.map.rotation);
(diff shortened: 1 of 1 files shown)


Tomcariello (approved):

ryanweiler92 (changes requested):
This looks really good and works well, there are just a few minor changes I think we should implement before merging. 

1. I think we can just dispatch one of our current actions to update the rotation with redux. It looks like in the `/modules/map` directory there is a `refreshRotation` action that does exactly what you need here. So instead of this 

```
  const updateRotationState = (radians) => debounce(() => {
    dispatch({
      type: 'MAP/UPDATE_ROTATION',
      rotation: radians,
    });
  }, 100);
  ```
  You can just import that action and use it something like this. Just remember to add the debounce.
  
  const updateRotation = (radians) => { dispatch(refreshRotation(radians)); };
  
  2. In the map object we are grabbing through redux, it looks like we can drill down to a lower level in the object in the useSelector hook. I.E. 
  `const map = useSelector((state) => state.map.ui.selected, shallowEqual);`
  You will have to go update the instances where map is being used in the component so for example, this 
  
  ```
    map.ui.selected.getView().animate({
      duration: 500,
      rotation: 0,
    });
```

Would just become this 

```
    map.getView().animate({
      duration: 500,
      rotation: 0,
    });
```

This way we are comparing a smaller object on rerenders.

ryanweiler92 (approved):
Thanks for making those changes. You can go ahead and merge this whenever, I ran all of the e2e tests locally and they passed so it should be fine.

PR #4519, opened 2023-07-25
Wv 2611 validate options error handling

PatchesMaps:
## Description

>The recent outage of SMAP products exposed some issues with handling errors at build time. These errors include:
 >* ~Synchronous fetching of GC colormaps, vector styles and vector data will all break. Change to async.~
 >* validateOptions is no longer validating if a layer has a projection. This means that if a layer-metadata file is present but not the GC info, it will not remove the layer, WV will continue to build but it will break when attempting to add a layer in WV. The fix is to add back in the projection validation.
 >* ~We need to test other validateOptions by excluding the layer from the pulled GC and/or excluding only the layer-metadata but including the GC info.~

>These build errors might cause issues with the GITC WV build if not addressed and next time a layer is missing from GC.


minniewong (approved):
Build works and I can load Worldview

ryanweiler92 (approved):

PR #4520, opened 2023-07-26
WV-2772: Modernizing date-selector Component

christof-wittreich:
## Description
This converts the date-selector component from a React Class component to a function component, per best modern practices for React. This process included replacing the lifecycle hooks and class state with `useEffect` and `useState`, respectively.

## How To Test
1. `git checkout wv-2772-modernize-dateselector-component`
2. `npm ci`
3. `npm run watch`
4. Locate the date-selector component, which is used as part of the Timeline component in the lower-left.
5. Verify that the new date-selector component functions the same as the existing date-selector component.

### DIFF
--- a/web/js/components/date-selector/date-selector.js
+++ b/web/js/components/date-selector/date-selector.js
@@ -1,9 +1,10 @@
 /* eslint-disable react/jsx-props-no-spreading */
-import React, { Component } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import DateInputColumn from './date-input-column';
 import util from '../../util/util';
 import { MONTH_STRING_ARRAY } from '../../modules/date/constants';
+import usePrevious from '../../util/customHooks';
 
 /*
  * DateSelector used within Timeline and AnimationWidget.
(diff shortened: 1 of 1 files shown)


ryanweiler92 (approved):

web/js/components/date-selector/date-selector.js
```diff
@@ -206,51 +207,47 @@ class DateSelector extends Component {
...
 
     // check if date is within min/max range
-    const dateTime = date.getTime();
+    const dateTime = newDate.getTime();
     const minDateTime = minDate.getTime();
     const maxDateTime = maxDate.getTime();
     const dateWithinRange = dateTime >= minDateTime && dateTime <= maxDateTime;
 
     // updateDate at this stage can still be invalid with pending timeunit changes
     // eslint-disable-next-line react/destructuring-assignment
```
Tomcariello, line 233: I don't think this override is actually doing anything here. Can we remove it?

web/js/components/date-selector/date-selector.js
```diff
@@ -259,19 +256,18 @@ class DateSelector extends Component {
   /**
   * @desc update date with newDate if valid from check and then reset temp time values
   *
-  * @param {String} date
+  * @param {String} dateIn
   * @param {Boolean} isRollDate
   * @returns {void}
   */
   // eslint-disable-next-line react/destructuring-assignment
```
Tomcariello, line 263: I don't think this override is actually doing anything here. Can we remove it?

web/js/components/date-selector/date-selector.js
```diff
@@ -1,61 +1,68 @@
 /* eslint-disable react/jsx-props-no-spreading */
```
Tomcariello, line 1: This override is functional, but it's worth considering getting rid of it (wv-2579 intends to address overrides). It appears we just need to eliminate `{...sharedProps}` when passing props in a handful of places.

Tomcariello (approved):
This looks good. Feel free to address the overrides (or not) as you see fit.

PR #4521, opened 2023-07-26
WV 2612 verbose mode 2

PatchesMaps:
## How To Test
1. `git checkout WV-2611-verbose-mode-2`
2. `npm run build:verbose`
3. Observe log output

### DIFF
--- a/tasks/build-options/getCapabilities.js
+++ b/tasks/build-options/getCapabilities.js
@@ -108,8 +108,10 @@ async function handleException (error, link, dir, ext, count) {
 }
 
 async function processVectorData (layer) {
-  const ident = layer['ows:Identifier']._text
-  if (argv.mode === 'verbose') console.warn(`Processing vector data for ${ident}:`)
+  if (argv.mode === 'verbose') {
+    const ident = layer['ows:Identifier']._text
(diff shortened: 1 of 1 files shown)


ryanweiler92 (approved):

tasks/build-options/validateConfigs.js
```diff
@@ -47,13 +53,15 @@ async function main () {
     validateFile(filePath)
   }
   if (invalidJsonFiles.length) {
+    if (argv.mode === 'verbose') console.warn(`${prog}: Invalid JSON files: ${invalidJsonFiles}`)
```
Tomcariello, line 56: I don't think this log is required since at this point we are throwing an error already.

Tomcariello (approved):
This looks good, nice job. Feel free to take or leave my minor suggestions.

tasks/build-options/validateConfigs.js
PatchesMaps: I want verbose mode to output the bad files themselves, not just the number of bad files. I guess i could throw the error in an else statement but it's verbose mode so I figured there was no harm in outputting both.

PR #4535, opened 2023-07-31
Wv 2742 investigate bbox coordinate

PatchesMaps:
## Description

> There seems to be an issue with editing bounding box coordinates for image snapshots in Arctic/Antarctic projections - anytime you try to edit the coordinates in the "Edit coordinates" box, it will state "Not visible" even if you are able to draw the bounding box with those coordinates.

## How To Test

1. `git checkout `
2. `npm ci`
3. `npm run watch`
4. Got to [this url](http://localhost:3000/?v=-7077888,-4075520,7077888,4075520&p=antarctic&l=Coastlines_15m,BlueMarble_ShadedRelief_Bathymetry,VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor&lg=true&t=2023-07-31-T14%3A16%3A47Z)
5. Draw a bounding box (snapshot/camera icon)
6. Try to edit the corner coordinates to a coordinate that is visible on the map
7. It should not say "Not Visible" as long as the edited corner is within the visible extent.

### DIFF
--- a/web/js/components/image-download/lat-long-inputs.js
+++ b/web/js/components/image-download/lat-long-inputs.js
@@ -1,9 +1,7 @@
 import React, { useState, useEffect } from 'react';
 import * as olProj from 'ol/proj';
-import {
-  clone as lodashClone,
-} from 'lodash';
 import { containsExtent, isEmpty } from 'ol/extent';
+import { fromExtent } from 'ol/geom/Polygon';
 import PropTypes from 'prop-types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { CRS } from '../../modules/map/constants';
@@ -27,9 +25,12 @@ function Input({
   const update = () => {
     try {
       const newInputValue = Number(inputValue);
-      const newArray = lodashClone(boundingBoxArray);
+      const newArray = structuredClone(boundingBoxArray);
       newArray[index] = newInputValue;
-      const crsCorrectedExtent = olProj.transformExtent(newArray, CRS.GEOGRAPHIC, crs);
+      // Due to a bug (ol v7.4.0) in transformExtent, we need to create a polygon before transforming it.
+      const extentPolygon = fromExtent(newArray);
+      const crsCorrectedExtentPolygon = extentPolygon.transform(CRS.GEOGRAPHIC, crs);
+      const crsCorrectedExtent = crsCorrectedExtentPolygon.getExtent();
 
       if (containsExtent(viewExtent, crsCorrectedExtent)
       && isValidExtent(newArray)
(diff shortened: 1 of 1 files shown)


ryanweiler92 (approved):

web/js/components/image-download/lat-long-inputs.js
```diff
@@ -27,9 +24,12 @@ function Input({
   const update = () => {
     try {
       const newInputValue = Number(inputValue);
-      const newArray = lodashClone(boundingBoxArray);
+      const newArray = structuredClone(boundingBoxArray);
       newArray[index] = newInputValue;
-      const crsCorrectedExtent = olProj.transformExtent(newArray, CRS.GEOGRAPHIC, crs);
+      // Due to a bug (ol v7.4.0) in transformExtent, we need to create a polygon before transforming it.
```
Tomcariello, line 29: Is this comment accurate & necessary to check into the repo?

Tomcariello (approved):
This looks good & addresses the ticket. Just a minor quibble on the value of including the comment in the PR.

web/js/components/image-download/lat-long-inputs.js
PatchesMaps: I created a [issue](https://github.com/openlayers/openlayers/issues/14961) in OpenLayers for the behavior that was causing the bug. I figured that the comment would be a nice breadcrumb as to why we're not using `transformExtent` so that we don't accidentally "fix" it in the future until the bug is truly resolved.

Tomcariello: If I am parsing the response to that ticket correctly, they are saying not a bug in `transformExtent`...? 

I don't want to get too sidetracked on this, just want to make sure that our comments are not misleading.

PatchesMaps: Yeah, `transformExtent` returns `[Infinity, Infinity, -Infinity, -Infinity]` when trying to transform extents around Antarctica. From my investigation, I believe that this is an issue with OpenLayers since the behavior is inconsistent in regards to which extents succeed and which ones fail. The `Infinity` extent then trips our `!isEmpty(crsCorrectedExtent)` check on line 36 and causes the described bug.

PatchesMaps: I hadn't seen the response to my issue. I'll remove the comment accordingly.

minniewong (approved):

PR #4536, opened 2023-07-31
WV-2773: Modernizing image-download-panel Component

christof-wittreich:
## Description
This converts the image-download-panel component from a React Class component to a function component, per best modern practices for React. This process included replacing the class state with `useState`.

## How To Test
1. `git checkout wv-2773-modernize-imagedownloadpanel-component`
2. `npm ci`
3. `npm run watch`
4. Click the "Take a snapshot" button in the top-right toolbar, then locate the image-download-panel component on the upper-right side of the screen.
5. Verify that the new image-download-panel component functions the same as the existing image-download-panel component.

### DIFF
--- a/web/js/components/image-download/image-download-panel.js
+++ b/web/js/components/image-download/image-download-panel.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import googleTagManager from 'googleTagManager';
 import {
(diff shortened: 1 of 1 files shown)


web/js/components/image-download/image-download-panel.js
```diff
@@ -30,25 +30,38 @@ const RESOLUTION_KEY = {
  * @class resolutionSelection
```
Tomcariello, line 30: This comment (lines 26-32) is not accurate. It should be updated or removed.

web/js/components/image-download/image-download-panel.js
```diff
@@ -59,84 +72,70 @@ export default class ImageDownloadPanel extends React.Component {
...
       { width, height },
       time,
-      fileType,
-      fileType === 'application/vnd.google-earth.kmz' ? false : isWorldfile,
+      currFileType,
+      currFileType === 'application/vnd.google-earth.kmz' ? false : currIsWorldfile,
       markerCoordinates,
     );
 
     if (url) {
```
Tomcariello, line 80: Is this check of `url` required? If not we should eliminate it, if so we may want to keep the `else` condition.

Tomcariello (changes requested):
I am getting` Invalid prop 'isWorldfile' of type 'string' supplied to 'ImageDownloadContainer', expected 'boolean'.` when I change the 'worldfile' dropdown value.

Tomcariello (approved):

PR #4556, opened 2023-08-10
WV-2791: Layer Threshold Max Decreasing Fix

christof-wittreich:
## Description
This fix prevents the threshold max of a layer decreasing on every page reload. This was done by modifying the value put into the url for the layer's threshold max. Previously, the url contained the low-end of the range label of the max value, and now the url will instead contain the high-end of the range label of the max value. This fix not only prevents the threshold max from decreasing on page reload, but also is a more accurate representation of the threshold maximum, now being the high-end of that max range.

In addition to this fix, also included in this PR is a minor style fix to the threshold range input track, to fix a Firefox-specific style issue with the range input track's border.

## How To Test
1. `git checkout wv-2791-layerthresholdmax`
2. `npm ci`
3. `npm run watch`
4. Click the "Add Layers" button on the main sidebar on the right, and add any layer with a threshold option (for example, the Aerosol Index layer). Then, click the options sliders icon on the layer card to open the layer options window, and locate the thresholds slider.
5. Move the right side of the slider to any point other than the absolute maximum.
6. Verify that reloading the page does not change the threshold you selected in the previous step, and that it correctly persists between page reloads without changing.

### DIFF
--- a/web/js/modules/palettes/selectors.js
+++ b/web/js/modules/palettes/selectors.js
@@ -7,7 +7,7 @@ import {
   cloneDeep as lodashCloneDeep,
 } from 'lodash';
 import update from 'immutability-helper';
-import { getMinValue, getMaxValue } from './util';
+import { getMinValue } from './util';
 
 /**
  * Gets a single colormap (entries / legend combo)
@@ -287,18 +287,13 @@ const toggleLookup = function(layerId, palettesObj, state) {
   return update(newPalettes, { [layerId]: { lookup: { $set: lookup } } });
 };
 
-export function findIndex(layerId, type, value, index, groupStr, state) {
+export function findIndex(layerId, value, index, groupStr, state) {
   index = index || 0;
   const { values } = getPalette(layerId, index, groupStr, state).entries;
   let result;
   lodashEach(values, (check, index) => {
     const min = getMinValue(check);
-    const max = getMaxValue(check);
-    if (type === 'min' && value === min) {
-      result = index;
-      return false;
-    }
-    if (type === 'max' && value === max) {
+    if (value === min) {
       result = index;
       return false;
     }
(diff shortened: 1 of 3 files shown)


ryanweiler92 (approved):

Tomcariello (approved):
This looks good to me.

However, by changing the URL parameter value we change the behavior of old (shared) URLs which is something we like to avoid. I think you should confirm this fix is good with @minniewong specifically before merging.

minniewong (changes requested):
I found an issue with layers with more chunked classifications e.g. Cloud Fraction. This is possibly out of scope/a new ticket?
1. Load Cloud Fraction or http://localhost:3000/?v=-102.66558715448775,-78.38767204653894,230.7386737597858,85.0134261243179&l=MODIS_Aqua_Cloud_Fraction_Day,MODIS_Aqua_CorrectedReflectance_TrueColor
2. Open the options panel and click on "Squash Palette"
3. It now sets the max to 95, and removes the bright red 95-100% classification from the map, like http://localhost:3000/?v=-102.66558715448775,-78.38767204653894,230.7386737597858,85.0134261243179&l=MODIS_Aqua_Cloud_Fraction_Day(max=95,squash=true),MODIS_Aqua_CorrectedReflectance_TrueColor&lg=false&t=2023-08-14-T16%3A04%3A33Z
4. It seems like this should not happen since when looking at the slider, it looks like you are at 100%. Also, if you uncheck the "Squash Palette" box, the max remains 95.

minniewong:
There are also some layers with dual colormaps like Cloud Effective Radius, where the slider needs to be updated into the single slider: http://localhost:3000/?v=-102.66558715448775,-78.38767204653894,230.7386737597858,85.0134261243179&l=MODIS_Aqua_Cloud_Effective_Radius_37,MODIS_Aqua_CorrectedReflectance_TrueColor&lg=false&t=2023-08-14-T16%3A04%3A33Z

minniewong:
Opened new tickets for the two above comments; requesting a solution that doesn't require changing URL patterns

minniewong (approved):

PR #4596, opened 2023-08-23
wv-2702 orbit tracks

Tomcariello:
## Description
We developed the associated layers section for orbit tracks before we had layers that were created from multiple satellites traveling in the same orbit direction and time of day. Now that we have a few layers like that, HLS Sentinel-2A & B and Landsat 8 & 9 and OPERA Surface Water Extent, we should consider including the satellite name as well so you can discern which orbit track is for what satellite.

Fixes # wv-2702.

## How To Test

- git checkout wv-2702-orbit-tracks
- npm ci
- npm run watch
- Load a layer with multiple satellites (such as "OPERA Dynamic Surface Water Extent Provisional (L3)" or "Reflectance (Nadir BRDF-Adjusted)") and a layer with a single satellite (such as 'Deep Blue Aerosol Type')
- Click the 'View Options' icon to open the options menu
- Confirm under "Associated Layers" section the text label includes the satellite name
- Enable an orbit track by clicking on the checkbox(es) in "Associated Layers"
- Confirm that the associated orbit track is listed within the layer row & that this text also includes the satellite name

## Merging

@nasa-gibs/worldview

### DIFF
--- a/web/js/components/layer/settings/associated-layers-toggle.js
+++ b/web/js/components/layer/settings/associated-layers-toggle.js
@@ -32,6 +32,7 @@ function AssociatedLayersToggle(props) {
       <h2 className="wv-header"> Associated Layers </h2>
       { associatedLayers.map((layer) => {
         const { id } = layer;
+        const satelliteName = layer.title.split(' - ')[0];
         const isEnabled = !!activeLayers[id];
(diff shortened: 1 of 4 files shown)


ryanweiler92 (changes requested):
ignore previous comment

ryanweiler92 (approved):

PatchesMaps (approved):

PR #4617, opened 2023-08-29
WV-2480: Highlight Event Tracks on Hover

christof-wittreich:
## Description
This feature adds the ability to hover over an event track, an event menu item, or any of the event track points, and display higher visibility for the event track. The line color, line thickness, arrow thickness, and z-index of the track changes when highlighted like this, as well as the menu item in the Event sidebar also appearing highlighted. This option is specifically for use when all event tracks are shown, for visibility and clarity reasons.

## How To Test
1. `git checkout wv-2480-trackhighlight-hover`
2. `npm ci`
3. `npm run watch`
4. Open the Events tab in the sidebar, then click the filter icon to open the Filter Events window. Enable the option for "Show tracks for all events", then click the Apply button to apply the changes.
5. Hover over an event track and verify that the track becomes highlighted (change in line color, line thickness, arrow size, appears above other track lines) and that the menu item in the Event sidebar becomes highlighted. Clicking on the track should also select the event.
6. Hover over the menu item for an event in the Event sidebar and verify that the track becomes highlighted (change in line color, line thickness, arrow size, appears above other track lines) and that the menu item in the Event sidebar becomes highlighted.
7. Hover over an event track point (one of the circles along the event track) and verify that the track becomes highlighted (change in line color, line thickness, arrow size, appears above other track lines) and that the menu item in the Event sidebar becomes highlighted.
8. Hover over an event icon (the icon indicating event type at one end of an event track) and verify that the track becomes highlighted (change in line color, line thickness, arrow size, appears above other track lines) and that the menu item in the Event sidebar becomes highlighted.


Tomcariello:
When I hover the event track point the line highlights appropriately but the tooltip flickers in/out as I move the mouse. Can we address this flickering to more closely resemble the current production behavior?

web/js/components/sidebar/event.js
```diff
@@ -16,23 +16,25 @@ function Event (props) {
     event,
     eventLayers,
     isSelected,
+    isHighlighted,
     layers,
     removeGroup,
     selectedDate,
     selectEvent,
+    highlightEvent,
```
Tomcariello, line 24: 'highlightEvent' is missing in props validation

web/js/map/natural-events/util.js
```diff
@@ -9,10 +9,12 @@ import { CRS } from '../../modules/map/constants';
 *
 * @param  {Object} clusterPoint
```
Tomcariello, line 10: This function definition should include "proj"
(I realize this isn't related to your work but wanted to mention it since I noticed it)

web/js/map/natural-events/util.js
```diff
@@ -118,10 +131,19 @@ export const getArrows = function(lineSegmentCoords, map) {
   });
 };
 
-export const getTrackLines = function(map, trackCoords) {
+/**
+* Create event tracklines
+*
+* @param  {Object} trackCoords
```
Tomcariello, line 137: This definition is missing the "map" arg

Tomcariello (changes requested):
This looks great to me, just a few housekeeping things & I'd like to see the tooltip addressed if possible.

minniewong (approved):

ryanweiler92 (changes requested):
You may already know this I checked out this branch just now and the e2e test is working fine. The problem seems to have been introduced from the changes that leaves the event track active when the layers tab is selected. To better see this in action you can add an `await page.waitForTimeout(5000)` command to the `event-test.spec.js` file after line 57 and then run the test with `npx playwright test event-test.spec.js --headed`. This will pause for 5 seconds so you can see what is happening right before it fails. 

I played around with the events and was able to replicate similar issues pretty easily by selecting all tracks, selecting a track, clicking the layer tab and then returning to the events tab. My guess is that you will likely have to dig through `event-track.js` and see if there might be a reference that is off, although it could be something else. I would try to add some console commands to see which functions are firing when you select the layers tab and then come back to the events tab. Because that seems to be the catalyst of the issue.

Tomcariello (approved):

web/js/map/natural-events/event-track.js
```diff
@@ -57,10 +62,13 @@ const updateSelection = function(newDate) {
...
  * @param {String} selectedDate
  * @param {Function} callback
+ * @param {Function} callbackHighlight
+ * @param {Function} callbackUnhighlight
  * @param {Function} showAllTracks
+ * @param {Boolean} isHighlighted
  * @return {Object} Object Containing track info and elements
  */
-const getTracksAndPoints = function (eventObj, proj, map, selectedDate, callback, showAllTracks) {
+const getTracksAndPoints = function (eventObj, proj, map, selectedDate, callback, callbackHighlight, callbackUnhighlight, showAllTracks, isHighlighted) {
```
PatchesMaps, line 71: How difficult would it be to refactor this function to reduce the number of arguments? Something like 
```
const getTracksAndPoints = function (eventObj, proj, map, selectedDate, callback, showAllTracks, options) {
  const {
    callbackHighlight,
    callbackUnhighlight,
    isHighlighted,
  } = options
  ...
  ```
  9 arguments for a function just seems like it's getting a bit bloated. No big deal if it would be too much extra work though. IMO, functions with more than 3 arguments can get a little unwieldy to handle.

File: web/js/map/natural-events/util.js [hunk 5]
```diff
@@ -7,12 +7,16 @@ import { CRS } from '../../modules/map/constants';
...
 * @param  {Object} clusterPoint
 * @param  {Boolean} isSelected
+* @param  {Boolean} isHighlighted
 * @param  {Function} callback
+* @param  {Function} callbackHighlight
+* @param  {Function} callbackUnhighlight
 * @return {Object} Openlayers overlay object
 */
-export const getTrackPoint = function(proj, clusterPoint, isSelected, callback) {
+export const getTrackPoint = function(proj, clusterPoint, isSelected, isHighlighted, callback, callbackHighlight, callbackUnhighlight) {
```
PatchesMaps, line 19: Can we reduce the number of arguments in this function?

web/js/map/natural-events/util.js
```diff
@@ -118,10 +133,21 @@ export const getArrows = function(lineSegmentCoords, map) {
...
+* @param  {Object} trackCoords
+* @param  {Boolean} isHighlighted
+* @param  {String} eventID
+* @param  {String} date
+* @param  {Function} callback
+* @param  {Function} callbackHighlight
+* @param  {Function} callbackUnhighlight
+* @return {Object} Openlayers overlay object
+*/
+export const getTrackLines = function(map, trackCoords, isHighlighted, eventID, date, callback, callbackHighlight, callbackUnhighlight) {
```
PatchesMaps, line 149: Can we reduce the number of arguments in this function?

PatchesMaps (changes requested):

PatchesMaps (approved):

PR #4620, opened 2023-08-31
wv-2798: Fix references

Tomcariello:
## Description
Point all URLs to production. Information about the existence of non-production instances should not be available publicly. 

Fixes #wv-2798 .

## How To Test
- git checkout wv-2798
- npm ci
- npm run build
- npm run watch
- Confirm that OrbitTracks Aqua Ascending works properly
- Confirm that Geographic NRT layers work properly

## Merging

@nasa-gibs/worldview


ryanweiler92 (changes requested):
I'm not sure if this was already discussed but there are a few other references to UAT and SIT in the `toolbar.js` file. We reference WV SIT, WV UAT and GIBS UAT in the requestNotifications function. Removing this will probably disable notifications for SIT and UAT but from the ticket description this seems like it needs to happen regardless. 

Removing these may also require some test e2e or unit tests to be updated but I'm not entirely sure.

Tomcariello:
> I'm not sure if this was already discussed but there are a few other references to UAT and SIT in the `toolbar.js` file. We reference WV SIT, WV UAT and GIBS UAT in the requestNotifications function. Removing this will probably disable notifications for SIT and UAT but from the ticket description this seems like it needs to happen regardless.
> 
> Removing these may also require some test e2e or unit tests to be updated but I'm not entirely sure.

Yea, there is a discussion about this in Slack (Aug 31st). Solution TBD...

Tomcariello:
@ryanweiler92 Fix applied to execute the request with less exposure to our test environments in the latest commit.

ryanweiler92 (approved):

minniewong (approved):

PR #4733, opened 2023-10-13
WV-2912 notification count

Tomcariello:
## Description
After some recent updates to the Notification system, the number of notifications is not correctly reflected in the "i" menu. For example, I posted 6 messages, and the "i" menu only says there are 4 messages. I also tried this with 3 and 5 notifications, and each time the "i" menu stated there were 4 notifications. (Images are in the ticket).

The primary cause of this issue is that with the Outage Notifications displaying on page load, SOME notifications are "seen" while others were not. This PR better manage these values so they remain in sync until all notifications have been seen by the user.

Fixes # wv-2912.

## How To Test
- git fetch --all
- git checkout wv-2912-notification-count
- npm run watch
- Either clear your cookies or open a private browser & navigate to http://localhost:3000/
- Notice the red number next to the info button (reflecting the total number of notifications - should be 5 currently)
- Dismiss the Outage Notification window & note the info button badge number is now 3 (because you saw & dismissed the 2 outages). Further, the badge should be yellow now, indicating ALERTS but not OUTAGES.
- Click the Info button & confirm that there is also a "3' next to the Notifications menu item
- Click Notifications to view notifications.
- Dismiss the Notifications window & confirm the yellow badge & the number in the sub menu are now gone because you saw & dismissed all notifications

## Merging

@nasa-gibs/worldview


Tomcariello:
@minniewong I agree that the outages should remain in the notification list. This is addressed in the latest commit. Thanks!

minniewong:
That fix looks good! 

Now lastly 😅, when you have read all the messages, and refresh the page or open a new instance, when I dismiss the 2 outage notifications, the yellow "3" badge still shows with the "i" icon, and there is a grey "3" next to the notifications - but I already read them previously (like 1 minute ago), so shouldn't it be cached that I read them, and I won't be shown any badges?

Tomcariello:
> That fix looks good!
> 
> Now lastly 😅, when you have read all the messages, and refresh the page or open a new instance, when I dismiss the 2 outage notifications, the yellow "3" badge still shows with the "i" icon, and there is a grey "3" next to the notifications - but I already read them previously (like 1 minute ago), so shouldn't it be cached that I read them, and I won't be shown any badges?

Fixed in the latest commit. 🤞

minniewong (approved):
It's working as I would expect it to 😄

ryanweiler92 (changes requested):
It looks like the E2E tests are failing for tests related to the changes being made

ryanweiler92 (approved):

PR #4828, opened 2023-11-22
WV-2886 Tour Story Animation Caching

christof-wittreich:
## Description
This removes the clearing of the cache between steps of a tour story, to eliminate the need for a long loading time for an animation that has already played (and therefore has been cached).

## How To Test
1. `git checkout wv-2886-tourstory-animationcache`
2. `npm ci`
3. `npm run watch`
4. Open any Tour Story with an animation (for example, 'Geostationary Imagery Every 10 Minutes!').
5. Proceed to the step with an animation and click the play button. Let it determine the buffer size and preload the buffer, and let the animation play out.
6. After the animation is played, go forward a step in the tour story, and then go back again to the previous step and press the play button again to play the animation.
7. Verify that the site doesn't need to determine the buffer size and preload the buffer, and that the animation plays immediately after clicking the play button. Verify that the animation looks correct, too.
8. Without reloading the page, finish the tour story, open a different tour story and go through that one, too. Then open the original tour story, go to the step with the animation, play the animation and verify that it did not need to load and that it looks correct.


Tomcariello (approved):

minniewong:
The animations in the Geostationary tour story look pretty good when you change steps, but the Atmospheric Rivers tour story has a strange stutter the first time you return to the animation (I have a video I can send). I wonder if it's because the animation is set to play on URL load? If so, we could test to see if it's better if it doesn't load automatically.

christof-wittreich:
@minniewong I do think it is because of the autoplaying. Do we want to work on trying to fix it to work with autoplaying?

minniewong:
I still seem to encounter the stuttering issue, but it could have been a bandwidth issue. Here are the steps to recreate:
1. Go to Atmospheric Rivers tour Story
2. Go to Step 3, let the animation play through twice
3. Go either forward a step or backward a step
4. Go back to step 3
5. See if the animation stutters on first load.

Tomcariello:
> I still seem to encounter the stuttering issue, but it could have been a bandwidth issue. Here are the steps to recreate:
> 
> 1. Go to Atmospheric Rivers tour Story
> 2. Go to Step 3, let the animation play through twice
> 3. Go either forward a step or backward a step
> 4. Go back to step 3
> 5. See if the animation stutters on first load.

@minniewong I also see the stuttering you mentioned during the first loop in Step #5.

minniewong:
The stuttering is gone but occasionally when I switch steps and come back to step 4, a "preloading buffer" bar shows up and takes ages (or doesn't even) load the animation.
Screenshot 2023-11-30 at 1 54 21 PM

minniewong:
No more stuttering and preloading issues but now I’m seeing issues with imagery loading for some other steps (like step 5) but I can’t tell where this issue is coming from, whether it’s something going on with the imagery or what.

PR #4935, opened 2024-01-16
WV-2887 Preload Tour Story Imagery (no custom palettes)

christof-wittreich:
## Description
This adds preloading to all the imagery in a tour story when a tour story is initiated. Layers that have custom palettes, custom min/max, and disabled values are not preloaded.

## How To Test
1. `git checkout wv-2887-preload-tourstory-noncustom`
2. `npm ci`
3. `npm run watch`
4. Open any Tour Story and navigate through each step.
5. Verify that the imagery for each step is already preloaded before being the step is rendered onscreen.


web/js/containers/tour.js
```diff
@@ -516,12 +527,47 @@ const mapDispatchToProps = (dispatch) => ({
...
       dispatch({ type: LOCATION_POP_ACTION, payload: location });
     }
   },
+  preProcessStepLink: async (search, config, promiseImageryForTour) => {
+    search = search.split('/?').pop();
+    const parameters = util.fromQueryString(search);
+    let layers = [];
+    let layers2 = [];
+    const promises = [];
+    const temp = [];
```
Tomcariello, line 546: Can we rename `temp` to something more meaningful?

web/js/containers/tour.js
```diff
@@ -516,12 +527,47 @@ const mapDispatchToProps = (dispatch) => ({
...
       });
     } else {
       dispatch({ type: LOCATION_POP_ACTION, payload: location });
     }
   },
+  preProcessStepLink: async (search, config, promiseImageryForTour) => {
+    search = search.split('/?').pop();
+    const parameters = util.fromQueryString(search);
+    let layers = [];
+    let layers2 = [];
```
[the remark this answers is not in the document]
christof-wittreich: Do you have a suggestion on a more descriptive name? `layers` is for all layers from `parameters.l`, and `layers2` is all the layers from `paremeters.l1` (secondary layers on a split screen).

minniewong (approved):
Definitely reduces the transitions between steps when the screen goes black. Some of the stories that jump from place to place across the world still struggle a little with loading the tiles, but this is a big improvement!

web/js/containers/tour.js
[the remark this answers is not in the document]
Tomcariello: Ha, I hate parameters.l & parameters.l1, too. Maybe I'm just fussy.

Comparison mode is typically referred to as A & B sides, so `layersA` & `layersB` would be more consistent with the rest of the codebase.

web/js/containers/tour.js
```diff
@@ -516,12 +527,47 @@ const mapDispatchToProps = (dispatch) => ({
...
+    let layersA = [];
+    let layersB = [];
+    const promisesParams = [];
+
+    if (parameters.l) {
+      layersA = layersParse12(parameters.l, config);
+      layersA = uniqBy(layersA, 'id');
+      layersA = layersA.filter((layer) => !layer.custom && !layer.disabled);
+      promisesParams.push({ layersA, dateString: parameters.t });
+      if (parameters.l1) {
```
Tomcariello, line 552: Is there a reason the parameters.l1 IF is inside the parameters.l IF? Why not 2 separate if statements?

christof-wittreich: I don't think there can be an instance where there can be `l1` without `l`, so I nested them to save on an unnecessary comparison. I'm happy to bring it out so it stays unnested though

web/js/containers/tour.js
```diff
@@ -165,6 +177,13 @@ class Tour extends React.Component {
       config,
       renderedPalettes,
     );
+    currentStory.steps.forEach((step) => {
+      preProcessStepLink(
```
Tomcariello, line 181: Is it possible to execute this step AFTER the first step loads? I am seeing significant slowdown on the first step of the story that is remediated by removing this code block.

We are also PreProcessing Step 1 here which is duplicated effort. Can we exclude step 1 from pre-processing?

christof-wittreich: To fix the slowdown on the first step, I changed it so that we only preload the next step, rather than all steps at the beginning. This also solves the duplicated effort on step 1 pre-processing.

Tomcariello (approved):

PR #5121, opened 2024-04-17
WV-2865 Classification/Palettes Layers Classification Toggle Behavior

christof-wittreich:
## Description
This fixes the issue of layers that have both classifications and a palette not properly updating when a classification is enabled/disabled. It also fixes behavior when adding a min or max to the palette, and then toggling a classification (previous behavior would incorrectly ignore the min/max set when classification is toggled). It also fixes behavior with refreshing the page with all classifications turned on when the default state for the layer includes disabled classifications (previous behavior incorrectly loaded default state rather than retain classification toggle state).

## How To Test
1. `git checkout wv-2865-classification-toggle`
2. `npm ci`
3. `npm run watch`
4. Add any layer that has classification toggles and a palette (like SSMIS_Sea_Ice_Concentration_Snow_Extent).
5. Verify that toggling on/off classifications correctly updates the layer itself.
6. Verify that setting a min or max for the palette and then toggling a classification still displays the correct previous set min/max on the layer itself.
7. Verify that enabling all classifications and reloading the page does not cause the classifications that are disabled by default to re-enable incorrectly.

### DIFF
--- a/web/js/modules/vector-styles/selectors.js
+++ b/web/js/modules/vector-styles/selectors.js
@@ -12,6 +12,10 @@ import { stylefunction } from 'ol-mapbox-style';
 import {
   getMinValue, getMaxValue, selectedStyleFunction,
 } from './util';
+import {
+  isActive as isPaletteActive,
+  getLookup as getPaletteLookup,
+} from '../palettes/selectors';
 import util from '../../util/util';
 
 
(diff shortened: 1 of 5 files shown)


minniewong (approved):

web/js/modules/vector-styles/selectors.js
```diff
@@ -115,6 +119,25 @@ const updateGlStylePalette = (glStyle, rgbPalette) => {
...
+    const thisCircleColor = glStyle.layers[i].paint['circle-color'];
+    thisCircleColor.forEach((color, index) => {
+      const regex = /rgba?\(.*\)/;
+      if (regex.test(color)) {
+        const colors = color.split('(')[1].split(')')[0].split(/,\s?/);
+        if (colors.length < 4) {
+          colors.push('255');
+        }
+        if (lookup[colors.join(',')]) {
+          thisCircleColor[index] = `rgba(${lookup[colors.join(',')].r}, ${lookup[colors.join(',')].g}, ${lookup[colors.join(',')].b}, ${lookup[colors.join(',')].a})`;
```
Tomcariello, line 133: We should only perform the join once:

```
colorStr = colors.join(',')
if (lookup[colorStr])  {
   thisCircleColor[index] = `rgba(${lookup[colorStr].r}, ${lookup[colorStr].g}, ${lookup[colorStr].b}, ${lookup[colorStr].a})`;
```

Tomcariello (approved):

PR #5244, opened 2024-05-29
WV-3190 Smallest Subdaily Timeline Interval

christof-wittreich:
## Description
This code change changes the behavior of the timeline interval for subdaily layers. Firstly, it allows for subdaily layers to set a custom interval of something other than 10 minutes, and secondly, if multiple subdaily layers are present and have multiple custom intervals, the smallest interval is chosen.

## How To Test
1. `git checkout wv-3190-smallest-interval-subdaily`
2. `npm ci`
3. `npm run watch`
4. Open a fresh worldview instance and add a subdaily layer with a custom interval of something other than 10 minutes, like `VIIRS_NOAA20_CorrectedReflectance_TrueColor_Granule`.
5. Verify that the timeline interval correctly reflects the custom interval.
6. Add another subdaily layer, with a different custom interval than the first subdaily layer added, like `Himawari_AHI_Air_Mass`.
7. Verify that the timeline interval correctly reflects the smaller of the two custom intervals.
8. Repeat the process with the same layers in a new worldview instance, but add the second layer from before first. Verify the behavior is the same.

### DIFF
--- a/web/js/containers/timeline/timeline.js
+++ b/web/js/containers/timeline/timeline.js
@@ -33,7 +33,10 @@ import {
   dateRange as getDateRange,
   hasSubDaily,
   subdailyLayersActive,
+  subdailyLayers,
   getActiveLayers,
+  getSubDaily,
+  getSmallestIntervalValue,
 } from '../../modules/layers/selectors';
 import { getSelectedDate, getDeltaIntervalUnit } from '../../modules/date/selectors';
 import {
@@ -278,7 +281,6 @@ class Timeline extends React.Component {
       animEndLocationDate,
       changeCustomInterval,
       customInterval,
-      customSelected,
       dateA,
       dateB,
       interval,
@@ -286,6 +288,8 @@ class Timeline extends React.Component {
       isAnimationWidgetOpen,
       isGifActive,
       hasSubdailyLayers,
+      subDailyLayersList,
+      newCustomDelta,
     } = this.props;
     const { frontDate, draggerTimeState, draggerTimeStateB } = this.state;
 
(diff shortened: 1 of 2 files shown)


minniewong (commented):
If you add a layer that has a custom interval (like a Level 2 TEMPO layer) then remove it, the interval selector now says "1440 minutes". Would be better if it went back to "1 Day".

Screenshot 2024-05-29 at 3 51 48 PM

minniewong (approved):

PatchesMaps (approved):

web/js/modules/layers/selectors.js
```diff
@@ -590,6 +590,44 @@ export const subdailyLayersActive = createSelector(
...
+);
+
+/**
+ * Gets smallest interval value of subdaily layers
+ * @param {Object} state
+ */
+export function getSmallestIntervalValue(state) {
+  const layers = getActiveLayers(state);
+  let smallestDelta = 1440;
+  if (layers && layers.length) {
```
Tomcariello, line 621: I would refactor this to only call `lodashGet(layers[i], 'dateRanges[0].dateInterval')` once.

 ```if (layers && layers.length) {
  for (let i = 0; i < layers.length; i += 1) {
    let interval = lodashGet(layers[i], 'dateRanges[0].dateInterval');
    if (layers[i].period === 'subdaily' && interval < smallestDelta) {
      smallestDelta = Number(interval);
    }
  }
}

web/js/containers/timeline/timeline.js
```diff
@@ -309,17 +313,21 @@ class Timeline extends React.Component {
...
     const subdailyInterval = customInterval > 3 || interval > 3;
 
     if (subdailyRemoved && subdailyInterval) {
       changeCustomInterval();
       selectInterval(1, TIME_SCALE_TO_NUMBER.day, false);
     }
 
-    if (subdailyAdded && !customSelected) {
-      changeCustomInterval(10, TIME_SCALE_TO_NUMBER.minute);
+    if (subDailyCountChanged) {
```
Tomcariello, line 325: I'd prefer if we clearly define the 1440 value for readability sake.

```
const isSubDaily = newCustomDelta !== 1440; // or < 1440, if that makes more sense
if (subDailyCountChanged) {
  if (isSubDaily) {
    changeCustomInterval(newCustomDelta, TIME_SCALE_TO_NUMBER.minute);
  } else {
    changeCustomInterval(1, TIME_SCALE_TO_NUMBER.day);
  }
}

PR #5332, opened 2024-07-05
WV-3242 index out of bounds error

PatchesMaps:
## Description

> When adding more than one TEMPO layer, sometimes there is an "Index out of bounds" error displayed in the console

## How To Test

1. `git checkout WV-3242-index-out-of-bounds`
2. `npm ci && npm run build && npm run watch`
3. Go to this [link](http://localhost:3000/?v=-177.36395603910125,-25.17776302922428,-4.107158456019675,74.38473697077572&ici=3&icd=1&l=Coastlines_15m&lg=true&t=2024-07-02-T00%3A57%3A00Z)
4. Add several TEMPO L2 layers and verify that there are no `index out of bounds` errors.

## PR Submission Checklist

## Merging

@nasa-gibs/worldview


web/js/mapUI/components/layers/addLayer.js
```diff
@@ -51,7 +51,12 @@ function AddLayer(props) {
       compareMapUi.create(ui.selected, mode);
     } else {
       const newLayer = await createLayer(def);
-      ui.selected.getLayers().insertAt(index, newLayer);
+      const layers = ui.selected.getLayers();
+      if (index <= layers.getLength()) {
```
christof-wittreich, line 55: This feels like a bit of a hacky fix to me. Why is this behavior different than when adding any other layers? Surely there is a solution to this further upstream that solves the actual problem, rather than accounting for the problem later on.

PatchesMaps: I'm not sure fixing the source of the bug is feasible within the scope of this story. The source of the issue is a race condition caused by [this useEffect](https://github.com/nasa-gibs/worldview/blob/main/web/js/mapUI/components/update-projection/updateProjection.js#L367) hook which is unfortunately critical to how AERONET layers are rendered. Modifying said hook to only reload layers when the newly rendered palette is from AERONET works for the most part but unfortunately causes AERONET layers to not render correctly when hydrated from the url.

minniewong (approved):

Tomcariello (approved):

christof-wittreich (approved):

PR #5347, opened 2024-07-17
WV-3220 Smaller Extent BBox For Finding High-Resolution Imagery

christof-wittreich:
## Description
This change modifies the bounding box used when searching for high-resolution imagery to be smaller than currently. This is to prevent situations where imagery is found, but only exists in the very corner of the extent. The new bounding box is 60% of the window width and 80% of the window height, centered on the middle of the window.

## How To Test
1. `git checkout wv-3220-smaller-extent-bbox`
2. `npm ci`
3. `npm run watch`
4. Open a fresh instance of Worldview
5. Add a layer that uses the imagery-search feature, like `HLS_S30_Nadir_BRDF_Adjusted_Reflectance`
6. Zoom in and position the screen so that imagery is sometimes in the center of the screen, but sometimes on the edges
7. Open the imagery-search feature and click to various other dates, ensuring that no dates shown in the search have imagery only on the edges of the screen and none towards the center

### DIFF
--- a/web/js/components/layer/settings/imagery-search.js
+++ b/web/js/components/layer/settings/imagery-search.js
@@ -46,8 +46,17 @@ export default function ImagerySearch({ layer }) {
       }
       return maxExtent[i];
     });
+    const xDiff = Math.abs(extent[0] - extent[2]);
+    const yDiff = Math.abs(extent[1] - extent[3]);
+    // Reduce width by 40% and height by 20%, to show only centered data
+    const smallerExtent = [
+      extent[0] + (xDiff * 0.2),
+      extent[1] + (yDiff * 0.1),
+      extent[2] - (xDiff * 0.2),
+      extent[3] - (yDiff * 0.1),
+    ];
     try {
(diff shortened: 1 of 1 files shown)


PatchesMaps (approved):

web/js/components/layer/settings/imagery-search.js
```diff
@@ -46,8 +46,17 @@ export default function ImagerySearch({ layer }) {
       }
       return maxExtent[i];
     });
+    const xDiff = Math.abs(extent[0] - extent[2]);
```
Tomcariello, line 49: Since we are using this block (lines 49-57) again at lines 79-87 I'd like to see this put into a function.

christof-wittreich: Lines 42-48 are also reused and had already existed there before my changes, which is why I did it this way. I can refactor it and bring all of that out into a function though

Tomcariello: I agree, lets simplify things where possible. Thanks!

minniewong (approved):

PR #5453, opened 2024-09-13
fix: style alignments

gabrielogregorio:
## Description

Fixes # .

- [X] On mobile, the alignment of the Camera and period icons was adjusted

Before VS with this PR

Captura de tela de 2024-09-12 22-46-51

Captura de tela de 2024-09-12 22-47-27

## How To Test

in development mode, valid and the styling is correct, on mobile and desktop they should be perfect and without regressions. At first I only changed the camera icon and the text "2024 SEP 12"

- Open http://127.0.0.1:3000/
- Stylizations must be correct

I couldn't think of a better name than `button-action-group-margin-fix`, suggestions?


minniewong:
@gabrielogregorio Thank you for your contribution though this fix overrides our responsive formatting so we cannot merge this PR.

PR #5514, opened 2024-10-17
WV-3375: Update LANCE acronym, remove www from earthdata web address

minniewong:
## Description

Fixes # WV-3375

Update LANCE acronym, remove www from earthdata web address

## How To Test

1. Go to the About page
2. Find the section about LANCE, see that it's been updated to "Earth observation" from "Eo"
3. Check to see links that were www.earthdata.nasa.gov in layer descriptions now work as earthdata.nasa.gov, except for https://www.earthdata.nasa.gov/apt/documents/chlor-a/v1.0 which does not work without www.

@nasa-gibs/worldview


PatchesMaps (approved):

minniewong:
> https://www.earthdata.nasa.gov/apt/documents/chlor-a/v1.0 does seem to work without the www, although it seems to load a slightly different version of the site. I assume this is what you mean by it doesn't work?

All I get is a page that says "loading" when I go to the site without www. I need to check with someone what's is going on with that.

christof-wittreich (approved):

PR #5676, opened 2025-01-31
WV-3459 Charting Release

christof-wittreich:
## Description
This change adds a more clear error message when the chart request fails, and allows charting options to be included in permalinks. It also adds multiple changes based on feedback about the function of the entire mode, including both UI updates and functionality updates.

## How To Test
1. `git checkout wv-3459-charting-error-message`
2. `npm run build`
3. `npm ci`
4. `npm run watch`
5. Open a fresh instance of Worldview
6. Add multiple layers that can be used for charting, such as `OMI_Aerosol_Index` and `MODIS_Combined_Value_Added_AOD`
7. Open the charting interface via the 'Start Charting' button
8. Select which layer is selected for charting, change the dates, select an AOI, and generate a chart
9. Reload the page with the chart still open, and verify that the page re-requests that chart with the same parameters and options automatically on reload
10. Verify that changing the charting options to other options and reloading the page works as well
11. Very that changing the AOI updates the coordinates, and vice versa
12. Verify that charting continues to work correctly


Tomcariello (approved):
To anyone testing, make sure you change features.charting: true in `config\default\common\features.json` prior to building.

We need to strongly consider a progress/status bar, especially with the chart request on permalink. Users are not going to know they need to wait a few seconds unless we make that clear.

minniewong:
Screenshot 2025-03-04 at 5 22 42 PM
This one is probably an outlier that we can address later, but the units are overwhelming :)

minniewong:
We should not allow users to try to make charts in the "wings" as it seems not to work
Screenshot 2025-03-04 at 5 46 53 PM

minniewong:
Screenshot 2025-03-04 at 5 48 45 PM
The text is really far apart from each other for the one day stats.

minniewong:
If you set up a chart query with the full map, [save the URL](http://localhost:3000/?v=-398.3082687323691,-197.6902372000871,260.78839140620875,205.65835270162006&l=Reference_Labels_15m(hidden),Reference_Features_15m(hidden),Coastlines_15m,OMI_Absorbing_Aerosol_Optical_Depth,VIIRS_NOAA21_CorrectedReflectance_TrueColor(hidden),VIIRS_NOAA20_CorrectedReflectance_TrueColor(hidden),VIIRS_SNPP_CorrectedReflectance_TrueColor(hidden),MODIS_Aqua_CorrectedReflectance_TrueColor(hidden),MODIS_Terra_CorrectedReflectance_TrueColor(hidden)&lg=true&cha=true&chl=OMI_Absorbing_Aerosol_Optical_Depth&chc=-179.5,-89.5,179.5,89.5&cht=2025-02-19-T05%3A00%3A00Z&cht2=2025-02-26-T05%3A00%3A00Z&t=2025-02-19-T05%3A00%3A00Z), then open that URL in another window, the bounding box coordinates in the edit coordinates window and on the corners of the bounding box change to some other coordinates. Subsequently, if I click "Generate Chart" here, I get the same chart back as the original chart, but I can't 100% be sure if it's using the correct bounding coordinates. Screenshot 2025-03-05 at 3 36 46 PM

minniewong (approved):

Tomcariello:
I think this PR looks very good! Here is some general feedback of the feature overall.

- You can't pan/zoom with the mouse when in charting mode, possibly due to the coordinate box being open? The Zoom +/- controls work & you can pan/zoom if you collapse the layer list. It feels a bit restrictive to me.
- I think the Alert/Note at the bottom of the chart is lacking context. I am told the # of data points has been reduced from 7460 to 31 but 7460 is meaningless to me. Can we include the period in the values (i.e. 7460 days to 31 days) so it has some context? 
- The charting button being unavailable on page load might be a bit frustrating for users. The button is there but what is it? How can I activate it? What does it do? Perhaps we can have some hover text to nudge users towards selecting a chartable layer?
- If you have multiple chartable layers in the layer list, all the imagery appears on the map which is a bit misleading since we can currently only chart 1 layer at a time. Could we only show the selected layer imagery on the map?

In the future:
- Clicking a point on the chart should take the user to that particular imagery. I ran a chart with Aerosol Index from Sept 2004 - March 2025 & 2007-2008 show a large reduction in the charted mean. Being able to quickly navigate to that imagery might be a nice addition. 
- A walkthrough/tour story to introduce Charting mode would be nice.

Tomcariello (approved):

PR #5685, opened 2025-02-05
Feature/add lerc decoder

grallewellyn:
NOTE THAT THIS IS STILL A WIP: I need to add LERC to the endpoints so it doesn't need to be hard coded in but I am not sure what the worldview standard process is for adding a layer
Right now, I am accessing the lerc layers by running Joe's ec2 instance and using localhost, so we need to publish the lerc layers and also add more than 1 day of data 
Once the LERC layer is added to the official endpoints, we can get the snapshot that matches the rest of the snapshots by running `npm run build` which gets the snapshot with a request to https://wvs.earthdata.nasa.gov/api/v1/snapshot
I am keeping the incorrect `web/images/layers/previews/geographic/VIIRS_VNP46A1_LERC_v1.jpg` in this PR as a reminder for now 

I also still need to create tests that the LERC layers are showing up properly 

## Description
Adds the LERC decoder to worldview so that LERC layers are now supported. Some other notes:
- I removed the original difference/ average code for the LERC color decoder https://git.earthdata.nasa.gov/projects/ONEARTH/repos/onearth-data-client/browse/javascript/lerc_ol3_prototype.js#764-809 because it did not appear to be a feature in worldview. This part of `drawTile` was for comparing the average/ difference between 2 layers. The closest feature in worldview was comparison mode which is just displaying 2 layers normally but side by side 
- Using default gray scalar function for original coloring 
```
export function getGreyScalar(val, min, max) {
  const colors = [];
  colors[0] = 255 * ((val - min) / (max - min));
  [colors[1], colors[2]] = [colors[0], colors[0]];
  return colors;
}
```
We agreed this was okay in a meeting 2/4/2025
- When the filter is on for `drawTile` (which is the case if the user altered the layer thresholds), then don’t show values above or below the current threshold 
- For setting the no data value, I didn’t want to alter the disabled key in the palette object because that would affect other layers, so I added a new tag- noDataValue. I did not see anything for schema for the palette, so it seems okay to add a new key to palette objects. Also, worldview during the build stage breaks the colormaps up if there is a transparent/ noDataValue, so I took this into account 
- I could have extracted out setting the lerc tileLoadFunction, but I think keeping it in createLayerWMTS is best for right now because we only support LERC with WMTS at the moment 
- I know the 2 if statements checking if the layer is a lerc layer and then doing a different action (not setting the tileClass and setting the tileLoadFunction) are not the prettiest, but I spent quite a bit of time trying to get around them and could not figure it out. Maybe another worldview developer could give suggestions or we can create a new ticket for it. I think there should be a better way to set a custom tileLoadFunction for other use cases and this would entail having this custom tile load function properly called when the palette changes (the reason I can’t set the tileClass for LERC layers is because the LERC til loader is not called when the palette is changed so the LERC layer will not display properly) 
- I know passing the state object to tileLoader isn’t ideal, but we need to be able to get the resolution, palette, and legend from the state in the LERC tileloader

Fixes [VISLABS-195](https://bugs.earthdata.nasa.gov/browse/VISLABS-195)

- Adding LERC layers to worldview 

[If this is a relatively large or complex change, start a discussion by explaining why you chose the solution you did and what alternatives you considered, etc...]
Potential discussion topics in bullets above 

## How To Test

[Provide whatever information a reviewer might need to know in order to verify that the changes made in this PR are working as expected. If there are special build steps that need to be taken in order to get these changes to run (building while on a separate branch, running `npm ci`, etc) include them here.]

- For bugfixes: What steps need to be taken in the UI to verify the bug is fixed?
- For enhancements and features: What is the expected functionality being added/modified? How can a reviewer verify this?

Can now open VIIRS Nighttime Imagery LERC layer to the map and view it. This LERC layer should have all functionality like changing the palette, min/max values, and opacity. All other layers and behavior should function the same 

1. Open with these URL parameters: http://localhost:3000/?lg=true&t=2022-07-01-T18%3A36%3A48Z
2. Add layers
3. Add the VIIRS lerc layer by searching it 
4. Verify expected result

I verified that I could delete my local copy of worldview and rebuild/ run 

## PR Submission Checklist

## Merging

@nasa-gibs/worldview


config/default/common/config/wv.json/layerOrder.json
```diff
@@ -56,6 +56,7 @@
 		"VIIRS_NOAA21_CorrectedReflectance_TrueColor",
 		"VIIRS_NOAA21_CorrectedReflectance_BandsM3-I3-M11",
 		"VIIRS_NOAA21_CorrectedReflectance_BandsM11-I2-I1",
+		"VIIRS_VNP46A1_LERC_v1",
```
jtroberts, line 59: We should rename this layer to maybe VIIRS_SNPP_DayNightBand_LERC.

grallewellyn: I agree and I will change once we ingest more data under that ID since the ID here is used for the imagery requests

PR #5783, opened 2025-03-28
Block lightning layer from Charting

christof-wittreich:
## Description
This change blocks the `LIS_Very_High_Resolution_Lightning_Daily_Climatology_LIS_Mean_Flash_Rate` layer from being charted.

## How To Test
1. `git checkout charting-block-lightning-layer`
2. `npm ci`
3. `npm run build`
4. `npm run watch`
5. Open a new instance of WV.
6. Add `LIS_Very_High_Resolution_Lightning_Daily_Climatology_LIS_Mean_Flash_Rate`, then verify that the Charting button is still greyed-out.
7. Add a second layer that is actually chartable, then enter Charting mode and verify that `LIS_Very_High_Resolution_Lightning_Daily_Climatology_LIS_Mean_Flash_Rate` is not visible in Charting mode as a valid layer to chart.

### DIFF
--- a/config/default/common/config/wv.json/layers/trmm/LIS_Very_High_Resolution_Lightning_Daily_Climatology_LIS_Mean_Flash_Rate.json
+++ b/config/default/common/config/wv.json/layers/trmm/LIS_Very_High_Resolution_Lightning_Daily_Climatology_LIS_Mean_Flash_Rate.json
@@ -6,7 +6,8 @@
       "tags": "",
       "group": "overlays",
       "layergroup": "Lightning",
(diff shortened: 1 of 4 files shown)


minniewong (approved):

Tomcariello (approved):

christof-wittreich:
@PatchesMaps That is probably a better way to do it. I've changed it to use layer config now instead

PatchesMaps:
@minniewong is `disableCharting` something we want to add to the docs?

PatchesMaps (approved):

PR #5789, opened 2025-04-01
WV-3450 Show Dams Layer When Disabling Classifications Fix

christof-wittreich:
## Description
Raster/zoomed out dams don't show any dams when any classification is disabled/turned off. Vector dams are fine.

Steps to replicate:

Load Worldview
Load the Dams layer
Open the settings/options panel and turn off one of the classifications, e.g. the yellow hydroelectricity classification
See error - none of the dams appear on the map. If you zoom in, the vector dams are correct and display all classifications except hydroelectricity

## How To Test
1. `git checkout wv-3450-dams-disabled-hidden`
2. `npm ci`
3. `npm run watch`
4. Open [this link](http://localhost:3000/?l=GRanD_Dams,Coastlines_15m,MODIS_Terra_CorrectedReflectance_TrueColor&lg=false&t=2025-04-01-T17%3A17%3A31Z)
5. Verify that the dams layer is visible and the colored points are visible.
6. Disable one of the classifications of the dams layer, then verify that that color point can no longer be seen on the layer, but other colored points can still be seen. Do this for both when the layer is zoomed in enough that the points are clickable, as well as when the layer is zoomed out further and the points are not clickable.


Tomcariello (approved):

minniewong:
Not sure how "fixable" this would be but if you turn off all the classifications and just leave one on, e.g. the yellow hydroelectricity one. You can still see that the ghosts of the other classifications overlaid on the yellow.
Screenshot 2025-04-04 at 3 13 01 PM

christof-wittreich:
@minniewong This is the case if you turn off any of the classifications. Since we are dealing with a single-layer .png file, all we can do is hide the colors that we have disabled. But if the original png file we get has another color overtop of a yellow dot, we'll only be able to see what isn't covered, even if we hide other colors.

This is the solution we decided earlier to go with for the time being, so that the layer can still be displayed properly when classifications are disabled, and I believe we wanted to pursue having the layer be a vector layer at any zoom level as the future solution to this issue.

minniewong (approved):

PR #5858, opened 2025-05-02
WV-3594 Implented New TiTiler Endpoint

christof-wittreich:
## Description
This change converts all TiTiler layers to the new endpoint.

## How To Test
1. `git checkout wv-3594-new-titiler`
2. `npm run build`
3. `npm ci`
4. `npm run watch`
5. Verify all TiTiler layers function correctly and the same as before
6. Verify the "custom" TiTiler layers function correctly and the same as before


Tomcariello (approved):

minniewong (changes requested):
I am noticing that, in my testing of the HLS Landsat Customizable layer, the simple band math indices work fine, but the more complex ones like MSAVI, SAVI, and EVI aren't returning imagery.

PR #5975, opened 2025-06-27
WV-3628 - Support EONET query by id instead of title

christof-wittreich:
## Description
This change modifies the way we handle EONET event objects to sort them by `id` instead of `title`. This allows for EONET to change their event category titles, since their ids will stay consistent and can still be used for sorting.

This change also replaced the old CustomToolip with UncontrolledTooltip, as the former seemed to have some issues with the edge of the screen and persistence.

## How To Test
1. `git checkout wv-3628-eonet-id-query`
2. `npm ci`
3. `npm run watch`
4. Open the Events tab on the sidebar
5. Verify that all icons are correctly present for each category (filter by each category individually to verify)
6. Verify that selecting an event of each kind still works correctly, and the icon on the map is still correct
7. Verify that hovering the icons shows the correct titles, both in the list and on the map
8. Verify that special behavior for some events still works correctly (different zoom levels for some categories, etc.)
9. Verify that there are no additional places in the natural events section of the code that need to be changed from title to id

### DIFF
--- a/e2e/test-utils/global-variables/selectors.js
+++ b/e2e/test-utils/global-variables/selectors.js
@@ -255,7 +255,7 @@ module.exports = (page) => ({
   thermAnomVIIRSnight: page.locator('#active-VIIRS_NOAA20_Thermal_Anomalies_375m_Night'),
   notifyMessage: page.locator('.wv-alert .alert-content'),
   filterIcons: page.locator('.filter-icons > div > .event-icon'),
-  dustHazeIcon: page.locator('.filter-icons > div > #filter-dust-and-haze'),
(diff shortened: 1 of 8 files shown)


minniewong (changes requested):
I'm having some issues with the tooltips and are staying open when that is undesired. Steps to recreate:
1. Click on an event e.g. Wildfires in Angola
2. Prevent the map from zooming to that location by moving the map
3. See that the tooltip remains on the screen
4. Zoom out and click on another event e.g. Tropical Storm Barry
5. See tooltip get stuck on the screen


Screenshot 2025-06-30 at 2 02 45 PM

minniewong (approved):

PR #6054, opened 2025-08-06
WV-3590 - Chart Window Additions

christof-wittreich:
## Description
This change adds a few enhancements to the Charting Mode chart window. Specifically: a minimap displaying the location of the bounding box on the world map, a list of the coordinates in the user's preferred format, the start and end dates (in the modal header), and a warning of any error days (if applicable) that expands to show the specific error dates.

This PR also includes mock data, which will be removed after approval but before merging.

## How To Test
1. `git checkout wv-3590-charting-minimap`
2. `npm ci`
3. `npm run watch`
4. Add any chartable layer to the layer list, then open Charting Mode and generate a chart (will use the mock data).
5. Verify that the coordinates shown in the chart window are correct and match the format selected by the user.
6. Verify the start and end dates in the modal header are correct.
7. Verify the minimap shows the correct area selected for the bounding box selected for the chart, and that bounding boxes both very large and very small display correctly. This includes zooming out far and drawing a box, and zooming in to different zoom levels and drawing a box.
8. Verify that the minimap can be zoomed in/out via the buttons attached to it.
9. Verify that error days are shown correctly if there are any, and verify that the warning isn't shown if there are no error days (will require removing them from the mock data). Verify the warning expands and collapses when the button is clicked.


minniewong (changes requested):
In Safari, the heading dates are rendered as NaN. (works fine in Chrome and Firefox)

Safari_NaNerror

PR #6070, opened 2025-08-14
WV-3579 - Prevent pan/zoom when in Charting Mode

christof-wittreich:
## Description
This change disables panning or zooming via any means while charting mode is enabled. It also disables certain features when in charting mode, similar to features disabled when in comparison mode.

## How To Test
1. `git checkout wv-3579-charting-panzoom`
2. `npm ci`
3. `npm run watch`
4. Add any chartable layer to the layer list, then open Charting Mode.
5. Verify that the zoom in/out buttons on the right-side of the screen are hidden, and that the map cannot be panned or zoomed.
6. Verify that collapsing the sidebar doesn't allow panning or zooming either.
7. Verify that snapshots, switching projections, animations, and switching tabs while in charting mode are disabled.

### DIFF
--- a/web/js/components/map/zoom.js
+++ b/web/js/components/map/zoom.js
@@ -6,7 +6,7 @@ import { mapUtilZoomAction } from '../../map/util';
 import HoverTooltip from '../util/hover-tooltip';
 
 function Zoom({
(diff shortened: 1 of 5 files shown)


Tomcariello (commented):
Everything you did looks good, but we should disable the 'change projection' button when in charting mode as well.

PR #6193, opened 2025-10-08
WV-3698 - Auto interval for variable time periods

christof-wittreich:
## Description
This change adds an Auto interval option when TEMPO imagery is added. Selecting it will allow forward and backwards advancing of the timeline to skip to the next TEMPO imagery point, even if the intervals are not the same. It also works for animation and gif download features.

## How To Test
1. `git checkout wv-3698-variable-timeperiods-support`
2. `npm ci`
3. Change `sources.json`, `processTemporalLayer.js`, and `describe-domains.worker.js` to point to the correct endpoint
4. `npm run build`
5. `npm run watch`
6. Go to [this url](http://localhost:3000/?v=-201.3639833339281,-43.20495955209448,-6.289054554777408,103.26218994380248&z=4&i=5&ias=true&l=TEMPO_L3_Formaldehyde_Vertical_Column,Coastlines_15m&lg=true&t=2025-09-12-T14%3A04%3A59Z) (L3) and use the Auto interval selected to go forward and backwards along the timeline. Verify that the time the Auto interval selects lines up with the next/previous imagery on the timeline
7. Use the Animation and GIF features with the Auto interval selected and verify they work as intended and show new imagery every frame
8. Go to [this url](http://localhost:3000/?v=-199.6548335008044,-19.790488565928655,-8.826708500804429,80.70348929865466&z=4&i=5&ias=true&l=TEMPO_L2_Cloud_Cloud_Fraction_Total_Granule(count=1),Coastlines_15m&lg=true&t=2025-09-12-T19%3A11%3A29Z) (L2) and do the same process as above


web/js/containers/gif.js
```diff
@@ -408,18 +408,20 @@ class GIF extends Component {
...
   } = animation;
   const { screenWidth, screenHeight } = screenSize;
   const {
-    customSelected, interval, customInterval, customDelta,
+    customSelected, smartSelected, interval, customInterval, customDelta,
   } = date;
   const increment = customSelected
     ? `${customDelta} ${TIME_SCALE_FROM_NUMBER[customInterval]}`
-    : `1 ${TIME_SCALE_FROM_NUMBER[interval]}`;
+    : smartSelected
```
PatchesMaps: Can we refactor this to avoid nested ternaries?

ZachTRice (approved):

PR #6560, opened 2026-04-06
WV-3739: Fix CMR calls blocking UI layer rendering, implemented async…

vicgonzalez25:
## Description

Fixes WV-3739

CMR API calls for granule layers were blocking all other layers from rendering. You'd add a granule layer, if CMR calls are slow, user have to wait for the CMR response before seeing anything else on the map.

This PR introduces async layer loading where each layer renders as soon as it's ready, instead of waiting for slow ones to finish.

- [x] Async layer loading - fast layers (WMTS) render immediately, slow layers (CMR/granule) render when ready
- [x] Race condition fix - toggling "Group Similar Layers" rapidly no longer corrupts layer order
- [x] Granule reload fix - granule layers now render properly on page refresh/share links
- [x] Early visibility updates - base layers render without waiting for CMR responses

### The approach

Each layer is inserted into the map using `insertAt()` as soon as it resolves, with the correct z-index calculated on the fly. The tricky part was handling race conditions - when you toggle grouping or change dates quickly, multiple `reloadLayers()` calls can overlap. We track operation IDs and skip insertions from superseded operations to prevent order corruption.

### Setup: Throttle CMR calls
To properly see the async loading behavior, throttle CMR requests in DevTools:
1. Open DevTools → Network tab
2. Click "Add request blocking condition" (or right-click a CMR request → Block request URL)
3. Add a throttle rule for `*cmr.earthdata.nasa.gov*`
4. Set network throttling to "Slow 3G" or add a custom profile with high latency

This simulates real-world conditions where CMR can be slow.

### Test 1: CMR not blocking other layers
1. With throttling enabled, open Worldview
2. Add a granule layer like `VIIRS NOAA-20 Corrected Reflectance (True Color) *BETA*`
3. Add some regular WMTS layers (Land Surface Temp, Fire layers, etc.)
4. **Verify**: WMTS layers should render immediately while the granule layer loads separately

### Test 2: Granule reload fix
1. Add a VIIRS granule layer
2. Refresh the page (or copy the URL and open in new tab)
3. **Verify**: Granule layer should render on load (previously showed 0/10 granules)

### Test 3: Group Similar Layers toggle
1. Add multiple layers from different categories
2. Toggle "Group Similar Layers" checkbox on/off several times quickly
3. **Verify**: Layers maintain correct order, grouping works properly

### Test 4: Date changes with granules
1. Add granule + WMTS layers
2. Scrub through dates on the timeline
3. **Verify**: Layers update without blocking each other

### Test 5: A/B Compare with granules
1. Add a VIIRS granule layer
2. Click Compare → A|B
3. In the B side, add a different granule layer (e.g., MODIS)
4. Toggle between Spy and Swipe modes
5. **Verify**: Both granule layers render correctly on their respective sides

### Test 6: Projection switch with granules
1. Add a VIIRS granule layer in Geographic projection
2. Switch to Arctic projection
3. Switch back to Geographic
4. **Verify**: Granule layer renders correctly after each projection change

### Test 7: Layer reorder while loading
1. Enable network throttling
2. Add a granule layer (starts loading slowly)
3. While granule is still loading, add 2-3 WMTS layers
4. Drag and reorder the WMTS layers in the sidebar
5. **Verify**: Layer order in sidebar matches map, no corruption when granule finishes loading

### Test 8: Complex state share link
1. Add 2 granule layers, 3 WMTS layers, enable "Group Similar Layers", enable Compare mode
2. Copy the share link and open in a new browser/incognito window
3. **Verify**: All layers, grouping, and compare mode restore correctly

## PR Submission Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/nasa-gibs/worldview/blob/main/.github/CONTRIBUTING.md) doc

## Merging

@nasa-gibs/worldview


christof-wittreich:
I'm seeing an issue when I load a fresh instance of WV, add`VIIRS NOAA-20 Corrected Reflectance (True Color) *BETA*` layer, and then hide the `Corrected Reflectance (True Color) - Terra / MODIS` layer. When I do this, the Granule layer also stops showing imagery on the screen. It looks like this may only be an issue at the most recent time on the timeline, as it seems to work properly when trying it at a different date/time.

christof-wittreich:
Another issue I see is that when hiding another layer while a Granule layer is present, the request seems to be blocked by something with Granule layer needing to load or something. If I add a Granule layer and then try to hide another visible layer, the other layer takes a while to hide the imagery on the map. This does not occur when no Granule layer is present.

This issue does seem to happen in Prod as well, so it wasn't introduced with these changes, but I feel like it is related to the issue being addressed in this PR and therefore should also be fixed.

christof-wittreich:
I don't think there should be imagery appearing here where there is a gap in the blue availability bar. The imagery shown is older imagery (about ~3 hours old).

Tested on `develop` and it isn't happening there or in Prod, so I think this is an issue introduced by the changes in this branch.
image

christof-wittreich:
@vicgonzalez25 If I go to [this link](http://localhost:3000/?v=-187.50974132891147,-107.01872055432655,89.25643165586635,94.71108855605858&z=4&ics=true&ici=5&icd=6&l=Coastlines_15m,VIIRS_NOAA20_CorrectedReflectance_TrueColor_Granule&lg=true&t=2026-03-27-T16%3A42%3A59Z), then enter Comparison Mode, the Granule imagery appears on the A side but not the B side. If I then reload the page while in Comparison Mode, the Granule imagery shows on both A and B sides.

minniewong (approved):

minniewong (changes requested):
Layer ordering is incorrect in polar projections: I noticed that if I add a new base layer on top of the granule corrected reflectance layer, the new base layer covers the granule corrected reflectance layer. 
Steps to recreate:
1. Go to Arctic projection
2. Turn off Terra Corrected Reflectance
3. Add NOAA-20 Granule Corrected Reflectance (True Color)
4. Go back to a time with imagery
5. Go to "Add Layers" 
6. Go to "Corrected Reflectance" and add a corrected reflectance layer with a different band combination like Corrected Reflectance (Bands M11-I2-I1) from Suomi NPP/VIIRS
7. See error - The new base layer appears on top of the granule corrected reflectance layer. 
8. Refresh the page and see that the granule corrected reflectance layer appears on top again. 
9. Add another base layer
10. See error - the base layer is now on top of the granule corrected reflectance layer (which is in turn correctly on top of the previously added base layer)

Image

christof-wittreich (approved):

minniewong (approved):

PR #6562, opened 2026-04-08
WV-4029: Convert App component from class to functional

vicgonzalez25:
## Summary

Converts the top-level `App` component in [web/js/app.js](web/js/app.js) from a class to a functional component with hooks.

## What changed

- `onload()` startup logic → one-time `useEffect` (still wrapped in `util.wrap`)
- `componentDidMount` / `componentWillUnmount` → single `useEffect` for the `keydown`, `resize`, and `orientationchange` listeners with cleanup
- `componentDidUpdate` watcher on `numberOutagesUnseen` → `useEffect` + `usePrevious` from customHooks.js
- `static setVhCSSProperty` → module-level const
- Removed `what-input` library — replaced `[data-whatintent]` selectors in reset.scss and toolbar.scss with native CSS `:focus-visible`

### DIFF
--- a/web/js/app.js
+++ b/web/js/app.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import { useEffect } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 // eslint-disable-next-line no-unused-vars
@@ -6,6 +6,7 @@ import whatInput from 'what-input';
 
 // Utils
 import util from './util/util';
+import usePrevious from './util/customHooks';
 import { STARTUP } from './util/constants';
 import MapInteractions from './containers/map-interactions/map-interactions';
 // Toolbar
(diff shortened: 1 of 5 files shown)


Tomcariello (approved):

web/js/app.js
```diff
@@ -1,11 +1,12 @@
-import React from 'react';
+import { useEffect } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 // eslint-disable-next-line no-unused-vars
 import whatInput from 'what-input';
```
christof-wittreich: If this line isn't doing anything, this is probably as good a time as any for us to remove it

christof-wittreich (changes requested):

christof-wittreich (approved):

jwhurley1 (approved):

PR #6698, opened 2026-05-31
fix: iOS Canvas memory leak on colormap threshold change (#1510)

### LINKED ISSUE #1510: iOS Canvas memory exceeds the maximum limit when moving colorbar thresholds
**Describe the bug**
If you move the colorbar thresholds of a layer on an iOS device you will create lot of canvas and raise the maximum limit of the device (limit that just change with the new webkit release -> https://github.com/WebKit/webkit/commit/5d5b478917c685e50d1032ccf761ca53fc8f1b74#diff-b411cd4839e4bbc17b00570536abfa8f )

**To Reproduce**
1. Open the website
2. Add Cloud Top Height (Day)
3. Open Settings of the layer
4. Move threshold
5. Open safari debug console
6. Go to the Canvas tab
7. Move threshold
8. See the canvas memory use exceeds the maximum limit error (if you have lot of ram it can take some time but on iPad it come fast (the limit is 256MB)
9. From this moment safari don't want to allow you to create more canvas, openlayers loose references everything crash

**Expected behavior**
No error message and no crash

 - OS: Mac, IPad, 
 - Browser safari
 - Version 12

**Additional context**
The problem comes from https://github.com/nasa-gibs/worldview/blob/360c06322c909164ec1fca117c896c1a053cf700/web/js/ol/lookupimagetile.js
Each time the colormap change a new canvas is created and also a new context. 
I use the same function to filter my data with openlayers because the `postcompose` don't work for me so if you find a way to resolve this i would be very grateful !

hesam-oxe:
Fixed an 8-year-old memory leak where each colormap threshold 
change created a new Canvas element without destroying the 
previous one, causing unbounded memory growth on iOS Safari.

## Root Cause
`LookupImageTile.load()` creates a new Canvas on every invocation 
without clearing the previous reference. On iOS, Safari enforces 
a Canvas memory limit (~256MB), causing crashes after repeated 
threshold adjustments.

## Fix
Added explicit Canvas cleanup (setting dimensions to zero and 
nullifying the reference) before creating a new Canvas in both 
the image load callback and the PNG decode path.

## Verification
- Safari dev tools Canvas tab no longer shows accumulation
- Memory usage remains stable after multiple threshold changes
- Desktop Chrome tab memory stays around ~470MB instead of 
  climbing to 1GB+

Closes #1510

### DIFF
--- a/web/js/ol/lookupimagetile.js
+++ b/web/js/ol/lookupimagetile.js
@@ -56,6 +56,12 @@ LookupImageTile.prototype.load = async function () {
     let imageProcessed = false;
 
     const onImageLoad = function() {
+      // Clear existing canvas to prevent memory leak (#1510)
+      if (that.canvas_) {
+        that.canvas_.width = 0;
+        that.canvas_.height = 0;
+        that.canvas_ = null;
+      }
       that.canvas_ = document.createElement('canvas');
       that.canvas_.width = that.image_.width;
       that.canvas_.height = that.image_.height;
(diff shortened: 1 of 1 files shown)


minniewong:
This issue has been fixed with https://github.com/nasa-gibs/worldview/pull/6829

PR #6877, opened 2026-08-04
Make getVisMetadata skipLayers configurable

### LINKED ISSUE #6687: make the layers used in getVisMetadata configurable
All layers are assumed to have metadata available at https://gibs.earthdata.nasa.gov/layer-metadata unless listed in the hardcoded skipLayers [here](https://github.com/nasa-gibs/worldview/blob/main/tasks/build-options/getVisMetadata.js#L72).  

It would be nice if this was configurable in some way.  For example it could just be another option in the layer config files.

**Describe alternatives you've considered**
In order to use the getCapabilities functionality for self-hosted layers I have added a patch like this in the Dockerfile:

```
# Patch getVisMetadata.js to ignore internal layers
COPY --from=config /tmp/custom_layers.json /tmp/custom_layers.json
RUN sed -i "s|const skipLayers = \[|const skipLayers = [ ...require('/tmp/custom_layers.json').layerOrder, |" tasks/build-options/getVisMetadata.js
```

This works for us but a first-party solution would be ideal.

philyuchkoff:
Fixes #6687

## Description

The list of layers skipped when fetching GIBS layer metadata during the build process is currently hardcoded in `tasks/build-options/getVisMetadata.js`. Self-hosted deployments serving their own layers have no first-party way to skip their internal layers and resort to patching the script (e.g. via `sed` in a Dockerfile).

This PR makes the skip list configurable: additional layers can be added via `vismetadata.skipLayers` in `config/default/common/features.json`. The existing default list is preserved, so behavior for the default NASA configuration is unchanged.

## Changes

- `tasks/build-options/getVisMetadata.js`: rename the hardcoded list to `defaultSkipLayers` and merge with `configuredSkipLayers` read from the vismetadata config.
- `config/default/common/features.json`: add `vismetadata.skipLayers` (empty by default).
- `doc/features.md`: document the new option.

## How To Test

- Run a build with a layer that does not exist in GIBS added to `vismetadata.skipLayers` and confirm it is no longer fetched / fails the build.
- With an empty `skipLayers` (default config), behavior is identical to before this change.
- `npm run lint` passes.

## PR Submission Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/nasa-gibs/worldview/blob/main/.github/CONTRIBUTING.md) doc

@nasa-gibs/worldview

### DIFF
--- a/config/default/common/features.json
+++ b/config/default/common/features.json
@@ -35,7 +35,8 @@
                 "POCLOUD": "PO.DAAC",
                 "SEDAC": "SEDAC",
                 "AERONET": "AERONET"
-            }
+            },
+            "skipLayers": []
         },
         "cmr": {
             "url": "https://cmr.earthdata.nasa.gov/search/"
--- a/doc/features.md
+++ b/doc/features.md
@@ -8,6 +8,19 @@ CGI files. Next, enable these features in your configuration file.
 
 By default during the build process we try to fetch images of layers from our Snapshots application to show as previews in the layer picker component.  However, this will only work for layers that are present in NASA GIBS (Global Imagery Browse Services).  If you are serving your own layers, you will likely wish to disable this feature by setting `"previewSnapshots": false` in `config/default/common/features.json`.
 
+## Skipping Layer Metadata
+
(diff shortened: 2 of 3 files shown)


philyuchkoff:
Note for reviewers: local `npm test` shows 8 pre-existing failures in `web/js/modules/tour/util.test.js` and `web/js/app.test.js` (`TypeError: Cannot delete property 'location' of #<Window>`) caused by running Node v26 instead of the required v24.15.0 (jsdom incompatibility). These are unrelated to this change — the diff does not touch `web/`.

minniewong:
This has been fixed.

Issue #6927, opened 2026-08-30
"Geostationary Imagery Every 10 Minutes" tour currently doesn't seem to show imagery at step 2 onwards


momentmal:
Just tried the current deployment of https://worldview.earthdata.nasa.gov

Seems like the "Geostationary Imagery Every 10 Minutes" tour doesn't load the accompanying imagery from step 2 onward. Tested under Windows / Google Chrome (also in Incognito mode)

<img width="1083" height="919" alt="Image" src="https://github.com/user-attachments/assets/04ac5216-5514-4d21-a879-c9642547900d" />

<img width="1083" height="919" alt="Image" src="https://github.com/user-attachments/assets/47af639c-2dc9-454e-9289-57c854f24bc3" />

minniewong:
@momentmal Thanks for the report. We'll look into why the imagery is missing.

utkutugrul:
A fix for this issue is already open in #6931. I added regression coverage in #6932 to verify that when GetCapabilities omits dateRanges, the configured historicalRanges restore both the layer startDate and dateRanges.

The new test fails against develop and passes with the #6931 fix. The full test suite also passes under the repository-required Node 24.15.0.

minniewong:
@momentmal - We've fixed the issue in our last release. Please check and see that it works for you!

momentmal:
@minniewong thanks so much!
now, if you also would get some imagery for Europe into the tool, that would be fantastic!

<img width="251" height="168" alt="Image" src="https://github.com/user-attachments/assets/06a7507a-33c5-48f7-aa8d-b2905ed055ae" />

minniewong:
@momentmal We need to update that note! We do pull this data from EUMETSAT: https://go.nasa.gov/4yYbmIN

momentmal:
> [@momentmal](https://github.com/momentmal) We need to update that note! We do pull this data from EUMETSAT: https://go.nasa.gov/4yYbmIN

thanks for the hint, GeoColor looks even nicer!

https://go.nasa.gov/4d7mASS

PR #6932, opened 2026-09-02
Add regression coverage for missing historical date ranges

utkutugrul:
## Summary
- add regression coverage for adjustStartDates when GetCapabilities omits dateRanges
- verify configured historicalRanges restore both the layer start date and date ranges
- preserve the existing missing-time warning

## Context
This is a focused follow-up to #6931 for the scenario reported in #6927. It targets the feature branch so it does not duplicate the active fix.

## Validation
- npm run test:unit -- web/js/modules/layers/util.test.js --runInBand
- full pre-push suite: 388 test suites / 9,478 tests passed in both default and Australia/Sydney time zones
- npm run lint (0 errors; 34 pre-existing warnings)
- confirmed the new test fails on develop and passes with #6931

### DIFF
--- a/web/js/modules/layers/util.test.js
+++ b/web/js/modules/layers/util.test.js
@@ -464,6 +464,20 @@ describe('adjustStartDates', () => {
     expect(layer.startDate).toBeDefined();
     expect(layer.startDate).not.toBe('2000-01-01');
   });
+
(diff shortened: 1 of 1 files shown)


christof-wittreich:
@utkutugrul The unit test coverage will be updated as part of #6931 once it is merged

